PR #7, opened 2018-10-20
Rephrase setup.py description

sahensley:
Initially, I was going to change commited to commitment but decided to rephrase the line.


cs01:
This is much better, thanks!

Issue #18, opened 2018-10-21
pipx install jupyter says "No binaries associated with this package."


aiguofer:
Trying to install Jupyter gives me the above error

```
pipx install --python /usr/bin/python3 jupyter
No binaries associated with this package.
```

cs01:
I updated pipx to look for binaries of dependent packages when the desired package has no binaries directly associated with it. This is the case with jupyter since it's a "metapackage" that just installs other packages.

```
>> pipx install jupyter
Installing package 'qtconsole' with pipx would install 1 binaries
  - jupyter-qtconsole
Installing package 'notebook' with pipx would install 4 binaries
  - jupyter-notebook
  - jupyter-nbextension
  - jupyter-serverextension
  - jupyter-bundlerextension
Installing package 'jupyter-console' with pipx would install 1 binaries
  - jupyter-console
Installing package 'nbconvert' with pipx would install 2 binaries
  - jupyter-nbconvert
  - python
No binaries associated with package jupyter.
```

Are any of these binaries what you were looking for?

aiguofer:
Ahh, nice! yeah that's perfect. Wonder why it says `python` is installed under `nbconvert`?

cs01:
Something to do with the way they defined their `setup.py` file. Maybe this line? https://github.com/jupyter/nbconvert/blob/master/setup.py#L181

cs01:
> Hey, was playing around with this. Why did you remove the functionality in beaff34?

1. The logic to install was easy to update, but then calls to `uninstall` and `list` would also have to be updated, so the state of beaff34 was only partially done and would have been annoyingly unfinished (IMO).
2. It took much longer to look up the binaries included in multiple dependencies compared to the single package, and it actually wasn't even fully recursive; it just looked at the children of the package (not the children's children and so on). The speed of this probably could have been improved though since a new python subprocess was invoked each time the dependencies of a package were retrieved. The script itself could have been re-written to find them all recursively which would probably run fast since it would all be from a single python subprocess. I didn't have the time to try this out so I just removed it and added the listing. If this is re-written and proves to be roughly as fast as is it now, I think we can add this feature!

cs01:
This has been added in 0.12.3.0.
```
pipx install jupyter --include-deps
```

PR #28, opened 2018-10-24
Improve binary path collection

aiguofer:
Based on
https://packaging.python.org/specifications/entry-points/#use-for-scripts
the entry points should only be in the console_scripts and
gui_scripts sections.

This should help prevent nbconvert from pulling python as a dependency binary in #18


cs01:
This is really nice work, thank you! I am going to start a contributors list and add your name on the README.

PR #31, opened 2018-10-25
rename main.py to __main__.py (fix #26)

### LINKED ISSUE #26: Rename `main.py` to `__main__.py`
If you have a module in a package named `__main__.py`, then `python -m pipx` will work as expected.

r-darwish:


cs01:
Thanks for making this. For some reason I find putting lots of code in `__main__` to be offputting. Is it standard practice?

I would prefer something like
```
from .main import cli

cli()
```

in `__main__` and leave setup.py unchanged.

cs01:
added in 017cbdf6

PR #32, opened 2018-10-25
Update README.md with curl error 23 notes

sahensley:
I added a note about what causes cURL error 23.  Eventually it may be worth moving to a troubleshooting section as other troubleshooting steps are collected.


cs01:
Good call, thanks.

PR #36, opened 2018-10-31
fix typo in spec help string

BackwardSpy:
Just a little typo I spotted when trying pipx out for the first time.

Looks like a nice tool so far!


cs01:
Thank you!

PR #39, opened 2018-11-22
Support Python 3.5.x

### LINKED ISSUE #38: Consider lowing Python version requirements
Hey, I recently attempted to use pipx on a Debian stretch system, which is still on Python 3.5.
As such, Python 3.5 support would be nice to have in pipx. As far as I can tell, the only feature that requires 3.6 that pipx requires is PEP 498 (literal string interpolation).

Would you be willing to use `string.format()` over PEP 498 strings?

neersighted:
Move to using logging.log()'s native printf-style formatting, as well as
string.format(), in order to support Python 3.5.

Python 3.5 is still packaged in Debian stable, and as such it should be
supported, in my opinion.

Fixes #38


varac:
Hej @neersighted Thx for this PR, I'd like to use pipx on stretch as well!
Can you pls rebase since there're conflicts ?

cs01:
Thanks for putting this together. Unfortunately pipx is a 3.6+ only project and there are no plans to change that. If you want to fork it and maintain a 3.5-compatible version you can do that, but I am going to go ahead and close this PR since it is not something pipx will be supporting.

PR #45, opened 2018-11-28
add dependencies to an existing pipx-managed virtualenv

BackwardSpy:
Hi, I've been using pipx for a little while now and I really like it!

One thing I found myself doing quite often was running commands like the following to add useful packages to my `ptpython` REPL:

```shell
~/.local/pipx/venvs/ptpython/bin/pip install requests
```

I would also occasionally do this to add an optional dependency for a package I'd previously installed with `pipx`.

I spent some time prototyping a feature for pipx called "inject" that wraps this into a convenient command. It's probably a little rough around the edges as this was my first proper foray into the pipx code base.

Here is a little snapshot of me playing with the new command, to demonstrate what it does:

```
$ pipx inject ptpython requests
Can't inject requests into nonexistent venv /home/chris/.local/pipx/venvs/ptpython!

$ pipx install ptpython
  installed package ptpython, 2.0.4
  These binaries are now globally available
    - ptipython
    - ptipython3
    - ptipython3.6
    - ptpython
    - ptpython3
    - ptpython3.6
done!

$ ptpython
In [1]: import requests
ModuleNotFoundError: No module named 'requests'

No module named 'requests'
In [2]: exit()

$ pipx inject ptpython requests
   package ptpython, 2.0.4
    - ptipython
    - ptipython3
    - ptipython3.6
    - ptpython
    - ptpython3
    - ptpython3.6
done!

$ ptpython
In [1]: import requests
In [2]: requests.get('https://httpbin.org/get')
Out[2]: <Response [200]>
```

Please let me know if you think this could be a useful addition to pipx.

### DIFF
--- a/pipx/main.py
+++ b/pipx/main.py
@@ -488,12 +488,12 @@ def install(
     print(f"done! {stars}")
 
 
-def inject(venv_dir: Path, package: str, package_or_url: str, verbose: bool):
+def inject(venv_dir: Path, package: str, verbose: bool):
     if not venv_dir.exists() or not next(venv_dir.iterdir()):
         raise PipxError(f"Can't inject {package} into nonexistent venv {venv_dir}!")
 
     venv = Venv(venv_dir, verbose=verbose)
-    venv.install_package(package_or_url)
+    venv.install_package(package)
 
     if venv.get_package_version(package) is None:
         raise PipxError(f"Could not find package {package}. Is the name correct?")
@@ -605,10 +605,7 @@ def run_pipx_command(args):
             force=args.force,
         )
     elif args.command == "inject":
-        package_or_url = (
-            args.spec if ("spec" in args and args.spec is not None) else args.dependency
-        )
-        inject(venv_dir, args.dependency, package_or_url, verbose)
+        inject(venv_dir, args.dependency, verbose)
     elif args.command == "upgrade":
         package_or_url = (
             args.spec if ("spec" in args and args.spec is not None) else package
@@ -748,7 +745,6 @@ def get_command_parser():
         "package", help="Name of the existing pipx-managed virtualenv to inject into"
     )
     p.add_argument("dependency", help="the package to inject into the virtualenv")
-    p.add_argument("--spec", help=SPEC_HELP)
     p.add_argument("--verbose", action="store_true")
 
     p = subparsers.add_parser("upgrade", help="Upgrade a package")
--- a/pipx/main.py
+++ b/pipx/main.py
@@ -59,7 +59,7 @@ PIPX_HOME and PIPX_BIN_DIR, respectively.
 )
 PIPX_USAGE = """
     %(prog)s [--spec SPEC] [--python PYTHON] BINARY [BINARY-ARGS]
-    %(prog)s {install, upgrade, upgrade-all, uninstall, uninstall-all, reinstall-all, list} [--help]"""
+    %(prog)s {install, inject, upgrade, upgrade-all, uninstall, uninstall-all, reinstall-all, list} [--help]"""
 
 
 def print_version() -> None:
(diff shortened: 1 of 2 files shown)


cs01:
Hey thanks for using pipx and for contributing!

This is currently possible with
```
pipx install ptpython --spec requests --force
```
Let me know what you think. I think it's a little awkward but it works. I think I like `inject` a little more, but I am really trying to keep the api small.

BackwardSpy:
That's totally fair! I hadn't realised that you could do it that way. I'm happy using `--spec` and `--force` for this functionality. Having a dedicated command does make it slightly cleaner to use, but I agree with your point about keeping the interface small.

I shall await your judgement! :smile:

taketwo:
I started to use `pipx` a few days ago and was looking for exactly this feature of "injecting" extra packages. If a command like the one in this PR was a part of the `--help` output, I would definitely discover it and use. The `--spec --force` workaround? Yes, apparently it works. No, I was not creative enough to come up with this.

So I think a dedicated command is not only about cleaner experience, but also about discoverability of the feature.

BackwardSpy:
Discoverability is certainly something that can be improved.

Would an alias be preferable? I could cut the code down quite a bit and just make it internally call through to the normal install functionality. It would still have a measure of redundancy to it, but it should be easier to maintain.

cs01:
Should an option be added to uninstall then as well? Has that been a part of your workflow? Seems like it would make sense to add it if we add the ability to inject.

> Would an alias be preferable? 

Good question. I will say no here because the "force" option I just added doesn't really align with running `inject`, and your error messages are tailored for the exact command so they'll make more sense to the user. 

Before merging I want to be *sure* we like the name "inject" because changing it will be a breaking API change. Not that I think it's bad, just want to brainstorm on other names. 

What do you think about these? 
* `inject`
* `add-dep`
* `add-package`
* `install-dep`

BackwardSpy:
I'm not precious about the inject name, that's just what came to mind as I was writing it. `add-package` is nice in that it gives us room to pair it with a `remove-package` command.

vlcinsky:
I tested the `--spec` with `--force` and see, that if I use it, the environment is not properly recreated if `pipx reinstall-all <python>` is used.

My expectation is, once the tool with modified packages is created, it shall allow reinstallation with the same (or better) result.

cs01:
> I tested the --spec with --force and see, that if I use it, the environment is not properly recreated if pipx reinstall-all <python> is used.
My expectation is, once the tool with modified packages is created, it shall allow reinstallation with the same (or better) result.

Currently what it does is a full uninstall (`rm -r` of the venv), then a new install of _just the package name_. What I can try to do is capture the output of `pip freeze` in the virtual env _before_ uninstalling, then instead of just installing the package it can install all the packages captured from the `pip freeze` call. Does that sound good?

vlcinsky:
@cs01 yes, whatever allows persistence of base and extra requirements would work. Base requirement seems to be encoded into virtualenv name. I did not find any file which would seem to track those extra requirements so I was not surprised, the reinstallation is ignoring them.

Regarding pip freeze - it would probably work, but as each package is ready to install other dependent ones, it could be sometime a bit too much (and there is a question, if reinstallation should preserve listed versions or not).

Good alternative would be to keep somewhere only those special requirements. One option could be to have `requirements.txt` where `pipx install <base_package> --spec <extra_package> --force` would add a line (or more for multiple `--spec`) with the `<extra_package>`.

High-level (but possibly with too much complexity) would be to use e.g. `pyproject.toml` to record these things, e.g. in the way, `poetry` does.

pipx/main.py
```diff
@@ -722,6 +741,16 @@ def get_command_parser():
...
     )
 
+    p = subparsers.add_parser(
+        "inject", help="Install a package into an existing virtualenv"
+    )
+    p.add_argument(
+        "package", help="Name of the existing pipx-managed virtualenv to inject into"
+    )
+    p.add_argument("dependency", help="the package to inject into the virtualenv")
+    p.add_argument("--spec", help=SPEC_HELP)
```
cs01: Spec is not needed since the package name is already provided. The dependency is already what pipx considers a spec, so this line should be removed and the other code that tries to use it should be removed too.

pipx/main.py
```diff
@@ -590,6 +604,11 @@ def run_pipx_command(args):
             verbose,
             force=args.force,
         )
+    elif args.command == "inject":
+        package_or_url = (
+            args.spec if ("spec" in args and args.spec is not None) else args.dependency
+        )
+        inject(venv_dir, args.dependency, package_or_url, verbose)
```
cs01: I think this should just be
```suggestion
        inject(venv_dir, args.dependency, verbose)
```

BackwardSpy:
Removed `--spec` from `pipx inject` as requested.

pipx/main.py
```diff
@@ -488,6 +488,20 @@ def install(
...
+    if not venv_dir.exists() or not next(venv_dir.iterdir()):
+        raise PipxError(f"Can't inject {package} into nonexistent venv {venv_dir}!")
+
+    venv = Venv(venv_dir, verbose=verbose)
+    venv.install_package(package)
+
+    if venv.get_package_version(package) is None:
+        raise PipxError(f"Could not find package {package}. Is the name correct?")
+
+    _list_installed_package(venv_dir)
```
cs01: This should be removed since the listing will be unchanged.

BackwardSpy: It's only there because I initially duplicated the `install` command. Removed!

cs01:
Great stuff, thank you for the PR!

BackwardSpy:
My pleasure!

PR #51, opened 2018-12-09
Added a development_mode install option. 

kannanmr:
This new option `--develop` allows code-edits with out doing `pipx install --force`. Very useful during development. 

Developer flow: 

- Go to your tool folder. 
- Do `pipx install --develop --spec . <pkgname>`. This installs the package as usual, the difference is `pip` is invoked with `--editable`.

Fixes: https://github.com/cs01/pipx/issues/50

### DIFF
--- a/pipx/main.py
+++ b/pipx/main.py
@@ -73,13 +73,19 @@ class PipxError(Exception):
 
 class Venv:
     def __init__(
-        self, path: Path, *, verbose: bool = False, python: str = DEFAULT_PYTHON
+        self,
+        path: Path,
+        *,
+        verbose: bool = False,
+        develop: bool = False,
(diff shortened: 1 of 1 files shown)


cs01:
Thank you for making this. There are a bunch of other functions that require these arguments as well, so I went ahead and made this change.

This has been added in pipx-app 0.10.4.0.

kannanmr:
Perfect, thanks for taking care of this. I am out of country on vacation right now, cannot wait to pull these changes in and start using. 

BTW, that was a good idea to introduce a generic 'pip_args' argument. 👍

nickolay:
For future reference, the command to use seems to be: `pipx install --editable .`

Issue #96, opened 2019-02-25
['Win10', 'Py3.72', 'pipx-0.12.3.1'] TypeError: unsupported operand type(s) for +=: 'set' and 'list'


Laurentiu-Andronache:
Wait! Don't help me! I'm trying to submit a commit and fix this one myself!

```
C:\Users\Nebu>pipx uninstall-all
Traceback (most recent call last):
  File "C:\Python37\Scripts\pipx-script.py", line 11, in <module>
    load_entry_point('pipx==0.12.3.1', 'console_scripts', 'pipx')()
  File "c:\python37\lib\site-packages\pipx\main.py", line 496, in cli
    exit(run_pipx_command(parsed_pipx_args, binary_args))
  File "c:\python37\lib\site-packages\pipx\main.py", line 191, in run_pipx_command
    commands.uninstall_all(PIPX_LOCAL_VENVS, LOCAL_BIN_DIR, verbose)
  File "c:\python37\lib\site-packages\pipx\commands.py", line 435, in uninstall_all
    uninstall(venv_dir, package, local_bin_dir, verbose)
  File "c:\python37\lib\site-packages\pipx\commands.py", line 415, in uninstall
    binary_paths += dep_paths
TypeError: unsupported operand type(s) for +=: 'set' and 'list'
```

Laurentiu-Andronache:
relevant pull request https://github.com/pipxproject/pipx/pull/97

cs01:
This has been fixed in 0.12.3.2

PR #99, opened 2019-02-25
for #96

Laurentiu-Andronache:
for #96


cs01:
Thanks for the fix! This change should be made at the source of the problem, in Venv.

cs01:
I went ahead and fixed this myself. Thanks for making me aware of it!

PR #101, opened 2019-02-27
do not advertise non-existent py2.6 support

luto:
[The readme](https://github.com/pipxproject/pipx#system-requirements) and parts of setup.py state that pipx is python 3.6+, but the setup.py mentions python 2.6. This PR removed the false advertising ;)


cs01:
Ah must have been a copy+paste thing. Thanks!

PR #118, opened 2019-03-26
add runpip command

cs01:
Add the `runpip` command, and replace cowsay with [pycowsay](https://github.com/cs01/pycowsay). 

Created a pull request so users can provide feedback before it's landed.

```
>> pipx runpip --help
usage: pipx runpip [-h] [--verbose] package [pipargs [pipargs ...]]

Run pip in an existing pipx-managed Virtual Environment

positional arguments:
  package     Name of the existing pipx-managed Virtual Environment to run pip
              in
  pipargs     Arguments to forward to pip command

optional arguments:
  -h, --help  show this help message and exit
  --verbose
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+0.12.4.0
+* Add `runpip` command
+
 0.12.3.3
 * Update logic in determining a package's binaries during installation. This removes spurious binaries from the installation. (#104)
 * Improve compatibility with Debian distributions by using `shutil.which` instead of `distutils.spawn.find_executable` (#102)
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
 <a href="https://travis-ci.org/pipxproject/pipx"><img src="https://travis-ci.org/pipxproject/pipx.svg?branch=master" /></a>
 
 <a href="https://pypi.python.org/pypi/pipx/">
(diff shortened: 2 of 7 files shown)


florimondmanca:
Hey, that sounds cool! But can you exhibit an example use case or problem this solves? Not a pipx user myself but I thought pipx aimed at running executables. What would be the point of installing extra packages in their virtualenv?

cs01:
Originally I had the same question when lots of folks wanted to "inject" additional packages into specific virtual environments, which is how pipx ended up with the `inject` command. That command has been well received, and I even found myself using it. 

The natural next step was to request the "uninject" command, which would add another command to pipx. Then, someone wanted to run `pip install -r requirements`, which would result in another pipx command :open_mouth:. Adding all these commands would get unweildy fast and would probably only serve 20% (or less) of pipx users. So instead,`runpip` is proposed as a catchall for all advanced uses that want to perform manual operations on pipx-managed virtual environments.

Some relevant conversations:

https://github.com/pipxproject/pipx/issues/67
https://github.com/pipxproject/pipx/issues/88

florimondmanca:
I see now! Makes a lot of sense.

cs01:
I went ahead and added this in a push to master along with some other changes.  You can use it in v0.13.

PR #127, opened 2019-04-06
Refine usage of 'entry points' to 'console scripts'. 

jaraco:
I suggest this change to use the more precise terminology of "console scripts", the specific form of entry points that pipx exposes. I don't feel strongly about it, but I do think it helps to be more precise here for clarity and to avoid perpetuating the perception that entry points and console scripts are synonymous.


jaraco:
I now understand that the readme needs to be updated in two places, first in the template, then run the generator, so I manually ported the change to the template in this PR.

cs01:
I think the entry point terminology is important to keep because it illustrates the idea that the user can enter the Python code through the terminal. pipx focuses on a certain type of entry point, the "console script". It is precise and accurate to call them "console script entry points". Can it be changed to that?

Issue #121, opened 2019-03-26
Make ensurepath cross-platform


ofek:
by using https://github.com/ofek/userpath

cs01:
Thanks for sharing. I have so much going on at the moment, but am open to PR's.

ofek:
Resolved by #131

PR #131, opened 2019-04-06
Use userpath to ensure the path

jaraco:
Ref #121.


jaraco:
As I started working on this, I began to think - maybe pipx shouldn't wrap this call, but should simply refer the user to use userpath. If only there were a way to invoke a runpy module from `pipx run`, you pipx could recommend `pipx run -m userpath append ~/.local/bin`. I guess you could recommend simply `{sys.executable} -m userpath append ~/.local/bin`.

jaraco:
In these last two commits, I remove the 'ensurepath' command. It seems to me if userpath already supplies a command that largely accomplishes the same goals as 'pipx ensurepath', it's not worth the maintenance burden to maintain a shim around it, but rather the instructions and warnings could simply defer to it. By making it a dependency, the command should be available alongside pipx.

cs01:
This is amazing! I wish I knew about it earlier. I definitely agree that it's nicer to just let users call userpath themselves. However, there is a bunch of code that uses the `ensurepath` command. https://github.com/search?q=pipx+ensurepath&type=Code

At the least the `ensurepath` command should remain during a transition period with a deprecation warning. 

pipx supports using the env var `PIPX_BIN_DIR` instead of the default location of ~/.local/bin, so ensurepath could call `userpath append $PIPX_BIN_DIR <any other user cli args>`, which might alleviate some confusion and would be a nice benefit of maintaining a shim.

cs01:
For example this homebrew formula would also have to be updated: https://github.com/Homebrew/homebrew-core/pull/39020/files

One other thought is that pipx should define a valid version range in setup.py since it’s depending on userpath’s api (assuming it has semantic versioning).

jaraco:
Good consideration. I was working from the assumption that the interface was only used manually an incidentally, so backward-compatibility would be bearable. I was hoping to avoid a deprecation period or any attempt to map the existing functionality to the new functionality, as both efforts would be substantially more involved.

The example you give (Homebrew) I think could handle a backward-incompatible release of pipx. They've effectively "vendored" the instructions for adding the user path, but they've also pinned the requisite version. Presumably, if pipx were release these proposed changes with a note about the incompatibility, the Homebrew maintainers would update the formula with a new caveat, following the release notes for the new pipx version.

Nevertheless, I respect your desire to give a deprecation period, so here's what I recommend:

1. Temporarily retain the existing implementation of `pipx ensurepath`, but add a message in all invocations directing users to the README about `userpath`.
2. Retain `userpath` changes as the preferred implementation (docs in new version make no reference to ensurepath).
3. After a satisfactory deprecation period, remove `ensurepath`.

How does that sound?

jaraco:
In this latest push, I've implemented steps 1 and 2 as I've proposed.

I've tried to create a new patch for (3), but I'm having difficulty with git. What you want to revert is 13c8137c8f79b12a1d102d48832ae837988421c0 and a0790bbb8601a378b94e4da50e1440112ae376ed.

Tests are failing on mypy again, even though I've merged with master; I'm not sure why. Error is `no library stub for setuptools`.

cs01:
Your 3 step plan sounds good to me.

The only error I see is 

```
Running 'black --check pipx tests'
would reformat /home/travis/build/pipxproject/pipx/pipx/commands.py
All done! 💥 💔 💥
```
which isn't a big deal. I'm going to go ahead and merge this and then fix the error. Thanks for this suggestion, it simplified things nicely :smile:.

jaraco:
I can no longer find the merged commit (85863f5):

```
pipx master $ git fetch origin 85863f5                                                                                                                                         
fatal: couldn't find remote ref 85863f5
```

I imagine there's probably some way to pull those commits, but I'm not sure how.

I appreciate that you like clean commits, but this approach of rewriting history has some negative effects:

- my plan for reverting the "compatibility" commits is no longer viable, as those commits don't exist. Even with a rebased commit, I'd like to be able to check out the original branch, perform the reverts, and then cherry-pick those revert commits.
- Attribution is lost (e.g. jaraco doesn't appear in [the contributors](https://github.com/pipxproject/pipx/graphs/contributors)).
- all the changes for the 0.13.1.0 release are collapsed into one commit.

I'm not sure what options you may have, but I wanted to express my concerns.

cs01:
Thanks for bringing that up, I should have proactively updated you in this. 

When I merged your pull requests I must have accidentally merged without squashing (which is what I normally do). This resulted in a messy master branch, and I think I tended up force pushing a big commit. Apologies for the inconvenience and lack of attribution. I definitely value your contributions and don’t intend for that to happen again should you create another pull request.  

Shouldn’t your commits still exist in your branch though?

jaraco:
Yes, some of those commits I made still exist, but not the merge commit nor other edits you may have made before or after the merge... so it's difficult for me to trace the ancestry of various changes. It's not a big deal. In c02f25b, I was able to merge with master, deferring to master for conflicts and then perform the reverts as I'd planned - so those should be rebaseable or cherry-pickable onto master easily enough.

PR #133, opened 2019-04-07
Run tests with tox

jaraco:
This change adds a tox.ini file and utilizes it to facilitate the venv creation for running tests. This simplifies the instructions for getting started and for developers that have tox installed, allows testing with a simple `tox` invocation. As a result, it fixes #25 and reduces the likelihood of #130 and #132.

It also removes the dependency on `make` to build the docs, which also has a dependency on an activated virtual environment.

I've attempted to make this change backward-compatible, such that it doesn't affect the existing development workflow, though my recommendation would be to adopt tox and remove the makefile.

Issue #132, opened 2019-04-06
Running test suite installs pipx into system python


jaraco:
Simply running `python setup.py test` in the repository causes `.` to be installed to the system site packages. That's unexpected and undesirable.

cs01:
If you follow the instructions in the readme this will not happen. https://github.com/pipxproject/pipx#contributing

jaraco:
I was unaware of the instructions. Unfortunately, I can't follow those instructions as they assume a virtualenv workflow that [doesn't play well with xonsh](https://xon.sh/python_virtual_environments.html). The instructions are also Unix-specific. A tox-based workflow should be compatible with other shells and platforms.

PR #135, opened 2019-04-07
Require venv

jaraco:
As revealed in #132 and #133, if the tests aren't run in a venv, if the instructions aren't followed exactly, the tests will fail. This modest change adds some assertions to the test suite to cause it to fail fast and with a clear error message when these expectations aren't met.


jaraco:
The tests are failing because the assertion isn't true in CI. It will become true in CI if #133 is accepted, so I suggest maybe accepting that change then re-run this one. I think it's better to align processes between CI and development rather than supporting separate dev/CI environments. Let me know your thoughts.

jaraco:
I recently learned about a feature of tox that allows the project to declare tox plugins to be present, [the requires directive](https://tox.readthedocs.io/en/latest/config.html#conf-requires). I'll update this PR to include that declaration also.

cs01:
Thanks.  It’s good to have you helping on this project, I love how proactive and thorough you are.

jaraco:
Oh, wow. So I uninstalled tox-venv and tried running this latest commit. Instead of giving me an error (that my dependencies weren't met), tox just built the necessary environment and then used that environment to invoke tox again:

```
pipx bugfix/venv-required $ pip uninstall tox-venv                                                                                                                             
Uninstalling tox-venv-0.4.0:
  Would remove:
    /Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/tox_venv-0.4.0.dist-info/*
    /Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/tox_venv/*
Proceed (y/n)? y
  Successfully uninstalled tox-venv-0.4.0
pipx bugfix/venv-required $ tox                                                                                                                                                
.tox create: /Users/jaraco/code/public/pipx/.tox/.tox
.tox installdeps: tox-venv, tox >= 3.2
python recreate: /Users/jaraco/code/public/pipx/.tox/python
...
```

That means the contributing instructions no longer _require_ the user to have `tox-venv` installed, only tox... and tox will ensure the requisite dependencies are available in `.tox/.tox` (if not already met).

jaraco:
This latest commit removes the mention of `tox-venv` from the new contributor's consideration. They may want to have it installed with their tox to (a) have it by default on Python 3 environments and (b) avoid the need for `.tox/.tox`, but that's a small optimization that's unimportant for the general case, and because it does influence how other projects would run under tox, best to leave that detail up to the developer for whatever's best for their environment.

jaraco:
Tests are failing, but with the same error as master, so I believe this change is ready to go. Feel free to close/re-open once master is passing again or accept despite the failure.

jaraco:
Yes, I see that now. Weird that 'templates/readme.md' has only the intended changes, but 'README.md' has reversions. I suspect in master, 'templates/readme.md' is missing changes.

cs01:
> Yes, I see that now. Weird that 'templates/readme.md' has only the intended changes, but 'README.md' has reversions. I suspect in master, 'templates/readme.md' is missing changes.

Darn it. Hopefully your message at the top of the README will prevent that from happening again. Sorry I missed that in the review.

> Tests are failing, but with the same error as master, so I believe this change is ready to go. Feel free to close/re-open once master is passing again or accept despite the failure.

Just pushed a fix to master. Will close/open.

jaraco:
Indeed, in 31332d and cec8da5 and 349c507 (and maybe others), we've missed a few edits.

jaraco:
I'll need to re-generate docs after pulling those changes from master.

cs01:
I am going to be migrating the docs to mkdocs soon (similar to [gdbgui](https://github.com/cs01/gdbgui)), so hopefully this won't be happening anymore.

cs01:
I resolved a bunch of merge conflicts in github’s web ui which may have compounded the issue as well.

PR #138, opened 2019-04-10
HTTP redirect, proxy support (pipx run https://....)

wtnb75:
replace `http.client` to `urllib.request` for some HTTP feature (30x redirect, proxy, etc...)

now you can `pipx run URL` behind proxy.

```
# export https_proxy=http://your-proxy-host:port
# pipx run https://gist.githubusercontent.com/cs01/fa721a17a326e551ede048c5088f9e0f/raw/6bdfbb6e9c1132b1c38fdd2f195d4a24c540c324/pipx-demo.py
pipx is working!
```

### DIFF
--- a/pipx/commands.py
+++ b/pipx/commands.py
@@ -3,7 +3,6 @@
 
 import datetime
 import hashlib
-import http.client
 import logging
 import multiprocessing
 import os
@@ -13,6 +12,7 @@ import subprocess
 import textwrap
 import time
 import urllib.parse
+import urllib.request
 from pathlib import Path
 from shutil import which
 from typing import List, Optional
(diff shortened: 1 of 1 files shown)


pipx/commands.py
```diff
@@ -180,14 +180,7 @@ def _remove_all_expired_venvs():
...
 def _http_get_request(url: str):
-    parts = urllib.parse.urlparse(url)
-    conn = http.client.HTTPSConnection(parts.hostname)
-    conn.request("GET", parts.path)
-    response = conn.getresponse()
-    if response.status != 200:
-        raise PipxError(response.reason)
-
-    return response.read().decode("utf-8")
+    return urllib.request.urlopen(url).read().decode("utf-8")
```
cs01: Nice! This is much simpler. Are there any downsides?

Also, can the status code be checked (`getcode()`) and an appropriate `PipxError` error be raised if it doesn't start with a `2` or `3`?

Would be nice if we could detect the actual encoding as described in this section of chardet: https://chardet.readthedocs.io/en/latest/faq.html#yippie-screw-the-standards-ill-just-auto-detect-everything

wtnb75:
@cs01 thank you for your review!

- handle exceptions and raise PipxError
  - urllib raise `urllib.error.HTTPError` when get status 40x or 50x.
  - ..and raise `urllib.error.URLError` when get invalid url(bad schema, host not found, etc...)
- use charset in content-type to decode.

cs01:
Thanks for the PR!

PR #139, opened 2019-04-11
Fix some typos: Envrionment -> Environment.

cjerdonek:
This fixes a few of the same typo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ Install and execute binaries from Python packages.
 Binaries can either be installed globally into isolated Virtual Environments
 or run directly in an temporary Virtual Environment.
 
-Virtual Envrionment location is ~/.local/pipx/venvs.
+Virtual Environment location is ~/.local/pipx/venvs.
 Symlinks to binaries are placed in ~/.local/bin.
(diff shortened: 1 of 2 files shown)


cs01:
Thanks! I don't think this is documented, but the README.md is generated by running `make docs`, which pulls in changes from the help text in the source code. It shouldn't be modified directly. If you want to regenerated it that is fine, and if not I can do it myself.

 In the future I will add a note saying how to generate the README and add a comment in it saying that it was autogenerated.

cjerdonek:
Okay, thanks. Does that mean the README is currently out of date with the code? Otherwise, how can you tell I didn’t use the generation process given the same changes occur in both spots (and the word length also remained the same)?

cs01:
I thought you changed text outside of the code, and therefore had to modify [templates/readme.md](https://github.com/pipxproject/pipx/blob/master/templates/readme.md) so future doc generation calls wouldn't have your change. 

But looking more closely it looks like you only changed code, and manually updated what would have been generated by `make docs`. Apologies for the confusion 😄  and thanks again for the pull request!

PR #141, opened 2019-04-21
WIP: Refresh template based on root doc

jaraco:
This PR attempts to address the edits accidentally made to the generated README and not the template as [identified here](https://github.com/pipxproject/pipx/pull/135#issuecomment-485266410).


cs01:
Nice! Thanks!

jaraco:
I found two more. Pushed as 13b5d58.

jaraco:
Actually, that's still not right.

PR #142, opened 2019-04-21
Another bit of template refresh

jaraco:
Following on #141, I believe this is the last change needed to get the template back in sync.


jaraco:
I've merged this into #135 and now the diff looks right to me. There was a still a straggling `Python` vs `python`, but I believe the end result is right. You can accept this PR or not, but when you accept 135, this should get accepted automatically.

Issue #155, opened 2019-05-04
issues with userpath dependency


AlJohri:
I have some issues with the direct and hard dependency on `userpath`.

The main issue is the dependency on a separate top level _python_ CLI tool: I use `pipx` to wrap ALL python CLI tools (that are not available in a package manager, i.e. homebrew). The setup instructions for pipx now **require** a separate python CLI tool which I ideally would have installed through pipx itself (`pipx install userpath`).

When installing via homebrew, only `pipx` binaries are added to the PATH- cli tools from dependencies are not added. Since I don't install **any** python packages globally (`pip freeze` is empty), I can't run `python3 -m userpath ...`.

I recommend either:

1. make `userpath` a subcommand of `pipx` similar to `ensurepath` OR
2. taking the approach of [pyenv](https://github.com/pyenv/pyenv#basic-github-checkout) and several other libraries out there of telling users to pipe the command into their shell profile:
    ```
    echo 'export PATH="$HOME/.local/bin:$PATH"' >> ~/.bash_profile
    ```
    
    - Zsh note: Modify your ~/.zshenv file instead of ~/.bash_profile.
    - Ubuntu and Fedora note: Modify your ~/.bashrc file instead of ~/.bash_profile

AlJohri:
cc @ofek

cs01:
@AlJohri pipx lists userpath as a dependency in its setup.py file. Does homebrew call pip or somehow install the dependencies listed in setup.py? If so, `python3 -m userpath ...` should work.

@jaraco I believe you had thoughts on this?

jaraco:
> 2. taking the approach of pyenv and several other libraries out there of telling users to pipe the command into their shell profile

This approach requires the documentation to either take into account all possible environments (Windows, bash, other shells) or require the user to translate the instructions to those environments. I'd really like to see a tool that handles all these cases. `userpath` does that.

> 1. make userpath a subcommand of pipx similar to ensurepath

My feeling was that by making it a subcommand, it requires pipx to supply an interface, essentially duplicating the interface that `userpath` already presents, and maintaining a coupling between the two libraries.

Instead, by merely requiring the `userpath` dependency, there's a nice separation of concerns. `userpath` focuses on its specialty and `pipx` focuses on its own and defers to `userpath` when path manipulation is needed.

> When installing via homebrew, only pipx binaries are added to the PATH.

[...]

PR #157, opened 2019-05-05
WIP: Remove ensurepath compatibility

jaraco:
This PR reverts the two commits that restored ensurepath compatibility and represents the follow-up work I had planned in #131. Here's what I did - grabbed the code from the accepted merge request, merged that with the latest master (2f4e48bba8a496b018d82b24e36986efaa4cea61), reverted the two commits (resolving conflicts), and now submitting the request. I fully expect this to be squashed or the two commits cherry-picked (to avoid re-introducing all the history).

I've marked it as WIP, as I don't expect this to be merged until (a) enough time has gone by that ensurepath can be dropped and (b) issue #155 and other emergent issues are satisfactorily resolved.


cs01:
@jaraco I think we talked about this somewhere, but I think the pipx API should go in the opposite direction and keep `pipx ensurepath`. It maintains the same requirement of non-breaking changes to `userpath` yet provides a more cohesive cli UI since everything is discoverable through `pipx --help`.

I created an issue (https://github.com/pipxproject/pipx/issues/178) and a pull request (https://github.com/pipxproject/pipx/pull/179).

jaraco:
Very well. Thanks for the consideration.

PR #158, opened 2019-05-05
Add (semi-private) 'userpath' subcommand to pass-through

jaraco:
This command takes an approach of very early intercepting a `userpath` subcommand and allowing the `userpath` library to execute it. This approach creates one distinct touch point and avoids dealing with interactions with the parsing of other subcommands.

### DIFF
--- a/pipx/main.py
+++ b/pipx/main.py
@@ -3,6 +3,7 @@
 import argparse
 import logging
 import os
+import runpy
 import shlex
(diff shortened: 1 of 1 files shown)


jaraco:
This is a proposal for consideration while investigating solutions to #155.

jaraco:
This is not the right approach.

Issue #44, opened 2018-11-27
Support bootstrap installations with pip install --user


kwill:
**User story:** I am an experienced GNU/Linux developer. I only install non-distro packages with installers I trust, so that I reduce the risk of running malicious code. For example, I use `pip install --user` and to install non-Debian Python packages from PyPI.

**Problem:** Thus, I don't want to run `curl https://raw.githubusercontent.com... | python3` to install pipx.

**Suggested solution 1:**

A PyPI package called `pipx-bootstrap` that functions as follows:

```
pip install --user pipx-bootstrap
pipx-bootstrap # installs pipx-app from PyPI into correct location
pip uninstall pipx-bootstrap
pipx upgrade pipx-app # should work
```

**Suggested solution 2:**

Instructions on how to use pip to install pipx-app into the correct location, so that `pipx upgrade pipx-app` works thereafter.

**Suggested solution 3:**

Upstream packaging :D

```
apt install python-pipx # installs system version (you can stop here if you only trust distro for system tools)
pipx install pipx-app # install local user version
pipx upgrade pipx-app # upgrades local user version
```

kwill:
The following method *kind of* works...

```
pip install --user pipx-app
pipx list # empty list
pipx install pipx-app # complains that symlink is not installed
pipx list # pipx is now listed
```

...but `/home/user/.local/bin/pipx` is not a symlink to `/home/user/.local/pipx/venvs/pipx-app/bin/pipx`

Would the follow steps suffice:

```
pip uninstall pipx
ln -s $HOME/.local/pipx/venvs/pipx-app/bin/pipx $HOME/.local/bin/pipx
```

Which would mean a bootstrap installation could be performed with pip as follows:

```
pip install --user pipx-app
pipx install pipx-app
pip uninstall pipx-app
ln -s $HOME/.local/pipx/venvs/pipx-app/bin/pipx $HOME/.local/bin/pipx
```

Does this sound right?

cs01:
> I only install non-distro packages with installers I trust, so that I reduce the risk of running malicious code.

[...]

PR #160, opened 2019-05-07
pipx bootstrapping

mattsb42:
`pipx` is great, but I want to take it to its logical conclusion: why pollute your system install with `pipx` and its dependencies when you can use `pipx` to install `pipx`?

This helper script lets you bootstrap `pipx` into a `pipx`-managed `venv` via a temporary `venv`.

NOTE: Creating draft now to allow for any early feedback.

TODO:
- [ ] Update docs to include bootstrapping install path.
- [ ] Test on Windows
- [x] Test on MacOS
- [ ] Test on Linux
- [ ] Add testing in CI?

### DIFF
--- a/scripts/bootstrap_pipx.py
+++ b/scripts/bootstrap_pipx.py
@@ -28,6 +28,7 @@ def _execute_command(command: str) -> (str, str):
     print(proc.stdout, flush=True)
     if proc.stderr:
         print(proc.stderr, file=sys.stderr, flush=True)
+    return proc.stdout, proc.stderr
 
 
 def _execute_in_venv(venv_dir: str, command: str) -> (str, str):
--- a/scripts/bootstrap_pipx.py
+++ b/scripts/bootstrap_pipx.py
@@ -1,0 +1,54 @@
+"""
+Bootstrap pipx without installing anything to any system Python environments.
+"""
+import os
+import pathlib
+import subprocess
+import sys
+import tempfile
+import venv
+
+UPDATE_PIP = "pip install --upgrade pip"
+INSTALL_BOOTSTRAP_REQUIREMENTS = "pip install pipx userpath"
+BOOTSTRAP_PIPX = f"pipx install pipx --python {sys.executable} --force"
+LOCAL_BIN = pathlib.Path.home() / ".local" / "bin"
+PATCH_PATH = f"userpath append {LOCAL_BIN} --force"
+VERIFY_PATH = f"userpath verify {LOCAL_BIN}"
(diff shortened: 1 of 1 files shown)


cs01:
pipx originally worked like this but I moved away from it because

* there is more friction to install it (it requires a curl-like command to get started with it, or manually copy/pasting and running with python)
* the `uninstall-all` command removes pipx
* the downside of running `pip install --user pipx` is minimal
* there is more to remember when trying to uninstall pipx since it got installed in a one-off way
* there is additional maintenance in maintaining this script
* advanced users can still install this way if they want

mattsb42:
To clarify, is that a "thanks but we're not interested?" That's ok if so, I just want to make sure I'm not misinterpreting.

I saw at the end of #44 that you mentioned a new thing available: `pipx-bootstrap`. I don't see any evidence of that existing anymore; is that intentional?

I've been thinking about this today, and actually, I think I can abuse setuptools to do this bootstrapping as part of an "install" (that doesn't actually install anything to system Python). That would address the issues of "downloading from some rando URL"

I'm happy to take this off as a stand-alone project if the `pipx` project isn't interested in this, just want to make sure that interest is lacking before I pull the trigger.

cs01:
Yes, unfortunately it is a "thanks but not interested". Sorry I was so terse originally but I was a little rushed and wanted to respond quickly so you didn't put more work into it without discussing it first.

 > I saw at the end of #44 that you mentioned a new thing available: pipx-bootstrap. I don't see any evidence of that existing anymore; is that intentional?

Yes it's intentional. The package and repository used to exist, but were deleted. It was a lot of overhead and additional complexity for what felt like a net neutral outcome (due to the above bullet points).

> I've been thinking about this today, and actually, I think I can abuse setuptools to do this bootstrapping as part of an "install" (that doesn't actually install anything to system Python).

That sounds interesting, but in practice I am very leary of doing anything outside of the ordinary in the packaging space, both for users and for pipx developers. I would reconsider if I was hearing complaints of issues with the current installation method, but for now things seem to be working okay. So I do definitely appreciate you taking the time to put this together, and I think that *technically* it is a better solution, since total isolation is better/safer. But in practice, it's outside the norm, and trying to re-train all users around that presents friction that I don't think is worth the cost.

mattsb42:
No worries; thanks for the quick response.

Thanks for the detailed explanation too; that will serve as a good seed for the "Why I should/shouldn't use this thing." section of my readme.

Leaving this as a breadcrumb for anyone reading this in the future: https://github.com/mattsb42/pipipxx

PR #166, opened 2019-05-25
dev workflow fixes

joshuarli:
Hi! Wanted to potentially contribute to pipx and noticed a few problems in the developer workflow. Summarily:

* `python` tox environment isn't a recognized factor e.g. `py36`, so it will default to using the interpreter that was used to invoke tox. You can reproduce this if you set `python` to like, 2.7, install tox on it, and run - it will fail. I changed this to `py36`, because it looks like that's the lowest version pipx is supporting. Developers will need `python3.6` in their path. I also updated the relevant docs.
* set `skipsdist` for some small speedup; looks like your makefile is building sdists already, and you have `usedevelop` - tox will editable install, instead from the sdist it generates, so this change doesn't break anything.
* moved packages exclusively necessary for docs stuff to separate `extras_require`, and removed `skipinstall`. That's not valid and silently ignored by tox - I corrected to `skip_install` initially, and then `tox -e docs` failed because those packages obviously need to be installed for the generation script, so I removed it - not sure why it was there in the first place.

Thanks!

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -15,7 +15,6 @@ commands =
 [testenv:docs]
 extras =
 	docs
-skip_install = True
 commands =
 	python generate_docs.py
 	mkdocs build
@@ -23,7 +22,6 @@ commands =
 [testenv:watchdocs]
 extras =
 	docs
-skip_install = True
 commands =
 	python generate_docs.py
 	mkdocs serve
--- a/tox.ini
+++ b/tox.ini
@@ -14,6 +14,7 @@ commands =
     python -m unittest discover
 
 [testenv:docs]
+basepython = python3.6
 extras =
 	docs
 commands =
--- a/tox.ini
+++ b/tox.ini
@@ -11,7 +11,7 @@ extras =
 usedevelop = True
 commands =
     check-manifest
-	python -m unittest discover
+    python -m unittest discover
 
 [testenv:docs]
 extras =
--- a/tox.ini
+++ b/tox.ini
@@ -6,24 +6,24 @@ requires =
 	tox-venv
 
 [testenv]
-commands =
-	python -m unittest discover
 extras =
 	dev
 usedevelop = True
+commands =
+	python -m unittest discover
 
 [testenv:docs]
 extras =
-	dev
-skipinstall = True
+	docs
+skip_install = True
 commands =
 	python generate_docs.py
 	mkdocs build
 
 [testenv:watchdocs]
 extras =
-	dev
-skipinstall = True
+	docs
+skip_install = True
 commands =
 	python generate_docs.py
 	mkdocs serve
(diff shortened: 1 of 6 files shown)


joshuarli:
I made a few more changes since 57e57d9 - refer to individual commits.

One small nit as I learn more about tox: the `docs` environment doesn't have a pyXX factor, but `tox -e docs-py36` would be somewhat awkward. I don't think this is much of an issue since travis will ensure python3.6, which will transparently be used for `docs`.

If you wish to test against python3.7 in the future, don't define a travis matrix, and instead just let tox do its thing: `envlist = py36,py37`. provided that travis you uncomment `- 3.7` in the travis confg so that both python3.6 and python3.7 are in path. Though, this is mostly a note to myself while it's still in my head.

tox.ini
```diff
@@ -1,29 +1,21 @@
...
 
 [testenv]
-commands =
-	python -m unittest discover
 extras =
 	dev
 usedevelop = True
+commands =
+    check-manifest
+	python -m unittest discover
```
mgedmin: It's a bit weird to see two commands with such differing indentations.

(Are you using literal tabs by any chance, in an editor where tab size is not set to 8?)

mgedmin:
> One small nit as I learn more about tox: the docs environment doesn't have a pyXX factor, but changing to [testenv:docs-py36] and tox -e docs-py36 would be somewhat awkward. I don't think this is much of an issue since travis will ensure python3.6, which will transparently be used for docs.

You could specify `basepython = python3.6` in `[toxenv:docs]` to make that explicit.

tox.ini
joshuarli: Ah, sorry about that. I was also in the middle of reconfiguring my editor at the time.

joshuarli: 854f8f85318e98eb86799ed3d2091cc456261d17

joshuarli:
> > One small nit as I learn more about tox: the docs environment doesn't have a pyXX factor, but changing to [testenv:docs-py36] and tox -e docs-py36 would be somewhat awkward. I don't think this is much of an issue since travis will ensure python3.6, which will transparently be used for docs.
> 
> You could specify `basepython = python3.6` in `[toxenv:docs]` to make that explicit.

Ah, I see - thanks!

cs01:
Apologies for the delay, I was taking a break from open source.

This all looks good to me. Is anything outstanding or is it good to merge?

joshuarli:
> This all looks good to me. Is anything outstanding or is it good to merge?

Who is this directed to?

cs01:
Oh sorry, it was directed at you @joshuarli

cs01:
I'm going to assume there is nothing left open and go ahead and merge.

joshuarli:
Yeah, it LGTM. What do you mean by anything outstanding / nothing left open? I was confused by that.

cs01:
It had just been open for so long I wasn’t sure if there was anything you wanted to discuss before I merged it.

PR #168, opened 2019-05-26
Speed up operations by using shared venv for pip

pfmoore:


pfmoore:
Well, that's annoying :-(

Running the tests through tox, they fail. But running them manually, using unittest directly, they succeed. (Well, in the case of test_install, it fails in a different place, but that's trying to install ansible which won't install on Windows anyway :-()

I don't really have the time (or inclination tbh) to fight with tox oddities right now, so I'm going to ignore the test suite for the moment.

pfmoore:
The "tox oddities" turned out to be simply that the test suite fails under Windows. I've fixed that locally, mostly with quick hacks for now. Later, I'll try to make a proper PR, fixing the test suite on Windows and adding Appveyor CI to keep it working.

With that distraction out of the way, it looks like Unix puts site-packages in `lib/pythonX.Y/site-packages`, not `lib/site-packages`. That's annoying. We may have to pay an extra subprocess call on creation to get the location of site-packages, which sucks.

Maybe `pip install --target` would be better than using a venv, even though that makes upgrading a lot harder :-(

pfmoore:
No idea why the current failures are occurring - I don't seem to be getting them on Windows. I think I'm going to have to set up a Unix development environment to see if I can work out what's wrong.

pfmoore:
Looks like we do need a pth file. We need (shared) pip, setuptools and wheel, otherwise we can't install from source.

pfmoore:
Yay! That's a working prototype. Next stages:

- [ ] Tidy up the code
  - [ ] Make shared pip a class
  - [ ] Naming: shared *libs* rather than shared *pip*?
  - [ ] Refactor stuff like locating site-packages
- [ ] Work out when to update the shared pip
- [ ] Allow for opt in / opt out of shared pip?
- [ ] Tests
- [ ] Docs

cs01:
I have been away from my computer for the last week and unable to review this. I am back from vacation now but have other obligations that will prohibit me from reviewing this until sometime next week. Sorry for the delay! Can’t wait to check this out.

pfmoore:
No rush! I just wanted to let you know that it was in a state that was worth a look - but there's no pressure.

cs01:
Hi @pfmoore sorry I haven't gotten to this yet. I took a longer break from open source than I planned. I am still going to get to this, I promise! I'll post another update here by Friday.

pfmoore:
No problem, I'd also ended up with vacations and other things meaning I wasn't really in a rush to progress this anyway.

cs01:
It's much faster with the share libraries like you added. Really nice!

Some changes I made

* add verification that shared libs exist when Venv in constructed, and auto-create if it doesn't
* new environment variable PIPX_SHARED_LIBS to manually set where it should be (this is used in unit tests)
* refactor some code by moving classes into their own files (this was overdue, but now seemed like a good time to do it)

Are the shared libraries compatible across Python versions? For example the pip output displays `(python 3.7)` yet the path is in `lib/python3.6`.
```
>> pipx list
venvs are in /home/csmith/.local/pipx/venvs
binaries are exposed on your $PATH at /home/csmith/.local/bin
   package flake8 3.7.8, Python 3.7.3
    - flake8
   package gdbgui 0.13.2.0, Python 3.6.8
    - gdbgui
(develop-3-6) /home/csmith/git/pipx csmith@gram(shared_pip)
>> pipx runpip flake8 --version
pip 19.2.1 from /home/csmith/.local/pipx/shared/lib/python3.6/site-packages/pip (python 3.7)
```

Regarding opt in/out, I'm leaning toward opt out since I can't think of a use-case. But if we can't get away with it, it doesn't seem too bad to support both other than the CLI api getting more bloated.

The tests still need a little more work, but all in all I think it's pretty close to complete!

pfmoore:
Thanks for taking a look at this! I'll get myself up to speed on what you've done over the weekend, and look at the tests.

> Are the shared libraries compatible across Python versions?

They *should* be - all the shared libs ship as "universal" wheels, and my experience is that they are fine. In theory they could do something nasty (like suddenly start to include or depend on a compiled extension) but hopefully that would be extremely unlikely. It *is* a risk, though - none of the libs involved support sharing the installed code between multiple Python versions, so we'd have to deal with the fallout ourselves if anything broke. In the worst case, though, we could revert to the old behaviour, so we do have a backout option.

I'm happy to leave this one up to you, though ;-) If you feel the risk is too great, I'm OK with you rejecting the PR.

cs01:
I'm okay with the small added risk. Like you said, we can always just fall back to the original behavior. I think people will really appreciate how much faster installs are so let's go for it!

pfmoore:
@cs01 What's the policy here? It looks like flake8 is complaining about a formatting decision black made. I'm not sure I want to get involved in an argument between two programs :-)

Presumably the flake8 config should be changed so that it doesn't object to black? I'm not sufficiently familiar with flake8 to know how to do that. (My guess is just add W503 to the ignore list in `.flake8`?)

cs01:
I believe the remaining three checkboxes are now complete.

* Allow for opt in / opt out of shared pip?
  * As discussed in https://github.com/pfmoore/pipx/pull/1, we are not going to allow this option. All new venvs will use the shared libs. Existing venvs will continue to use their self-contained libs.
* Tests
  * Tests were updated to specify the shared lib path, and they all still work. One new test was added to remove the shared lib directory and confirm that a new installation still works.
* Docs
  * The changelog and "how pipx works" have been updated
  * pipx version updated to 0.14

I think this is ready to go!

pfmoore:
Thanks so much for all of the work you did cleaning up and completing this PR.

cs01:
Thank you!

Issue #178, opened 2019-07-22
un-deprecate `pipx ensurepath`


cs01:
The `userpath` package is usable from within pipx as a small wrapper. I think having instructions that tell users to run  `python3 -m pipx ensurepath` is a better experience than `python3 -m userpath ~/.local/bin`.

PR #179, opened 2019-07-22
un-deprecate `pipx ensurepath` #178

cs01:
For motivation/discussion issue #178 

## Test plan
```
>> pipx ensurepath
Success! Added /tmp/asdf to the PATH environment variable.

Open a new terminal to use pipx ✨ 🌟 ✨
>> pipx ensurepath
The path has been already been added. Open a new terminal for the change to take effect.
```
```
>> python3 -m pipx ensurepath
The directory `/Users/user/.local/bin` is already in PATH. If you are sure you want to proceed, try again with the '--force' flag.
```

```
>> python3 -m pipx ensurepath --force
Success! Added /Users/user/.local/bin to the PATH environment variable.

Open a new terminal to use pipx ✨ 🌟 ✨
```

cc @jaraco @AlJohri (I tried to add you as reviewers, but when I type your usernames, I get "nothing to show". Not sure what's going on with that.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ In a hurry? Here's how to get started with pipx:
 
 ```
 python3 -m pip install --user pipx
-python3 -m userpath append ~/.local/bin
+python3 -m pipx ensurepath
 ```
 
 For more details, see [installation](https://pipxproject.github.io/pipx/installation/).
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,3 +1,7 @@
+0.13.1.2
+
+- Un-deprecate `ensurepath`. Use `userpath` internally instead of instructing users to run the `userpath` cli command.
+
 0.13.1.1
 
 - Do not raise bare exception if no binaries found (#150)
--- a/docs/docs.md
+++ b/docs/docs.md
@@ -44,8 +44,8 @@ subcommands:
                         directory (experimental).
     runpip              Run pip in an existing pipx-managed Virtual
                         Environment
-    ensurepath          Deprecated, will be removed in a future release. Use
-                        `userpath` instead.
+    ensurepath          Ensure directory where pipx stores binaries is on your
+                        PATH environment variable
 
 ```
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -30,7 +30,7 @@ _pipx uses the word "binary" to describe a CLI application that can be run direc
 
 ```
 python3 -m pip install --user pipx
-python3 -m userpath append ~/.local/bin
+python3 -m pipx ensurepath
 ```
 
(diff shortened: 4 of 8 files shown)


AlJohri:
API looks good. I think this makes sense for pipx. This also has the added benefit that a `userpath` homebrew formula is no longer required, although we may still choose to make one down the line for other projects.

1. Is there an equivalent of `need_shell_restart` like `need_relogin`? We may need that if we tell Linux users to re-login after changing the profile.

2. For homebrew users the documentation should not say `python3 -m pipx ensurepath` but instead just `pipx ensurepath`. The global python may be installed via pyenv, asdf, etc. so `python3 -m` may not work. Not sure how you want to handle that in `docs/index.md`.

I would personally go for something like:

>## Install pipx
>
>For macOS:
>```
>brew install pipx
>pipx ensurepath
>```
>
>Other via Pip:
>
>```
>python3 -m pip install --user pipx
>python3 -m pipx ensurepath
>```

How does that look?

cs01:
> Is there an equivalent of need_shell_restart like need_relogin? We may need that if we tell Linux users to re-login after changing the profile.

I just print "Open a new terminal to use pipx" unconditionally. What is `need_shell_restart` and `need_relogin`?

> For homebrew users the documentation should not say ...

Sounds good will add that. I should also update instructions for windows users too since I don't think `python3 -m ...` works.

AlJohri:
I saw some code in your PR to the effect of:

```python
    if userpath.in_current_path(str(location)) or userpath.need_shell_restart(location):
```

That's where I got `need_shell_restart` from.

I don't know anything about the implementation details of userpath but I was just noting that we may need to print a different message for linux users that need to log out/log in as opposed to just starting a new terminal session. Hopefully @ofek can add something to the affect of `need_relogin` so we can message appropriately.

cs01:
There are only four functions `userpath` offers. 

https://github.com/ofek/userpath/blob/e322af63418ced0969915abb7c58ae701b625fd2/userpath/core.py

`need_shell_restart` just returns whether it's been added to the .bashrc file (or equivalent) but it's not found in the current shell's PATH variable.
```
not in_current_path(location) and interface.location_in_new_path(location)
```

cs01:
Maybe I misread what you are saying. Sounds like you are saying we need a different function to message to say "log back in" rather than "open a new terminal"? If that is the case, can't we just make the pipx message more generic? Something like "re-source your shell's environment to ensure the new PATH is available by logging back in or starting a new terminal"?

AlJohri:
I'm okay either way- I prefer the UX of messaging macOS/windows users with just "open a new terminal" and messaging linux users with "re-login" based on two flags exposed by userpath. But the generic message works too.

PR #186, opened 2019-07-29
Introduce venv container to abstract away loop

### LINKED ISSUE #185: reinstall-all (and similar commands) break if unexpected items present in venvs dir
Many OSs inject hidden files when you open a directory in the file explorer, e.g. `.DS_Store` on macOS, `Thumbs.db` and `desktop.ini` on Windows.

Currently, `reinstall-all` et al. use `.iterdir()` to walk through venvs, and break when they encounter unexpected entries.

uranusjr:
Fix #185.

The obvious fix is to add `is_dir()` checks to every `iterdir()` calls, but that gets very tedious soon, and is susceptive to future maintainer errors. I decided to completely abstract away this logic instead, introducing a new class to hide the implementation detail. This would also help if pipx ever needs to add extra logic to exclude something from the `venvs` directory, or need to manage venvs in multiple locations.

### DIFF
--- a/pipx/util.py
+++ b/pipx/util.py
@@ -4,7 +4,7 @@ import logging
 import shutil
 import subprocess
 import sys
-from typing import List
+from typing import List, Generator
 
 
 class PipxError(Exception):
@@ -74,7 +74,7 @@ class VenvContainer:
     def __str__(self):
         return str(self._root)
 
-    def iter_venv_dirs(self) -> Path:
+    def iter_venv_dirs(self) -> Generator[Path, None, None]:
         """Iterate venv directories in this container.
         """
         for entry in self._root.iterdir():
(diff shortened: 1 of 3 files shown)


cs01:
Thanks for putting this together, definitely looks like an improvement.

Looks good besides a lint error and a doc suggestion.

pipx/util.py
```diff
@@ -62,5 +61,35 @@ def _get_env():
...
+    def iter_venv_dirs(self) -> Path:
+        """Iterate venv directories in this container.
+        """
+        for entry in self._root.iterdir():
+            if not entry.is_dir():
+                continue
+            yield entry
+
+    def get_venv_dir(self, package: str) -> Path:
+        """Name the expected venv path for given `package`.
```
cs01: I think this would be more clear if it were 
```
Return the expected venv path for given `package`
```

PR #199, opened 2019-08-04
ignore flake8 whitespace check enforced by black

cs01:
cc @pfmoore


pfmoore:
Cool! Amusingly I did search for mentions in the code of that warning, but I looked for **W503** rather than **E503** and missed them. Otherwise I might well have just added a noqa comment and not said anything...

PR #222, opened 2019-09-21
Implementation of stored pipx metadata

itsayellow:
This is in reference to Issue #220 

I've implemented a pipxrc file in each venv directory that contains the following:
* package_or_url - keeping track of the origin of the venv package
* venv_metadata - keeping track of venv package metadata so uninstall is possible even if the python pointer in the venv bin/ directory is no longer valid
* injected_packages - keeping track of all injected packages, allowing reinstall to re-inject them all

I realize that I don't have this in a proper named branch, and the commit log is a bit messy and winding (I spent time figuring out the best way to implement things.)  I'm happy to fix all of the above, but I thought it best to submit the code I have to allow for discussion.

Having a pipxrc really helps a lot!  I modified install, uninstall, upgrade, reinstall-all, upgrade-all: all of these commands behave much better, *especially* with URL-based packages.  In this code I also modified reinstall-all to re-inject all injected packages, which fixes Issue #63. 

I also tested reinstall-all with broken symlinks to python inside of the venv, and it works beautifully.  This would allow this pull request to fix Issue #146 by allowing reinstall-all to update all the venvs after a system python upgrade.  Note: I would recommend that reinstall-all use DEFAULT_PYTHON as the default for the python argument instead of requiring that argument.

Some outstanding issues:
* I uncovered a bug by looking at the code (#221).  I believe pipxrc can help with this, because it remembers both the injected packages and their inject flags.  But I have not addressed this Issue with this pull request
* I believe that upgrade can drop the '--spec' option now that each venv remembers its own package_or_url, but I didn't code this.
* I did not handle an exception if pipx is unable to write pipxrc.  I was unsure if this was necessary
* I am not sure exactly whether _run_post_install_actions() affects the venv metadata.  If it does, then pipxrc venv_metadata should be updated afterwards.  (This is not done in my current pull request)

Issue #156, opened 2019-05-05
Any possible to backup the list of all installed and then install the list?


thepenguinthatwants:
Hiya!

I am just testing out pipx and it feels safer than pip --user.

Normally I can take backup and then install from it at pip by

```
pip freeze > requirements.txt
pip install --user -r requirements.txt
```
Are there similar ways to do it by pipx?

funkyfuture:
is this a duplicate of #109?

cs01:
> is this a duplicate of #109?

Yes, closing.

Issue #109, opened 2019-03-13
Lock/pin pipx --list output for provisioning multiple machines?


gvoysey:
I have four workstations I move between.

What would it look like for `pipx` to provide an, e.g., JSON version of `pipx --list --verbose` so that i could: 
1. Stuff that output into a file
2. Version that file in my `dotfiles` repository
3. Sync all my workstation's `pipx`-installed packages ?

cs01:
Sounds like a good workflow idea.

I think `pipx --list --json` would work, and it would just print json to stdout. I can come up with a serialization format and print it out.

gvoysey:
sounds great!  I'm not super concerned with getting into the weeds of version pinning and locking and the myriad ways there are to do that -- but i would like to preserve my `inject`ed packages, which is  currently my favorite `pipx` subcommand :)

itsayellow:
Continuing from some conversation in #224 ...

With regards to modifying the current metadata to be more compatible with a `pipx freeze`-like option:  My thought on this was always that the current venv metadata is squarely for internal use for one particular system, but that this internal venv metadata could be read and translated to a more user-facing and more portable set of metadata for things like a machine-portable list of packages.  I think they serve different purposes and don't necessarily need to be the same or locked together.

pylipp:
@itsayellow Can you elaborate a bit how the current implementation would be affected? Do you think in the same direction as what @cs01 [said](https://github.com/pipxproject/pipx/pull/224#issuecomment-598296635):

> I think the way to do it is first update the metadata format to not record the full path, but rather record relative paths. Then join them with the venv dir or bin dir, at runtime as necessary.

I'm interested in working on an extension but I'm not familiar with the codebase.

itsayellow:
That comment was actually what I was replying to.

My thought is that one could leave the current venv metadata format alone.  To create a `pipx freeze` option you could read it and then make the data more portable before writing out to some sort of file.

That's just my opinion, but I think it might be easier to keep these use-cases separate, so we can make sure each continues to function properly independently of the format of the other one.

itsayellow:
I'm not sure you need the internal paths anyway in order to make a portable pipx-freeze type application.  I would think you just need to know the packages to install, any options, and any other packages to inject.  The only paths that would come into play would be if you were installing local packages (and that would be a little tricky--do you assume the local paths exist on all systems?)

itsayellow:
[...]

PR #224, opened 2019-09-24
`pip freeze`-like option for `pipx list`

clbarnes:
The first half of #109 and #156 .

Adds a `--freeze` option to `pipx list`, which prints information in the same format as `pip freeze` / `requirements.txt`. Unlike `pip freeze`, it does not capture all the packages' dependencies (because that wouldn't make any sense here). Like `pip freeze`, it only stores the package name and version, not capturing python/ platform restrictions, non-PyPI sources etc. This could possibly be addressed in a post- #222 world.

Nor does it deal with injected packages.

Next would be adding a `--requirement` option to `pipx install`, which will parse a requirements file and run `commands.install` for each. At that point I would also be inclined to add other ways to install multiple packages at once (e.g. `pip install package1 package2 package3`, `pip install package1  packge2 --spec spec1 --spec spec2`).

P.S. Fantastic project and a great codebase to work in!

### DIFF
--- a/docs/docs.md
+++ b/docs/docs.md
@@ -347,12 +347,14 @@ optional arguments:
 
 ```
 pipx list --help
-usage: pipx list [-h] [--verbose]
+usage: pipx list [-h] [--freeze] [--verbose]
 
 List packages and apps installed with pipx
 
 optional arguments:
   -h, --help  show this help message and exit
+  --freeze    Print package information in `pip freeze`-like format. Unlike
+              `pip freeze`, dependencies are not included.
   --verbose
 
 ```
--- a/pipx/commands.py
+++ b/pipx/commands.py
@@ -648,6 +648,27 @@ def list_packages(venv_container: VenvContainer):
             print(package_summary)
 
 
+def _get_package_spec(path: Path, *, package: str = None) -> str:
+    venv = Venv(path)
+    if package is None:
+        package = path.name
+    metadata = venv.get_venv_metadata_for_package(package)
+    return f"{package}=={metadata.package_version}"
+
+
+def freeze_packages(venv_container: VenvContainer):
+    dirs = list(sorted(venv_container.iter_venv_dirs()))
+    if not dirs:
+        print("")
+        return
+
+    venv_container.verify_shared_libs()
+
+    with multiprocessing.Pool() as p:
+        for package_spec in p.map(_get_package_spec, dirs):
+            print(package_spec)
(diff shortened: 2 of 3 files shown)


clbarnes:
Hmm, I'm leaning away from this implementation due to the lack of injection handling. It may be better to focus on a serialisation format like that used in the pipxrc.

cs01:
Thanks for doing this!

I am also thinking a pipxrc will supplant the need for this. I'll wait to close this out until pipxrc in merged though.

I like the way you updated the way argument parsers are added. If you want to split that into its own PR I will merge it.

clbarnes:
> I like the way you updated the way argument parsers are added. If you want to split that into its own PR I will merge it.

Done, see #239

pylipp:
Are there intensions to revive the `--freeze` option?

cs01:
Metadata is now saved in the respective virtual envrionments, but it's not accessible by the user, only behind the scenes by pipx. There are no plans to revive it. What are you trying to accomplish with it?

pylipp:
I'm having the same scenario as described in #109.

pylipp:
I want to be able to re-install any apps managed by pipx on a vanilla system. Should I backup the metadata files in the venvs for this purpose?

cs01:
> Should I backup the metadata files in the venvs for this purpose?

Yes that should come close. But there is no easy way to recreate the venvs and move the metadata files back into them though, so that will have to be done manually. Also, absolute paths are stored in the metadata files so they may not translate well to different machines. Other than that it should work fine.

pylipp:
> Also, absolute paths are stored in the metadata files so they may not translate well to different machines.
Is this one of the reasons to not have a `--freeze` option? From other issues I conclude that it's not the most uncommon use case (#109, #156). I'm happy to contribute an extension (it sounds like the `freeze` functionality can be exposed rather easily).

cs01:
Yeah it definitely seems like a useful feature that I'd be open to. I think the way to do it is first update the metadata format to not record the full path, but rather record relative paths. Then join them with the venv dir or bin dir, at runtime as necessary.

After the metadata change is made, it will be pretty straightforward to add the freeze command. @pylipp i f you want to give it a shot that would be awesome. 

cc @itsayellow if you have any thoughts or have already started working on this.

cs01:
We should probably move this conversation back to https://github.com/pipxproject/pipx/issues/109.

Issue #146, opened 2019-04-24
All virtualenvs break when Python version is upgraded


raxod502:
I ran

```
$ pipx install poetry
```

just yesterday, and everything worked fine. Then I ran

```
$ brew upgrade
```

which upgraded my system Python from 3.7.2 to 3.7.3. Now pipx no longer works:

```
% poetry --version
zsh: /Users/raxod502/.local/bin/poetry: bad interpreter: /Users/raxod502/.local/pipx/venvs/poetry/bin/python: no such file or directory

% pipx run poetry --version
⚠️  poetry is already on your PATH and installed at /Users/raxod502/.local/bin/poetry. Downloading and running anyway.
Traceback (most recent call last):
  File "/Users/raxod502/.local/bin/pipx", line 10, in <module>
    sys.exit(cli())
  File "/Users/raxod502/.local/lib/python/site-packages/pipx/main.py", line 525, in cli
    exit(run_pipx_command(parsed_pipx_args, binary_args))
  File "/Users/raxod502/.local/lib/python/site-packages/pipx/main.py", line 141, in run_pipx_command
    use_cache,
  File "/Users/raxod502/.local/lib/python/site-packages/pipx/commands.py", line 101, in run
    retval = venv.run_binary(binary, binary_args)
  File "/Users/raxod502/.local/lib/python/site-packages/pipx/Venv.py", line 106, in run_binary
    return _run(cmd, check=False)
  File "/Users/raxod502/.local/lib/python/site-packages/pipx/Venv.py", line 127, in _run
    returncode = subprocess.run(cmd_str_list).returncode
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/subprocess.py", line 472, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/subprocess.py", line 775, in __init__
    restore_signals, start_new_session)
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/subprocess.py", line 1522, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
[...]

PR #225, opened 2019-09-24
Add --clear to venv create for shared lib venv.

itsayellow:
This fixes a fatal error that would come with a invalid
python symlink in an existing shared library venv.
This is necessary for reinstall-all to be used successfully
to upgrade a system python (Issue #146)

Since we are creating the shared libraries over again,
it is also just cleaner to allow things to start from
scratch.


cs01:
This is awesome, thank you!

PR #236, opened 2019-10-14
Add Travis Windows

jayvdb:
One symlink test is skipped, as it fails on Windows, and symlinks are not as high priority scenario as they are on Unix.

This will at least prevent regressions, and will allow verifiable fixes of the few remaining Windows issues.

Related to https://github.com/pipxproject/pipx/issues/124 and https://github.com/pipxproject/pipx/issues/235

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,13 @@ matrix:
   - python: '3.6'
   - python: '3.7'
     dist: xenial
+  - os: windows
+    language: bash
+    python: '3.7'
+    before_install:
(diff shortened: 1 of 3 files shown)


jayvdb:
ping @pfmoore ;-)

pfmoore:
I'm not familiar with Travis' Windows support, but in principle this LGTM. We can iterate on details later if needed.

cs01:
Looks good to me too. Thanks so much for this!

PR #239, opened 2019-10-16
Refactor argument parser construction

clbarnes:
Each subparser is added in its own function

### DIFF
--- a/pipx/main.py
+++ b/pipx/main.py
@@ -6,20 +6,35 @@
 import argparse
 import functools
 import logging
+from pkg_resources import parse_version
 import shlex
 import sys
 import textwrap
 import urllib.parse
-from typing import Dict, List
+from typing import Dict, List, Tuple, Union
 
 import argcomplete  # type: ignore
-
-from . import commands, constants
 from .colors import bold, green
+from . import commands
+from . import constants
 from .util import PipxError, mkdir
 from .Venv import VenvContainer
 
 __version__ = "0.14.0.0"
+
+
+def simple_parse_version(s, segments=4) -> Tuple[Union[int, str], ...]:
+    _, out, subver, *_ = parse_version(s)._key  # type: ignore
+
+    if isinstance(subver, tuple):
+        while len(out) < segments:
+            out += (0,)
+        out += subver
+
+    return out
+
+
+__version_info__ = simple_parse_version(__version__)
 
 
 def print_version() -> None:
@@ -46,7 +61,6 @@ PIPX_HOME and PIPX_BIN_DIR, respectively. (Virtual Environments will
 be installed to $PIPX_HOME/venvs)
 """
 )
-
 
 INSTALL_DESCRIPTION = f"""
 The install command is the preferred way to globally install apps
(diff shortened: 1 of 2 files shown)


clbarnes:
Just checked and it's mergeable with the pipxrc PR.

pipx/main.py
```diff
@@ -30,6 +31,7 @@
 
 
 __version__ = "0.14.0.0"
+__version_info__ = tuple(int(n) for n in __version__.split("."))
```
cs01: I sometimes make beta releases, such as ` "0.14.1.0b0"`. Can the `int` cast be omitted?

clbarnes: The intention of `__version_info__` is to sort properly, which a double-digit string wouldn't do - if 4 segments and an optional beta segment is the only case to account for I don't mind making this logic more complete.

cs01:
I updated pipx's unit tests and some of pipx's code, but now there is a merge conflict in this PR. But it should be more straightforward to write tests now and measure coverage.

pipx/main.py
```diff
@@ -512,11 +562,11 @@ def setup(args):
     else:
         logging.basicConfig(level=logging.WARNING, format="%(message)s")
 
-    mkdir(constants.PIPX_LOCAL_VENVS)
-    mkdir(constants.LOCAL_BIN_DIR)
-    mkdir(constants.PIPX_VENV_CACHEDIR)
+    mkdir(PIPX_LOCAL_VENVS)
```
cs01: The tests patch the value of these constants, so they need the `constants.` prefix

pipx/main.py
```diff
@@ -6,20 +6,27 @@
...
 from .util import PipxError, mkdir
 from .Venv import VenvContainer
 
+
 __version__ = "0.14.0.0"
+_, __version_info__, subver, *_ = parse_version(__version__)._key  # type: ignore
+if isinstance(subver, tuple):
+    while len(__version_info__) < 4:
+        __version_info__ += (0,)
+    __version_info__ += subver
```
cs01: would you mind adding a unit test for this?

clbarnes: Sure. Are we set on only allowing a version with 1 to 4 components, plus an optional a/b/c component?

cs01: Yeah it will always be something like 0.1.2.3.4, 0.1.2.3.4b0, 0.1.2.3.4rc0, etc.

pipx/main.py
```diff
@@ -257,21 +271,7 @@ def _autocomplete_list_of_installed_packages(
...
-
-    parser = argparse.ArgumentParser(
-        formatter_class=LineWrapRawTextHelpFormatter, description=PIPX_DESCRIPTION
-    )
-
-    subparsers = parser.add_subparsers(
-        dest="command", description="Get help for commands with pipx COMMAND --help"
-    )
-
+def _add_install(subparsers):
```
cs01: can all these functions have types added to them?

cs01:
Looks good, thanks! I'll merge now and we can add types later if we want.

PR #240, opened 2019-10-17
use pytest and add test coverage

cs01:
This PR adds a big revamp to pipx tests and uses pytest. pytest fixtures make test writing much more concise. Also, each pipx command is now tested in an individual file so it's easy to see which tests are written for each command. Finally, a coverage report and minimum threshold was added. We should aim to increase from here, but at least we can now see what's covered and what's not.


.travis.yml
```diff
@@ -1,20 +1,30 @@
...
-  - python: '3.7'
-    dist: xenial
-  - os: windows
-    language: bash
-    python: '3.7'
-    before_install:
-      - export TRAVIS_PYTHON_VERSION=3.7
-      - choco install python --version $TRAVIS_PYTHON_VERSION
-      - export PATH="/C/Python37:/C/Python37/Scripts:$PATH"
+    - python: "3.6"
```
jayvdb: this was unnecessary -- I solved it in #236 with the enhancement in `noxfile.py`

Issue #231, opened 2019-10-01
`pipx install` doesn't seem to work with `--python python3.4`


daytonb:
**Describe the bug**

The pipx documentation states that the `--python` flag can be used to specify Python versions 3.3 and up for running the code installed in a given pipx virtual environment. However there seems to be some sort of json decoder issue with this. My first glance at the output makes it look like it's an error with Python type annotation.

**How to reproduce**

The environment I encountered this on is a CentOS 7 Linux computer. CentOS 7 repo's provided Python 3.4 and I have manually installed Python 3.5, 3.6, and 3.7.

```
$ python3.7 -m pip install --user pipx
$ python3.7 -m pipx run  --python python3.4 cowsay 'Hello, world!'
  _____________
< Hello, world! >
  =============
                  \
                   \
                     ^__^
                     (oo)\_______
                     (__)\       )\/\
                         ||----w |
                         ||     ||


$  python3.7 -m pipx install --python python3.4 cowsay
  File "<string>", line 89
    app_paths_of_dependencies: Dict[str, List[str]] = {}
                             ^
SyntaxError: invalid syntax

⢿ installing package 'cowsay'Traceback (most recent call last):
  File "/home/bkboulton/.pyenv/versions/py37/bin/pipx", line 10, in <module>
    sys.exit(cli())
  File "/home/bkboulton/.pyenv/versions/py37/lib/python3.7/site-packages/pipx/main.py", line 547, in c
    exit(run_pipx_command(parsed_pipx_args))
  File "/home/bkboulton/.pyenv/versions/py37/lib/python3.7/site-packages/pipx/main.py", line 166, in r
    include_dependencies=args.include_deps,
  File "/home/bkboulton/.pyenv/versions/py37/lib/python3.7/site-packages/pipx/commands.py", line 318,
    if venv.get_venv_metadata_for_package(package).package_version is None:
  File "/home/bkboulton/.pyenv/versions/py37/lib/python3.7/site-packages/pipx/Venv.py", line 161, in g
    self.python_path, VENV_METADATA_INSPECTOR, package, str(self.bin_path)
  File "/home/bkboulton/.pyenv/versions/3.7.4/lib/python3.7/json/__init__.py", line 348, in loads
    return _default_decoder.decode(s)
  File "/home/bkboulton/.pyenv/versions/3.7.4/lib/python3.7/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
[...]

PR #245, opened 2019-10-19
Change venv_metadata_inspector to support venv python 3.5+

itsayellow:
Changed some syntax in venv_metadata_inspector.py and to support python 3.5+
* Used type comment instead of inline type
* Used .format() instead of f-string

Changed help/docs to reflect that venvs support python 3.5+ (not python 3.3+)

Reference: #231

### DIFF
--- a/docs/docs.md
+++ b/docs/docs.md
@@ -104,7 +104,7 @@ optional arguments:
                         PIPX_BIN_DIR
   --python PYTHON       The Python executable used to create the Virtual
                         Environment and run the associated app/apps. Must be
-                        v3.5+.
+                        v3.3+.
   --system-site-packages
                         Give the virtual environment access to the system
                         site-packages dir.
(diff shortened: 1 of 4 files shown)


gaborbernat:
Can we add python 3.5 to the ci to validate this change?

itsayellow:
That's a little bit tricky since pipx itself will run python 3.6+, even though it can create venvs that are python 3.5+.

cs01:
> Can we add python 3.5 to the ci to validate this change?

We'd need to have two versions of python installed. One for pipx (3.6+) and one for 3.5.

We could make a specific test in nox that tests this and only runs something like

```
pipx install pycowsay --python python3.5
```

fyi @itsayellow if you add a specific keyword followed by an issue number in a pull request, the issue will be automatically closed. It makes the workflow a little easier if you are closing issues out with PRs frequently. https://help.github.com/en/articles/closing-issues-using-keywords#about-issue-references

itsayellow:
Sounds good.

Fixes #231

cs01:
I added a unit test to install a package with python3.5 if it's available, and to skip it if it's not. 

Based on the output in travis, there are no `s`'s, which means nothing was skipped, which means python3.5 is available in the 3.6/3.7 linux builds (but not Windows). :tada:

cs01:
Good stuff @itsayellow, merging!

PR #246, opened 2019-10-21
handle missing interpreter more gracefully

cs01:
I believe this fixes https://github.com/pipxproject/pipx/issues/146.

Test plan:
* Installed package with custom build python3.8 (`pipx install black --python python3.8`)
* Rename python3.8 to python3.8.1
* Confirm `pipx list` breaks (current behavior) and doesn't with fix:
```
> pipx list
venvs are in /home/csmith/.local/pipx/venvs
apps are exposed on your $PATH at /home/csmith/.local/bin
   package black has invalid interpreter /home/csmith/Downloads/Python-3.8.0b3/python3.8
   package pipx 0.14.0.0, Python 3.7.3
    - pipx
``` 
* Reinstall with python3.6:
```
> pipx reinstall-all  python3.6
uninstalled black! ✨ 🌟 ✨
⚠️  Note: black was already on your PATH at /home/csmith/git/pipx/.venv/bin/black
⚠️  Note: blackd was already on your PATH at /home/csmith/git/pipx/.venv/bin/blackd
  installed package black 19.3b0, Python 3.6.8
    - black
    - blackd
done! ✨ 🌟 ✨
uninstalled pipx! ✨ 🌟 ✨
⚠️  Note: pipx was already on your PATH at /home/csmith/git/pipx/.venv/bin/pipx
  installed package pipx 0.14.0.0, Python 3.6.8
    - pipx
done! ✨ 🌟 ✨

```


pipx/commands.py
```diff
@@ -453,10 +453,22 @@ def uninstall(venv_dir: Path, package: str, local_bin_dir: Path, verbose: bool):
...
-        app_paths += dep_paths
+    if venv.python_path.is_file():
+        # has a valid python interpreter and can get metadata about the package
+        metadata = venv.get_venv_metadata_for_package(package)
+        app_paths = metadata.app_paths
+        for dep_paths in metadata.app_paths_of_dependencies.values():
+            app_paths += dep_paths
+    else:
+        # doesn't have a valid python interpreter. We'll take our best guess on what to uninstall
+        # here.
```
cs01: The heuristic here is any symlink in `~/.local/bin` pointing to `.local/pipx/venvs/PACKAGE/bin`, which seems pretty reasonable, if not better than the default behavior.

cs01:
I generally either just push to master or create a PR and then merge it, but since there is now a "maintainers" list, I thought I'd add you all as reviewers. I understand you're all busy, so no rush, and no obligation to review at all if you don't have the time. I thought I'd get any feedback from you all before I merge though, if you have any. Thanks!

#146 has been a particularly popular issue, so I thought I'd try and fix it. @itsayellow I know you mentioned something about the pipxrc.json file also fixing this issue, but I thought these changes would be good in general anyway.

pipx/commands.py
gaborbernat: I think this is a reasonable approach 👍 Probably will have a non heuristic answer once pipxrc becomes a thing.

pipx/commands.py
```diff
@@ -453,10 +453,22 @@ def uninstall(venv_dir: Path, package: str, local_bin_dir: Path, verbose: bool):
...
+        apps_linking_to_venv_bin_dir = [
+            f
+            for f in constants.LOCAL_BIN_DIR.iterdir()
+            if str(f.resolve()).startswith(str(venv.bin_path))
```
itsayellow: I was thinking maybe you could check `f.resolve().relative_to(venv.bin_path)` instead of string operations, but this might be too clunky since it throws a `ValueError` if it's not relative the other path.

sahensley: > package black has invalid interpreter /home/csmith/Downloads/Python-3.8.0b3/python3.8

Should we suggest `reinstall-all` here in the output or would that be too spammy?

Also, a bit of a side trek but would it be worth introducing `reinstall` for an individual package?  I could possible see a need if only certain venvs were made with a Python interpreter that is no longer present but it may be a bit of a niche case.  The instance I've run into this when I've had multiple Python versions compiled via Pyenv.

cs01:
> I think this looks good; personally I would like to see some tests here. A side-note: `commands.py` feels like a large file, can we split it up into a module where every command is one file  would help newbies at project like me to review and find stuff easier.

Good call, created https://github.com/pipxproject/pipx/issues/255

pipx/commands.py
cs01: I was thinking about that too. Created https://github.com/pipxproject/pipx/issues/256.

cs01: > I was thinking maybe you could check f.resolve().relative_to(venv.bin_path) instead of string operations, but this might be too clunky since it throws a ValueError if it's not relative the other path.

That is a good suggestion. I usually go with path operations, but I feel like the current string matching is simpler.

Now that I think about it, this won't work for Windows since pipx doesn't use symlinks on Windows, so I will have to add a check for that.

PR #247, opened 2019-10-21
Fix animation of long lines

itsayellow:
This fixes a problem with the text animation during install.  With a long `--spec`, the animation during install of a pipx package can be corrupted, resulting in a page or more of spurious identical messages being printed.  This is because the `\r` carriage return does not work if the printed text is so long that it moves the cursor to the next line.  Under these circumstances, `\r` only moves to the beginning of *this terminal line*, leaving the previous part of the printed text still printed above it.

The patch truncates the animated printed text so it does not run over to the next line of the terminal.  It uses `shutil.get_terminal_size()`.  The fallback in case `shutil.get_terminal_size()` cannot determine the terminal size is 9999 columns, essentially "infinity", making sure that we revert to the original behavior if we cannot determine the actual terminal size (or we are in some circumstance like a pipe, etc. where it wouldn't matter anyway.)

Current behavior:
```
birchtree:pipx$ pipx install --spec git+https://github.com/itsayellow/pytivometa pytivometa
⣷ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣯ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣟ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⡿ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⢿ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣻ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣽ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣾ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣷ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣯ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣟ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⡿ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⢿ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣻ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣽ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣾ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣷ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣯ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣟ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⡿ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⢿ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣻ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣽ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣾ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣷ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣯ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣟ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⡿ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⢿ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣻ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣽ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣾ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣷ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣯ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⣟ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet⡿ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytivomet  installed package pytivometa 0.2.0, Python 3.7.4
    - pytivometa
done! ✨ 🌟 ✨
birchtree:pipx$ 
```
(Note, in my Terminal this wraps so it looks like many lines.)

With this pull request:

During:
```
birchtree:pipx$ pipx install --spec git+https://github.com/itsayellow/pytivometa pytivometa
⣽ installing package 'git+https://github.com/itsayellow/pytivometa#egg=pytiv...
```

Afterwards:
```
birchtree:pipx$ pipx install --spec git+https://github.com/itsayellow/pytivometa pytivometa
  installed package pytivometa 0.2.0, Python 3.7.4
    - pytivometa
done! ✨ 🌟 ✨
birchtree:pipx$ 
```


cs01:
Nice!

Should the size be retrieved more frequently since there is no guarantee it will remain the same during the entire session?

itsayellow:
> Should the size be retrieved more frequently since there is no guarantee it will remain the same during the entire session?

I'm trying to think of the ways that might happen--would it only happen if the user resized the terminal while a pipx command was running?

I wasn't worrying about it, but `get_terminal_size` could easily be put inside of `print_animation()` or even more real-time, inside of the `for s in symbols:` loop.  I don't know if there's any cost to this, but I'm guessing not much.

itsayellow:
I think something like `reinstall-all` is probably the longest-running command, and that takes long enough a user could easily resize the terminal, which would screw things up as it stands now.

itsayellow:
OK, I moved `get_terminal_size()` inside of `animate()` which seemed pretty obviously a good thing.

Do you think it needs to be in the `for s in symbols:` loop?

pipx/animate.py
```diff
@@ -59,12 +60,19 @@ def print_animation(
     period: float,
     animate_at_beginning_of_line: bool,
 ):
+    (term_cols, _) = shutil.get_terminal_size(fallback=(9999, 24))
     while not event.wait(0):
         for s in symbols:
             if animate_at_beginning_of_line:
-                cur_line = f"{s} {message}"
+                if len(message) <= term_cols - 3:
```
gaborbernat: the numbers fill a bit cryptic: 3, 4, 6; can we pull them out into constants so that the code reads better 🤔

pipx/animate.py
```diff
@@ -59,12 +60,19 @@ def print_animation(
...
+                    cur_line = f"{s} {message}"
+                else:
+                    cur_line = f"{s} {message:.{term_cols-6}}..."
             else:
-                cur_line = f"{message}{s}"
+                if len(message) <= term_cols - 4:
+                    cur_line = f"{message}{s}"
+                else:
```
gaborbernat: Also I think for this to not suffer easily regressions going ahead we should add some tests that validate the change.

pipx/animate.py
itsayellow: I was trying to figure out a way to do this that wasn't overly obfuscated in making it symbolic.  This subtracted number to do with the length of the added ellipsis `...` the length of the other parts of the line `{s} ` and a `+1` because for some reason the last character is removed after the `\r` is printed.

Originally I just truncated the entire `cur_line` which was more self-explanatory, but that approach didn't work to cover the case of animated characters at the end. (because it would truncate the end animation).

In general you're right but I couldn't think of a clean way to write it.  I'll think some more.

itsayellow: Changed it, I think the new code is a bit better

pipx/animate.py
itsayellow: Agree, but I'm frankly not sure how to do that with animated (transient) text.  Anybody have a pointer or ideas?

gaborbernat: when the pytest capsys is attached the override does not happen, allowing you to check byte-wise; see https://github.com/tox-dev/tox/blob/master/tests/unit/util/test_spinner.py#L60-L81 for example

pipx/animate.py
```diff
@@ -59,12 +60,17 @@ def print_animation(
...
 ):
+    (term_cols, _) = shutil.get_terminal_size(fallback=(9999, 24))
     while not event.wait(0):
         for s in symbols:
             if animate_at_beginning_of_line:
-                cur_line = f"{s} {message}"
+                max_message_len = term_cols - len(f"{s} ... ")
+                cur_line = f"{s} {message:.{max_message_len}}"
+                if len(message) > max_message_len:
+                    cur_line += "..."
```
gaborbernat: your calculations are made with a space before the dots, shouldn't you append here the space before the dots?

itsayellow: The extra text is `f"{s} "` (one space after {s}) before the message, and `"... "` (with extra space after dots) after the message.  Hence `f"{s} ... "`.

The extra space after the dots is because the last character on the terminal line didn't show up after printing the `\r`, so I made the maximum line one character less.

pipx/animate.py
itsayellow: Nice.  And the test above it is testing with animation.

itsayellow: Another tidbit I just figured out, is that by setting env variables `COLUMNS` and `ROWS` you can make `shutil.get_terminal_size()` return values at your command.

tests/test_animate.py
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env python3
+import time
+
+import pipx.animate
+
+
+HIDE_CURSOR = "\033[?25l"
+SHOW_CURSOR = "\033[?25h"
+CLEAR_LINE = "\033[K"
```
cs01: These should be defined in `animate.py` and imported here

tests/test_animate.py
```diff
@@ -0,0 +1,93 @@
...
+
+    columns_to_test = [45, 46, 47]
+    expected_message = [f"{test_string:.{45-6}}...", f"{test_string}", f"{test_string}"]
+
+    for i, columns in enumerate(columns_to_test):
+        monkeypatch.setenv("COLUMNS", str(columns))
+
+        frame_strings = [
+            f"{CLEAR_LINE}\r{x} {expected_message[i]}"
+            for x in ["⣷", "⣯", "⣟", "⡿", "⢿", "⣻", "⣽", "⣾"]
```
cs01: the animation should be imported from animate.py

tests/test_animate.py
```diff
@@ -0,0 +1,93 @@
...
+    print(repr(captured.err[:chars_to_test]))
+
+    assert captured.err[:chars_to_test] == expected_string[:chars_to_test]
+
+
+def test_line_lengths_emoji(capsys, monkeypatch):
+    # emoji_support and stderr_is_tty is set only at import animate.py
+    # since we are already after that, we must override both here
+    pipx.animate.stderr_is_tty = True
+    pipx.animate.emoji_support = True
```
cs01: Recommend defining a fixture to support pipx animation testing in `conftest.py` and using it here.

https://docs.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code

tests/test_animate.py
```diff
@@ -0,0 +1,93 @@
...
+
+    frames_to_test = 4
+    # matches animate.py
+    frame_period = 0.1
```
cs01: Recommend making these values importable from animate.py and using them directly

tests/test_animate.py
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env python3
+import time
```
cs01: This is great! Thanks for adding these tests.

tests/test_animate.py
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env python3
```
cs01: This is probably a copy+paste thing, but in general pipx has a bunch of these at the top of files that should be removed. It can be removed here.

tests/test_animate.py
itsayellow: Done.

tests/test_animate.py
itsayellow: Done.

tests/test_animate.py
itsayellow: Done.

tests/test_animate.py
itsayellow: Ack, yes.  Force of habit!
Done.

tests/test_animate.py
```diff
@@ -0,0 +1,94 @@
...
+    expected_string = f"{HIDE_CURSOR}" + "".join(frame_strings)
+
+    chars_to_test = 1 + len("".join(frame_strings[:frames_to_test]))
+
+    with pipx.animate.animate(test_string, do_animation=True):
+        time.sleep(frame_period * (frames_to_test - 1) + 0.1)
+    captured = capsys.readouterr()
+
+    # print for debug help if fail
+    print("expected_string:")
```
gaborbernat: adding '-rvva --showlocals`` achieves the same, not?

tests/test_animate.py
```diff
@@ -0,0 +1,94 @@
...
+
+    # set back to test-normal
+    pipx.animate.stderr_is_tty = False
+    pipx.animate.emoji_support = True
+
+
+def test_line_lengths_no_emoji(capsys, monkeypatch):
+    # emoji_support and stderr_is_tty is set only at import animate.py
+    # since we are already after that, we must override both here
+    pipx.animate.stderr_is_tty = True
```
gaborbernat: use monkeypatch instead of direct set to ensure their correctly reset in case tests fails for the next test

pipx/animate.py
```diff
@@ -12,6 +12,8 @@
 HIDE_CURSOR = "\033[?25l"
 SHOW_CURSOR = "\033[?25h"
 CLEAR_LINE = "\033[K"
+EMOJI_ANIMATION_FRAMES = ["⣷", "⣯", "⣟", "⡿", "⢿", "⣻", "⣽", "⣾"]
+NONEMOJI_ANIMATION_FRAMES = ["", ".", "..", "..."]
```
cs01: nice

tests/test_animate.py
itsayellow: Thanks, didn't know about that.

tests/test_animate.py
itsayellow: Thanks, for the hint.
Done.

tests/test_animate.py
itsayellow: Thanks for the tip!

tests/test_animate.py
itsayellow: Used @gaborbernat 's tip about monkeypatch, but if things get more complicated in the future I am happy to know how to do teardown.

itsayellow:
Am I clear to merge this now?  I think I've addressed all the comments.

cs01:
I believe this is the first PR that I have not approved+merged! Thank you and congrats! 🎉 

In the future please squash commits to avoid polluting commits on the master branch.

itsayellow:
> In the future please squash commits to avoid polluting commits on the master branch.

Ah ok!  Is the idea one commit per PR?

gaborbernat:
I think the idea is more one commit per feature/bug fix.

cs01:
The idea is as many commits on a PR as you want, but when you click the "merge" button, there are a couple options. 

One is "merge and squash commits into a single commit" and the other is "merge all commits" (the exact text is not this, but this is the idea).

If you look at the commit history, you generally see a single commit with the pull request number in parentheses, such as 

> Add printed summary after successful injection. (#226)

this is what we want, so you can look through commits on master and get a basic idea of all the PRs that went into the codebase.

If you don't mind, I will do some work with git and do a force push so the master branch looks like this.

itsayellow:
I don't mind at all, thanks for offering to clean it up.  Thanks for the tip.

PR #260, opened 2019-10-27
move pipx to src/pipx

### LINKED ISSUE #259: move source code from `pipx` to `src/pipx` directory

Moving pipx's source to a directory that is hard/impossible to import accidentally will solve confusion during development and use of pipx.

Moving to the `src/pipx` directory will make it harder to accidentally run with an unexpected installation of pipx since imports will fail if the local package is not installed. It will also makes sense at a high level for the folder names, i.e. `docs`, `tests`, `src`.

cs01:
fixes #259, see that issue for motivation


itsayellow:
Should we also change Venv.py -> venv.py, SharedLibs.py -> shared_libs.py in this reorganization?

PR #261, opened 2019-10-28
[docs] add maintainers, update contributing instructions

cs01:
* add maintainers
* update contributing instructions


docs/contributing.md
```diff
@@ -1,65 +1,132 @@
...
+
+
+## Dependencies
+pipx uses an automation tool called [nox](https://pypi.org/project/nox/) for development, continuous integration testing, and various tasks.
+
+`nox` defines tasks or "sessions" in `noxfile.py` which can be run with `nox -s SESSION_NAME`. Session names can be listed with `nox -l`.
+
+Install nox for pipx development:
+```
+pip install --user git+https://github.com/cs01/nox.git@5ea70723e9e6 nox
```
itsayellow: Should we have a note here about equivalent `make develop`?  Is there a reason you dropped it?

docs/contributing.md
```diff
@@ -1,65 +1,132 @@
...
+```
+
+!!! note
+    A specific version of nox must be used for pipx development until [nox issue 233](https://github.com/theacodes/nox/issues/233) is resolved.
+
 
 ## Developing pipx
-pipx uses pipx to develop -- it's recursive! But don't worry, it's not that scary. You'll be up in running in no time. You'll also need `make` installed. If you don't have `make`, you can look at pipx's `makefile` to see the what the make targets correspond to).
+Clone pipx, then enter the root of the pipx source tree. Run this command to see which commands you can run with nox.
```
itsayellow: Is this confusing seeing as we have a directory called `src/pipx`?  Maybe instead of `pipx source tree`, use `pipx repository` ?

itsayellow:
Did you mean to modify `noxfile.py` inside this documentation PR?

itsayellow:
I like all the updates to contributing.md!  Lots of good information about the development environment.  In general looks great to me.

docs/contributing.md
cs01: It felt kind of messy having `make` and `nox` together. I also wasn't sure I could assume windows users had `make installed`.

docs/contributing.md
cs01: Ah yeah good call. I'll make that change.

cs01:
> Did you mean to modify `noxfile.py` inside this documentation PR?

Yes it was, but I can separate it out if you want. It makes more sense to only build the docs and the package with a single version, and it makes the docs a little simpler too so I just did it in the same commit.

itsayellow:
> > Did you mean to modify `noxfile.py` inside this documentation PR?
> 
> Yes it was, but I can separate it out if you want. It makes more sense to only build the docs and the package with a single version, and it makes the docs a little simpler too so I just did it in the same commit.

In that case changing `noxfile.py` is ok with me in this PR.

What are the ramifications of changing a list of python versions into a single string?  Why does it make the docs simpler?  (I'm not knowledgeable enough about nox to know if this would have side effects.)

cs01:
> What are the ramifications of changing a list of python versions into a single string? Why does it make the docs simpler? (I'm not knowledgeable enough about nox to know if this would have side effects.)

When you run `nox -l` you will get a session name like `lint-3.7` instead of just `lint`. So we can change the python version and the docs don't need to be updated.

PR #265, opened 2019-10-29
test_install: python3.5 compatibility with pyenv

itsayellow:
With the current test in `test_install.py` involving installing a package using python3.5, our current setup is incompatible with pyenv and fails with it.

The reason for this is because using `shutil.which()` returns a path to the pyenv shim script for the python3.5 executable, not the path to the executable itself.  This shim needs both env and bash for its shebang, and we don't have `/usr/bin` and `/bin` on our path, so this shim script fails.

This pull request adds a helper function that is pyenv-aware, `python_which()`, which first tries to use `pyenv which` if it exists to get the real path to the executable, and then if pyenv is not on the system falls back to using the same `shutil.which()` method as before.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -1,3 +1,5 @@
+from shutil import which
+import subprocess
 import sys
 from typing import List
 from unittest import mock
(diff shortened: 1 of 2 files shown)


cs01:

itsayellow:

Done--rebased.

PR #268, opened 2019-11-01
run app appargs fix for -- double-dash

### LINKED ISSUE #267: `pipx run` cannot pass double-hyphen '--' to app
**Describe the bug**
When running using `pipx run` to run a command that takes a double-dash `--` as an argument, pipx will swallow the `--` and not pass it to the command.

**How to reproduce**
`pipx run nox -- tests/test_install.py`

**Expected behavior**
nox should get arguments `['--', 'tests/test_install.py']`

**Actual behavior**
nox receives `['tests/test_install.py']` in `binary_arguments`, and therefore does not process the extra argument, because nox needs an explicit `--` before any positional arguments:
```carvel:pipx$ pipx run nox -- tests/test_install.py 
usage: nox [-h] [--version] [-l] [-s [SESSIONS [SESSIONS ...]]] [-k KEYWORDS]
           [-v] [-r] [--no-reuse-existing-virtualenvs] [-f NOXFILE]
           [--envdir ENVDIR] [-x] [--no-stop-on-first-error]
           [--error-on-missing-interpreters]
           [--no-error-on-missing-interpreters] [--error-on-external-run]
           [--no-error-on-external-run] [--install-only] [--report REPORT]
           [--non-interactive] [--nocolor] [--forcecolor]
           ...
nox: error: Unknown argument(s) 'tests/test_install.py'.
carvel:pipx$ 
```

I believe that argparse is swallowing the first '--' that it sees in the command-line for itself and not passing it on in the argument list.  It's assuming that you are using the double-dash for pipx and not any other reason.  Possibly we can disable this, but I couldn't find a way to do that.  

Another option might be to modify the args before sending it to `argparse.parser`, to double-up any `--` found after the first positional argument after run.  Kind of hacky but another option.

itsayellow:
Fixes #267 

This PR allows all arguments after `pip run <app_name>` to be sent verbatim to the `<app_name>` app.  Currently a `--` argument is possible to be removed by argparse before sending the list of arguments to the app.

The code included in a new `parse_args()` function doubles a `--` argument if it will be removed by argparse.  argparse will only remove the first `--` in a line, so this ensures that the appargs specified by the user will all be sent to the app.


cs01:

cs01:
I attempted to rebase this in the `appargs_fix` branch to save you the effort.

itsayellow:
> I attempted to rebase this in the `appargs_fix` branch to save you the effort.

Can you "pretend" I'm clueless and explain how to use your remote branch to rebase my fork's branch? 🙂

itsayellow:

Done--rebased.

tests/test_run.py
```diff
@@ -42,3 +42,24 @@ def test_run_script_from_internet(pipx_temp_env, capsys):
...
+    assert "< -- >" not in captured.out
+    run_pipx_cli(["run", "--", "pycowsay", "--", "hello"])
+    captured = capsys.readouterr()
+    assert "< -- hello >" not in captured.out
+    run_pipx_cli(["run", "--", "pycowsay", "hello", "--"])
+    captured = capsys.readouterr()
+    assert "< hello -- >" not in captured.out
+    run_pipx_cli(["run", "--", "pycowsay", "--"])
+    captured = capsys.readouterr()
+    assert "< -- >" not in captured.out
```
cs01: I think in this case we would be better off with unit tests of `parse_args` rather than e2e tests. 

i.e. monkeypatch sys.argv and call `parse_args` with the various permutations you have here.

tests/test_run.py
```diff
@@ -44,22 +46,38 @@ def test_run_script_from_internet(pipx_temp_env, capsys):
...
+        ["--", "--", "hello"],
+        ["hello", "--"],
+        ["hello", "--", "--"],
+        ["--"],
+        ["--", "--"],
+    ]
+    for i, input_argv in enumerate(input_argv):
+        monkeypatch.setattr(sys, "argv", input_argv)
+        parsed_pipx_args = pipx.main.parse_args(parser)
+        assert parsed_pipx_args.appargs == expected_appargs[i]
```
cs01: This is awesome. pytest has a `parameterize` decorator that adds some nice reporting if something fails. Recommend using it, but not a big deal. 

https://docs.pytest.org/en/latest/parametrize.html

tests/test_run.py
```diff
@@ -42,3 +44,40 @@ def test_run_script_from_internet(pipx_temp_env, capsys):
...
         ]
     )
+
+
+def test_appargs_doubledash(pipx_temp_env, capsys, monkeypatch):
+    parser = pipx.main.get_command_parser()
+
+    input_argv = [
+        ["pipx", "run", "pycowsay", "--", "hello"],
+        ["pipx", "run", "pycowsay", "--", "--", "hello"],
```
cs01: An intuitive interface to `pipx`, IMO, with be to support the `--` argument like nox does. pipx would forward anything after the first `--` to the app. So `pipx nox pycowsay -- hello` would say "hello", not "-- hello".

`pipx run nox -- -s lint` would run `nox -s lint`

and `pipx run nox -- -s lint -- nox_posargs` would run `nox -s lint -- nox_posargs`. It looks like currently the way it's implemented in this PR, it  would pass `nox -- -s lint -- nox_posargs`.

itsayellow: I think that's how it works now.  But I am confused, because before I was trying to use a `pipx run` with nox in the past, and it seemed like pipx was removing the `--` wherever it was.  But now while writing the tests, it appears that pipx only removes the first `--` if it is not inside the appargs.  i.e. it may only remove the first argument after the app name if it is `--`.

So somehow it seems to behave differently now than when I was testing it a week or so ago.  I don't know if that's my mistake or code difference.

The other thing is that to my mind things get confusing if pipx or its argparse is doctoring anything after `pipx run <appname>`  In my mind I really expect that all to go verbatim to the app, as if the command line was `<appname> [...]` for `pipx run <appname [...]`.

If we told the user to always put a `--` before `<appname>` then it would basically work fine without any code modification.  That was one way I was thinking about addressing this, with just specifying the help text.

itsayellow:
Now that I see the only unwanted possible `--` that is removed after the app name in `pipx run` is only the first argument after the app name, perhaps I can withdraw the code in this pull request.

@cs01, how about I just add some text to the help to make sure that users know the first `--` after the app name needs to be doubled if it is to be sent to the app?

cs01:
> @cs01, how about I just add some text to the help to make sure that users know the first -- after the app name needs to be doubled if it is to be sent to the app? And just forget about all the code changes?

I think documenting that `--` can be an explicit delimiter between pipx arguments and the app's arguments would be great. I actually didn't know pipx worked like that by default, but I'm glad it does (it's just argparse's default behavior and we get it for free). By documenting that, and maybe a sample nox command on how to pass positional arguments to the app would be good.

tests/test_run.py
clbarnes: Agreed, perfect case for a parametrized test. It'll allow each one to pass or fail individually.

Alternatively, zip would be my preferred way to keep input and output iterating in sync with each other.

itsayellow:
> An intuitive interface to `pipx`, IMO, with be to support the `--` argument like nox does. pipx would forward anything after the first `--` to the app. So `pipx pycowsay -- hello` would say "hello", not "-- hello".
> 
> `pipx run nox -- -s lint` would run `nox -s lint`
> 
> and `pipx run nox -- -s lint -- nox_posargs` would run `nox -s lint -- nox_posargs`. It looks like currently the way it's implemented in this PR, it would pass `nox -- -s lint -- nox_posargs`.

I think the way you prefer seems confusing to me 🙃, especially when you have two apps (like pipx and nox) getting important information (and possibly removing) the `--`.

My preference would be that everything starting with the app/package name would be passed on verbatim as if instead of `pipx run <app> <appargs>` you had installed the app on your system and were just typing `<app> <appargs>` on the command line.  I find it confusing if pipx is possibly removing a `--` in what would be `<appargs>` when I wanted to send that `--` on to the `<app>`.

The way out of this currently would be to be explicit, and put the double-dash `--` __before__ the app/package name:
```
pipx run -- nox -s lint -- nox_posargs
```
That way I can think in my head what I'm really trying to run is
```
nox -s lint -- nox_posargs
```
and I don't have to worry that pipx is going to remove a `--` somewhere after the `nox`

Is my example above ok to put as guidance in the help?  To put the `--` before the app name?

Otherwise to my mind we say, "pipx will remove the first double-dash after the app name, unless you also put a double-dash anywhere beforehand in the pipx command.  Also if you put the double-dash later after the app name (not the first argument after the app name), it will be left intact and sent to the app."  That is VERY confusing to me.

clbarnes:
How do either of those options compare to tools like node's `npm` and ruby's `bundler`? They're doing more or less the same thing and are extremely widely used in their respective ecosystems. If "least surprise" is a design goal, making pipx consistent with those would be my choice.

itsayellow:
> How do either of those options compare to tools like node's `npm` and ruby's `bundler`? They're doing more or less the same thing and are extremely widely used in their respective ecosystems. If "least surprise" is a design goal, making pipx consistent with those would be my choice.

This is a really good thought.  Unfortunately it seems that that each one does it a different way, one the way I prefer, the other one like @cs01 prefers.
npm uses the method that separates the command with and the arguments with a `--`:
```
npm run-script <command> [--silent] [-- <args>...]
```
wheras bundler sends `command` and all following arguments after it unedited to the command.  Even if the first argument after command is `--`. (I just tested it to be sure)
```
bundle exec [--keep-file-descriptors] command
```

clbarnes:
Perfect <_<

If in doubt, I'd go with npm's behaviour as a tiebreaker, as it's probably more widely used. Familiarity over purity.

suominen:
The double dash is a guideline set by POSIX in [IEEE Std 1003.1, Chapter 12: Utility Conventions](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap12.html), subchapter 12.2: Utility Syntax Guidelines, **Guideline 10**:

> The argument `--` should be accepted as a delimiter indicating the end of options. Any following arguments should be treated as operands, even if they begin with the '-' character. The `--` argument should not be used as an option or as an operand.

`argparse` processes the first double dash and treats any remaining arguments as arguments (as opposed to options), including any additional double dashes.

See also:
* [When and how was the double dash introduced as an end of options delimiter in Unix/Linux](https://unix.stackexchange.com/questions/147143/when-and-how-was-the-double-dash-introduced-as-an-end-of-options-delimiter)
* [argparse: use of double dash to separate options and positional arguments](https://code.activestate.com/lists/python-list/693344/) (best to read the whole thread)

itsayellow:
> The double dash is a guideline set by POSIX in [IEEE Std 1003.1, Chapter 12: Utility Conventions](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap12.html), subchapter 12.2: Utility Syntax Guidelines, **Guideline 10**:

I understand about the double-dash, I think the discussion was more about whether our doc guidance should position the double-dash should be between the command to run and the list of its arguments, OR whether it should should precede both.

cs01:
>  think the discussion was more about whether our doc guidance should position the double-dash between the command to run and the list of its arguments, OR whether it should should precede both.

It should precede both.

If anything I would rather get rid of pipx's support for working _without_ the `--`, similar to how `nox` does it.

itsayellow:
> It should precede both.

Oh I misunderstood you before, I thought you wanted the double-dash between command and command arguments.  Well great!  Then I whole-heartedly agree.

So to recap and double-check, this would be the recommendation:
```bash
pipx run --spec git+https://blah.com/user/app.git -- app apparg1 apparg2 apparg3
```

itsayellow:
I just re-did this PR as a modification to __require__ the double-dash `--` before the `app` and `appargs` arguments.  The usage help text has been modified to show a `--` before the `app` argument, and it exits with the customary argparse error if you do not put a `--` before the `app` argument.

cs01:
Sorry I think I was not clear in my last post.

The pipx API should look like this:
```
pipx run --spec git+https://blah.com/user/app.git app -- apparg1 apparg2 apparg3 --appflag
```

`--spec` is a valid pipx argument, and the app name is a required positional pipx argument. Only the arguments we want to forward to the app should be preceded by `--`. They act as a delimeter between pipx arguments and the app-to-be-run's arguments. This will avoid ambiguity when the user wants to run help for pipx vs help for the app. Consider these three

* `pipx run app --help` - works today, prints help for app not pipx, but it's ambiguous and can only be confirmed by running and confirming it works.
* `pipx run app -- --help` - also works today. Is explicit and not ambiguous. Should be required going forward.
* `pipx run --help app -- --help` - should print help for pipx since its arguments are handled before launching the app


The argparse docs have this

> If you have positional arguments that must begin with - and don’t look like negative numbers, you can insert the pseudo-argument '--' which tells parse_args() that everything after that is a positional argument

And the `rm` command's man page has this:
```
...
To  remove a file whose name starts with a '-', for example '-foo', use
       one of these commands:

              rm -- -foo
              rm ./-foo
```

The command for running `rm` if it were an app pipx could run would be

```
pipx run rm -- -- -foo
```
Where the first `--` means "everything after this gets forwarded directly to `rm`", and the remainder is run by rm, `-- -foo`.

It doesn't follow the POSIX conventions mentioned by @suominen to do `pipx run -- app` because it would mean `app` is being forwarded to some other process when it is actually and argument to pipx itself.

itsayellow:
I totally understand and agree with everything in your last comment up and until this point:

> It doesn't follow the POSIX conventions mentioned by @suominen to do `pipx run -- app` because it would mean `app` is being forwarded to some other process when it is actually an argument to pipx itself.

In reading of @suominen's links I don't see anything about whether an argument is forwarded to other processes.  Described in those links, it is just that `--` is supposed to separate options from positional arguments (which is actually the title of one of the links.)  Since `app` is a positional argument, I don't understand why `--` would go after it.

Posix2 from the stack exchange link:
> Guideline 10:    The argument "--" should be accepted as a delimiter indicating the end of options.  Any following arguments should be treated as operands, even if they begin with the '-' character.  The "--" argument should not be used as an option or as an operand.

I don't think of `app` as an option, so in my reading this would put the `--` before it.

suominen:
These two should behave the same:

```
pipx run -- app --help
pipx run app -- --help
```

Traditionally UNIX utilities follow this calling convention:

```
command [[-o|--option] ...] [arg ...]
```

In other words, options must precede arguments. Thus these two should behave the same:

```
pipx run foo --help
pipx run -- foo --help
```

Here is a sample shell script I called `test-getopts`:
```
#!/bin/sh

PATH=/bin:/usr/bin
export PATH

c=false

while getopts a:b:c opt
do
    case "${opt}" in
    a) a="${OPTARG}";;
    b) b="${OPTARG}";;
    c) c=true;;
    *) usage 1>&2; exit 1;;
    esac
done
shift $((${OPTIND} - 1))

echo arguments: "${@}"
```

And here is a sample run:

```
> sh -x test-getopts -a x -c foo -b
+ PATH=/bin:/usr/bin
+ export PATH
+ c=false
+ getopts a:b:c opt
+ a=x
+ getopts a:b:c opt
+ c=true
+ getopts a:b:c opt
+ shift 3
+ echo arguments: foo -b
arguments: foo -b
```

Notice also how the `-x` is processed by `sh` but what look like options after the first argument are actually handled as arguments by it, and passed to the `test-getopts` script. You can see examples of this "everywhere."

In the end it is up to the program itself how it parses its argument vector, and many modern tools allow options and arguments to be mixed. Here is an example from the AWS CLI manual:

```
aws s3 cp /tmp/foo/ s3://bucket/ --recursive --exclude "*" --include "*.jpg"
```

Promoting a single convention in the `pipx` documentation is a good thing. Using the double dash (`--`) in documentation and usage texts makes it clear what the user should expect to be passed on to the app.

Enforcing that a double dash is used is counter-intuitive to me, but I also understand the motivation behind that, and would not want to spend time arguing against it: for me it would just become "one of those quirks" that I'd have to remember when it comes to `pipx`. :smirk:

suominen:
I think `pipx` is already processing `--` "correctly" -- just try these:

```
pipx --help run
pipx --help run
pipx run --help
pipx run -- --help
pipx -- run --help
```

Oops: the last one "fails."

cs01:
The summary of my motivation here is: 

There is a collection of arguments passed to and validated by pipx (`app` is part of these arguments). 

There is a **different** collection of arguments forwarded, completely unparsed and untouched, to the app. 

Those two collections of arguments should be separated explicitly. The best mechanism and convention for this is with the `--` delimiter since

> The argument -- should be accepted as a delimiter indicating the end of options.

cs01:
> I think pipx is already processing -- "correctly"

I just tried these two, which are probably as tricky as it will get, and they both worked. I thought the first one wouldn't, but it did. 

```
pipx run nox -r -s lint -- nox-lint-args
pipx run nox -- -r -s lint -- nox-lint-args
```

itsayellow:
> So in summary: I like the way everything behaves, and I don't think it's necessary to _require_ the `--`. The only thing that we might want to add is documentation of this behavior.

Yep I agree too with that.

itsayellow:
I will write some more docs, and not make the `--` required so it can stay flexible.

I think the only thing I still worry about is the "gotcha" to unsuspecting users that "got" me once.  That the user wants to run the command:
```
pyapp -- pyapparg1 pyapparg2
```
And they use `pipx run` like so (wrongly and naively):
```
pipx run pyapp -- pyapparg1 pyapparg2
```
Which would remove the `--` before passing only `pyapparg1 pyapparg2` to pyapp.

What about writing a note to stderr telling the user that the `--` in this case is not being passed to pyapp?  This note would be written only in cases where the next item in the argv list after `app` is a `--`.  I think it might immediately help solve user confusion such as in the above example.

suominen:
> ```
> pipx run pyapp -- pyapparg1 pyapparg2
> ```
> 
> Which would remove the `--` before passing only `pyapparg1 pyapparg2` to pyapp.

As the `--` is after the first non-option argument, it should not be removed. So that's a bug in argument parsing.

```
> sh -x test-getopts -- -a
+ PATH=/bin:/usr/bin
+ export PATH
+ c=false
+ getopts a:b:c opt
+ shift 1
+ echo arguments: -a
arguments: -a
```

itsayellow:
> As the `--` is after the first non-option argument, it should not be removed. So that's a bug in argument parsing.

That was what was confusing me about our current pipx run parsing, and what motivated me to start this PR.

cs01:
> What about writing a note to stderr telling the user that the -- in this case is not being passed to pyapp? This note would be written only in cases where the next item in the argv list after app is a --. I think it might immediately help solve user confusion such as in the above example.

Sounds perfect!

itsayellow:
> The summary of my motivation here is:
> 
> There is a collection of arguments passed to and validated by pipx (`app` is part of these arguments).
> 
> There is a **different** collection of arguments forwarded, completely unparsed and untouched, to the app.
> 
> Those two collections of arguments should be separated explicitly. The best mechanism and convention for this is with the `--` delimiter since
> 
> > The argument -- should be accepted as a delimiter indicating the end of options.

I understand your motivation, I'm just worried it's different than expected for users accustomed to the usual unix CLI behavior of `--`.  In my mind an "option" argument is anything that is optional (i.e. starts with a `-` or `--` or an argument of such an option).  Whereas in `pipx run`, `app` is positional and required.

If I'm belaboring this a little, it's because it sounds like the next release of pipx my be a major API-changing one and it seems now is the time to discuss this.

I always think of app and appargs being tightly coupled, as one command invocation that is run inside of a temporary pipx environment.  I'm wondering if other users think the same way.

One very easy solution would to make app_and_appargs all one argument, giving it the `nargs=argparse.REMAINDER` argument.  That way you could still precede it with `--` if you want to, but it wouldn't change anything in-between `app` and `appargs`.

But I'll defer to @cs01 if this is just not the way you want to go. 🙂

cs01:
Could you remind me if you are advocating for requiring `--` every time, or only requiring that its position  be immediately preceding the CLI app (or URL) to run?

itsayellow:
I'd be up for not requiring anything.  I'm requesting that the `--` only be removed if it comes before all positional arguments, but not be removed if it comes between the positional `app` and `appargs` arguments.

Because as @suominen notes, it's not the usual unix CLI behavior for a `--` coming after a positional (non-option) argument to be removed.

My solution would be to replace `app` and `appargs` with something equivalent to `app_and_appargs` that has the `nargs=argparse.remainder` set.  That way everything including and after `app` would be passed on verbatim.

cs01:
> One very easy solution would to make app_and_appargs all one argument, giving it the nargs=argparse.REMAINDER argument. That way you could still precede it with -- if you want to, but it wouldn't change anything in-between app and appargs.

This sounds good to me, let's do it.

cs01:
Note that pipx does support the case of a URL being the app, but I don't think that should be unworkable with the solution you proposed.

cs01:
LMK when this is ready for review :+1:

itsayellow:
I've summarized our current test suite for `pipx run` argument parsing in the table below.  Just to make sure we're all on the same page.

@suominen, could you take a look at it and see if the arguments are being parsed as you expect?

pipx command | app and appargs to execute
-------------|----------------------------
pipx run -- pycowsay -- hello | pycowsay -- hello
pipx run -- pycowsay -- -- hello | pycowsay -- -- hello
pipx run -- pycowsay hello -- | pycowsay hello --
pipx run -- pycowsay hello -- -- | pycowsay hello -- --
pipx run -- pycowsay -- | pycowsay --
pipx run -- pycowsay -- -- | pycowsay -- --
pipx run pycowsay -- hello | pycowsay -- hello
pipx run pycowsay -- -- hello | pycowsay -- -- hello
pipx run pycowsay hello -- | pycowsay hello --
pipx run pycowsay hello -- -- | pycowsay hello -- --
pipx run pycowsay -- | pycowsay --
pipx run pycowsay -- -- | pycowsay -- --
pipx run -- -- pycowsay -- | -- pycowsay --

itsayellow:
If we all like the way `--` is being parsed then I think this is ready to go.

AFAIK, the only changes compared to current release `pipx run` behavior are:
* the usage info for `pipx run` has changed, instead of `app` and `appargs` there is now only a combined`app ...` with one line of help text.
* Now, if you put a `--` directly after `app`, it will be passed to the app as the first apparg.

suominen:
LGTM -- thank you!

cs01:
@itsayellow you rock :metal: 

Looks good besides the Windows CI failure (which is unrelated to your changes). Created https://github.com/pipxproject/pipx/issues/296. I am fine merging this before the Windows CI issue is fixed.

PR #271, opened 2019-11-04
Fix help text for reinstall-all --python arg.

itsayellow:
This is a tiny fix for an awkward help blurb for the --python option of reinstall-all.  I made it more clear and more in line with install help text.

I'm assuming this won't be problematic, and if there are no objections I will merge it myself.

### DIFF
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -411,7 +411,7 @@ def _add_reinstall_all(subparsers):
         "--python",
         default=constants.DEFAULT_PYTHON,
         help=(
-            "The Python executable used to reinstall the Virtual Environment "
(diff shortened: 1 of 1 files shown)


src/pipx/main.py
```diff
@@ -410,7 +410,10 @@ def _add_reinstall_all(subparsers):
     p.add_argument(
         "--python",
         default=constants.DEFAULT_PYTHON,
-        help="The Python version to reinstall the package's CLI app with. Must be v3.5+.",
+        help=(
+            "The Python executable used to reinstall the Virtual Environment "
+            "and run the associated app/apps. Must be v3.5+."
+        ),
```
cs01: I would suggest changing "reinstall" to "recreate"

PR #277, opened 2019-11-15
fallback to threads if no semaphore support

### LINKED ISSUE #229: Unable to run 'pipx list' on Termux (Android)
Running `pipx list` on Termux results in following stack trace:
```
venvs are in /data/data/com.termux/files/home/.local/pip
x/venvs
apps are exposed on your $PATH at /data/data/com.termux/
files/home/.local/bin
  File "/data/data/com.termux/files/usr/lib/python3.7/mu
ltiprocessing/synchronize.py", line 28, in <module>
    from _multiprocessing import SemLock, sem_unlink
ImportError: cannot import name 'SemLock' from '_multipr
ocessing' (/data/data/com.termux/files/usr/lib/python3.7
/lib-dynload/_multiprocessing.cpython-37m.so)

During handling of the above exception, another exceptio
n occurred:

  File "/data/data/com.termux/files/home/.local/bin/pipx
", line 10, in <module>
    sys.exit(cli())
  File "/data/data/com.termux/files/home/.local/pipx/ven
vs/pipx/lib/python3.7/site-packages/pipx/main.py", line 
547, in cli
    exit(run_pipx_command(parsed_pipx_args))
  File "/data/data/com.termux/files/home/.local/pipx/ven
vs/pipx/lib/python3.7/site-packages/pipx/main.py", line 
198, in run_pipx_command
    commands.list_packages(venv_container)
  File "/data/data/com.termux/files/home/.local/pipx/ven
vs/pipx/lib/python3.7/site-packages/pipx/commands.py", l
ine 638, in list_packages
    with multiprocessing.Pool() as p:
  File "/data/data/com.termux/files/usr/lib/python3.7/mu
ltiprocessing/context.py", line 119, in Pool
    context=self.get_context())
  File "/data/data/com.termux/files/usr/lib/python3.7/mu
ltiprocessing/pool.py", line 158, in __init__
    self._setup_queues()
  File "/data/data/com.termux/files/usr/lib/python3.7/mu
ltiprocessing/pool.py", line 251, in _setup_queues
    self._inqueue = self._ctx.SimpleQueue()
  File "/data/data/com.termux/files/usr/lib/python3.7/mu
ltiprocessing/context.py", line 112, in SimpleQueue
    return SimpleQueue(ctx=self.get_context())
  File "/data/data/com.termux/files/usr/lib/python3.7/mu
ltiprocessing/queues.py", line 332, in __init__
    self._rlock = ctx.Lock()
  File "/data/data/com.termux/files/usr/lib/python3.7/mu
ltiprocessing/context.py", line 66, in Lock
    from .synchronize import Lock
  File "/data/data/com.termux/files/usr/lib/python3.7/mu
ltiprocessing/synchronize.py", line 32, in <module>
    " synchronization primitives needed will not" +
ImportError: This platform lacks a functioning sem_open 
implementation, therefore, the required synchronization 
primitives needed will not function, see issue 3770.
```
This is on OnePlus 5T with Android 9.0.

Technically it's a Python and/or Android issue, where `multiprocessing` module isn't working because Android kernel is lacking in semaphores syscalls. Also see https://bugs.python.org/issue3770

AFAIK there is still a way to workaround this, but it would require switching `multiprocessing` to `threading` module. For example, this is how `black` dealt with similar issue: https://github.com/psf/black/pull/533

justsh:
Platforms like Android Termux do not have support for semaphores and
thus cannot use `multiprocessing.Pool`

Use the multiprocessing.dummy package as a fallback, which provides the
Pool API backed by a ThreadPool.

Fixes #229

### DIFF
--- a/src/pipx/commands/commands.py
+++ b/src/pipx/commands/commands.py
@@ -3,7 +3,6 @@
 import datetime
 import hashlib
 import logging
-import multiprocessing
 import shlex
 import shutil
 import subprocess
@@ -15,6 +14,19 @@ import urllib.request
 from pathlib import Path
 from shutil import which
 from typing import List
+
+try:
+    # Instantiating a Pool() attempts to import multiprocessing.synchronize,
+    # which fails if the underlying OS does not support semaphores.
+    # Here, we import ahead of time to decide which Pool implementation to use:
+    # one backed by Processes (the default), or one backed by Threads
+    import multiprocessing.synchronize  # noqa: F401
+except ImportError:
+    # Fallback to Threads on platforms that do not support semaphores
+    # https://github.com/pipxproject/pipx/issues/229
+    from multiprocessing.dummy import Pool
+else:
+    from multiprocessing import Pool
 
 import userpath  # type: ignore
 from pipx import constants
@@ -574,7 +586,7 @@ def list_packages(venv_container: VenvContainer):
 
     venv_container.verify_shared_libs()
 
-    with multiprocessing.Pool() as p:
+    with Pool() as p:
         for package_summary in p.map(_get_package_summary, dirs):
             print(package_summary)
 


src/pipx/commands/commands.py
```diff
@@ -16,6 +15,15 @@
 from shutil import which
 from typing import List
 
+try:
+    import multiprocessing.synchronize
```
itsayellow: Is it not possible just to do:
```
try:
    from multiprocessing import Pool
except ImportError:
    from multiprocessing.dummy import Pool
```
```

Comment (justsh) [reply src/pipx/commands/commands.py]: Nope. Per the stack trace in #229, the error occurs when a Pool is constructed, not imported, at which point `multiprocessing.synchronize` fails to import the needed semaphore classes from `_multiprocessing`.

```
File "/data/data/com.termux/files/home/.local/pipx/venvs/pipx/lib/python3.7/site-packages/pipx/commands.py", line 638, in list_packages
    with multiprocessing.Pool() as p:
  File "/data/data/com.termux/files/usr/lib/python3.7/multiprocessing/context.py", line 119, in Pool
  # snipped...
    File "/data/data/com.termux/files/usr/lib/python3.7/multiprocessing/context.py", line 66, in Lock
    from .synchronize import Lock
```

I see that flake8 complains about the unused import, that's my bad. I ran `nox` locally but didn't see any warnings during linting, will investigate on my end.

I can import both a `Pool` and a `ThreadPool` and move the `try` down to when the Pool is instantiated if that's preferable to a `#noqa`.

Or just switch to threads, your call.

itsayellow: I'd lean toward `# noqa` with an explanatory comment for why we are trying to import something we are not using (a brief version of your above comment.)

Probably @cs01 should chime in on hist preferences.

clbarnes:
In general, is there a benefit to using a `multiprocessing.Pool` over a `multiprocessing.dummy.Pool` in general? Could we use the "fallback" in all cases to prevent this platform-specific behaviour?

itsayellow:
> In general, is there a benefit to using a `multiprocessing.Pool` over a `multiprocessing.dummy.Pool` in general? Could we use the "fallback" in all cases to prevent this platform-specific behaviour?

I believe that multiprocessing.dummy is just use threads but with same API as multiprocessing.  The advantage to using multiprocessing.Pool would be to really use all the user's processors to execute the tasks in parallel, which should be faster than using threads.

clbarnes:
Ah, yes, my fault, got confused about the many implementations of concurrency in python. 

We possibly should benchmark the use of threads just in case (even if it's just some "anecdotal" benchmarks) - I suspect the bottlenecks in pip installation are downloads, decompressing zip archives, and compilation if necessary. All that's left is really copying files around. All of these things should release the GIL, so threads may not be much slower to run than processes. Multiple CPUs will be used, as I understand it, because python threads are backed by OS threads (again, as I understand it) and it's python which imposes the lock.

However, if processes is the way to go, I think trying to import the synchronization is a good way to decide which concurrency implementation to use. I'd definitely throw in a comment. I'd be happy with a noqa in that circumstance.

uranusjr:
> I suspect the bottlenecks in pip installation are downloads, decompressing zip archives, and compilation if necessary.

Those pip calls are all in subprocesses, which is not GIL-ed anyway, so it doesn’t matter whether you use threads or forks. A benchmark would be useful, and I suspect the difference would be insignificant.

But this really is a whole other topic on the other hand. I’d suggest merging this first (since it does fix a problem without introducing regressions) with the suggested comment, and open a new issue for whether to switch completely to dummy.

justsh:
> I'd lean toward # noqa with an explanatory comment for why we are trying to import something we are not using (a brief version of your above comment.)

Added the `#noqa` and updated the comment, feel free to nit.

> A benchmark would be useful, and I suspect the difference would be insignificant.

FWIW in the limited, localized testing I did before submitting this PR,
using Processes was only faster than Threads by tenths of a second.
As a naive profile I checked out one worktree with `multiprocessing` and one with `multiprocessing.dummy` and used `/usr/bin/time pipx run --spec=<worktree_path> pipx list` against the dozen venvs I already manage with pipx.

itsayellow:
And after #222 lands, we won't even run a subprocess to get metadata on a given package, so it may matter even less.

PR #282, opened 2019-11-28
DOC: Fixed completion command in README.md

SaturnFromTitan:
Seems like the command in the README is wrong. I hope this wasn't changed recently - at least I couldn't find seemingly related issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Upgrade pipx with `python3 -m pip install -U pipx`.
 
 Shell completions are available by following the instructions printed with this command:
 ```
-pipx completions
+pipx completion
 ```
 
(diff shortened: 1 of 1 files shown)


cs01:
There are a few problems here
* `completion` is the command for the old pipx, before the name is transferred, see https://github.com/pipxproject/pipx/pull/230. If you are using that version, you you are using the wrong tool, https://github.com/khalti/pipx. There is an open issue to stop accidental installations of that version of pipx: https://github.com/pipxproject/pipx/issues/237
* Modifications to README.md should be made by changing the `templates/index.md` file, then generating the readme and the github pages files. But since this PR was wrong in the first place it is not relevant.

Can someone please undo this merge?

gaborbernat:
Done 👍

cs01:
Thank you 👍

PR #289, opened 2019-12-06
Pipx metadata changelog

itsayellow:
Extra changelog items and rebuild of docs for PR #222

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,9 +1,9 @@
 dev
 
-- Now store metadata information for each application installed, including install options like `--spec`, and injected packages. This information allows upgrade, upgrade-all and reinstall-all to work properly even with non-pypi installed packages.  (#222)
-- Change `upgrade`, removing options `--spec`, `--include-deps`.  Use original options used to install application instead. (#222)
-- Change `upgrade-all`, removing options `--include-deps`, `--system-site-packages`, `--index-url`, `--editable`, `--pip-args`.  Use original options used to install each application instead. (#222)
-- Change `reinstall-all`, removing options `--include-deps`, `--system-site-packages`, `--index-url`, `--editable`, `--pip-args`.  Use original options used to install each application instead. (#222)
+- Metadata is now stored for each application installed, including install options like `--spec`, and injected packages. This information allows upgrade, upgrade-all and reinstall-all to work properly even with non-pypi installed packages.  (#222)
+- `upgrade` options `--spec` and `--include-deps` were removed.  Pipx now uses the original options used to install each application instead. (#222)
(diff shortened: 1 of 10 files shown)


itsayellow:
btw I just noticed that commit 8011c12dda4f916ef185f7d0dfc8a4db73a5b044 "null out pythonpath" changed the version number in `main.py` from "0.14.0.0" to "0.14.0.1".  I assume that was intentional?

docs/changelog.md
```diff
@@ -1,5 +1,9 @@
 dev
 
+- Now store metadata information for each application installed, including install options like `--spec`, and injected packages. This information allows upgrade, upgrade-all and reinstall-all to work properly even with non-pypi installed packages.  (#222)
+- Change `upgrade`, removing options `--spec`, `--include-deps`.  Use original options used to install application instead. (#222)
+- Change `upgrade-all`, removing options `--include-deps`, `--system-site-packages`, `--index-url`, `--editable`, `--pip-args`.  Use original options used to install each application instead. (#222)
+- Change `reinstall-all`, removing options `--include-deps`, `--system-site-packages`, `--index-url`, `--editable`, `--pip-args`.  Use original options used to install each application instead. (#222)
```
cs01: nit: The way this is phased sometimes comes across like a user has to take action. For example 

> Now store metadata information

could be written as

> Metadata is now stored for each application...

And

> Use original options used to install each application instead.

might be more clear as 

> Original options are used to install each application instead

cs01:
Hmmm after thinking a little more, the README and docs should reflect what is on PyPI, so maybe we shouldn't be generating docs until a new release is published to PyPI.

docs/changelog.md
itsayellow: Yeah you're right, I was in "git commit" mode when writing these.  I'll make them more passive.

itsayellow: I fixed the ones I added in this PR, but I notice that some of the previous changelog entries sound similarly commanding.  I left them be for now.

cs01:
Looks good other than the version change. It should remain at 0.14.0.0 for now still.

itsayellow:
> Looks good other than the version change. It should remain at 0.14.0.0 for now still.

Right, I forgot to do that.  Should be good now, only changelog entries.

cs01:
Looks good. I was hoping to see the version go back to 0.14.0.0 and corresponding doc changes, but that can be done in another PR.

itsayellow:
> Looks good. I was hoping to see the version go back to 0.14.0.0 and corresponding doc changes, but that can be done in another PR.

Oh I can do that too if you want.  I thought you said you were going to request those changes from the PR that changed them.

I can definitely change the version back.  And fix the docs.

itsayellow:
@cs01, maybe you can double check my reversions to the docs.  I believe they are correct.

I changed the version back to 0.14.0.0 in main.py

I changed docs/docs.md back to the version from October 19, 2019 (e835671559563cec610d3522782ca)

Issue #255, opened 2019-10-27
break `commands.py` into individual files


cs01:
**How would this feature be useful?**

`commands.py` has grown to be quite large, so breaking it into file names corresponding to the command name will help reduce the barrier to entry for new devs in pipx's codebase.

**Describe the solution you'd like**

Make a folder called `commands` and in that folder add files corresponding to each pipx command, such as `commands/list_command.py`, `commands/install_command.py`, etc.

**Describe alternatives you've considered**

Break `commands.py` into files in the same directory. This will make it harder to distinguish between libraries, classes and commands though, so I think putting commands in a `commands` dir will be the option of least astonishment.

**Notes**
* Created per suggestion in https://github.com/pipxproject/pipx/pull/246#pullrequestreview-304527855.
* Marking as good first issue because it's purely a refactor. Existing unit tests should be sufficient.

ptisserand:
Hi, could I look at this issue ?

cs01:
Sure it's all yours if you want to give it a shot. I would suggest splitting out each command one or two at a time and making a PR for each.

cs01:
One other thing -- you might want to work off of https://github.com/pipxproject/pipx/pull/260 since it moves the files from `pipx/commands.py` to `src/pipx/commands.py`. If you don't we might end up with merge conflicts.

ptisserand:
Ok thanks, I didn't see #260

ptisserand:
Do you know how could I handle to keep commands.py and have a directory named commands ?
Or maybe should I move actual commands.py to commands/non_splitted.py ?
This file will be used only when splitting commands.py file is work in progress.

gaborbernat:
I think this is pretty straightforward, just make sure that commands.py get splited into a package as of:

```
commands
|- __init__.py
|- a
|- b
|- c
```

cs01:
I think commands/commands.py should work, and eventually it will disappear.

cs01:
And if you find it is easy enough to do in one pass without the intermediate file, that’s fine too.

ptisserand:
> And if you find it is easy enough to do in one pass without the intermediate file, that’s fine too.

Yes I think it will be easier for me, since I don't know how to have the following structure with python modules:
- commands.py
commands/
   - __init__.py
  - list_command.py
  - install_command.py

gaborbernat:
Definitely don't have the module and the package at the same time.

cs01:
[...]

PR #291, opened 2019-12-14
Remove duplicate code in commands.commands

chrahunt:
Progresses #255.


cs01:
Ah good catch, thank you.

PR #299, opened 2019-12-20
Remove gcc-related path from test PATH

chrahunt:
As suggested in [#298 (comment)](https://github.com/pipxproject/pipx/pull/298#issuecomment-567566622).


cs01:
Nice, thanks!

Issue #147, opened 2019-04-24
Making spec the argument of install


apnewberry:
In https://github.com/pipxproject/pipx/issues/144, @raxod502 mentioned 

> I am quite confused why the package name I give pipx wouldn't normally be passed to pip install. I don't see any explanation of what the alternative is.

Why do we separately specify the package name and the requirement-specifier? For example, the API I wonder about is just copying pip's interface:

```
$ pip install --help  

Usage:   
  pip install [options] <requirement specifier> [package-index-options] ...
  pip install [options] -r <requirements file> [package-index-options] ...
  pip install [options] [-e] <vcs project url> ...
  pip install [options] [-e] <local project path> ...
  pip install [options] <archive url/path> ...

```

The advantages are:

- A distribution name can be specified, or a url or a path. 
- Any number of requirements can be specified.
- I don't have to learn a new syntax, I can just use pip's which I already know.

clbarnes:
I believe it is because pipx needs to make a virtualenv based on the package name, which is not always clear from the spec (e.g. if it's a wheel file or github repo). Requiring a package as well as a spec means you don't have to parse edge cases to find out the package name. Agreed that it would be nice if it weren't necessary.

tekumara:
I agree duplicating the `pip` interface would provide and much more intuitive and easier first user experience

uranusjr:
The problem is that you cannot know the package name (which you propose for determining the venv’s name) without running setup first, but pipx needs a venv to run setup. So duplicating pip’s interface would mean pipx needs to create two venvs and throw one away just to discover the package name. Not a good trade off IMO.

tekumara:
If a package name is not provided, could it be inferred from the URL?

uranusjr:
It can in some cases (and in some cases even reliably), but there is no guarantee a Git repo is named the same as the package. It’s all about edge cases, unfortunately.

uranusjr:
As mentioned above, pipx names the venv after the package, so a wrongly inferred package name would result in a wrongly named venv. This means that pipx would be unable to correctly identify those packages, and in turn fail to upgrade or uninstall a package correctly.

raxod502:
> duplicating pip’s interface would mean pipx needs to create two venvs and throw one away just to discover the package name

[...]

PR #302, opened 2019-12-24
inject, install accept package spec without --spec option

itsayellow:
This PR allows the user to specify any valid pip package specification instead of a package name for the main argument to `pip install` and each dependency argument to `pip inject`.  It removes the `--spec` option for `pip install` because now that option is unnecessary.  And it enables a package spec for `pip inject` which formerly wasn't possible.

It takes about 2 extra seconds on my local computer to determine the package name from package spec.  If `package_spec` is a valid pypi name and not a local path, I make a shortcut and assume that it is the package name and don't try to independently determine the name (in which case it takes no extra time).

I tried to add some tests, but there could be more tests added, specifically regarding the many different VCS and other URL methods to specify pip packages.  One issue with that is that `git+...` package specifications need `git` in the path, which I wasn't sure how to do.  (https://github.com/pipxproject/pipx/pull/298#issuecomment-568642820 seems like a good way forward on this.)  Additionally, I just wasn't sure where to find good package urls to use to test.

This PR is in reference to issue #147

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -16,12 +16,12 @@ pipx install --include-deps jupyter
 pipx enables you to test various combinations of Python versions and package versions in ephemeral environments:
 ```
 pipx run BINARY  # latest version of binary is run with python3
-pipx --spec PACKAGE==2.0.0 run BINARY  # specific version of package is run
-pipx --python 3.4 run BINARY  # Installed and invoked with specific Python version
-pipx --python 3.7 --spec PACKAGE=1.7.3 run BINARY
-pipx --spec git+https://url.git run BINARY  # latest version on master is run
-pipx --spec git+https://url.git@branch run BINARY
-pipx --spec git+https://url.git@hash run BINARY
+pipx run --spec PACKAGE==2.0.0 BINARY  # specific version of package is run
+pipx run --python 3.4 BINARY  # Installed and invoked with specific Python version
+pipx run --python 3.7 --spec PACKAGE=1.7.3 BINARY
+pipx run --spec git+https://url.git BINARY  # latest version on master is run
+pipx run --spec git+https://url.git@branch BINARY
+pipx run --spec git+https://url.git@hash BINARY
 pipx run pycowsay moo
 pipx --version  # prints pipx version
 pipx run pycowsay --version  # prints pycowsay version
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -24,14 +24,14 @@ pipx --spec git+https://url.git@branch run BINARY
 pipx --spec git+https://url.git@hash run BINARY
 pipx run pycowsay moo
 pipx --version  # prints pipx version
-pipx run pycowsay  --version  # prints pycowsay version
-pipx --python pythonX pycowsay
-pipx --spec pycowsay==2.0 pycowsay --version
-pipx --spec git+https://github.com/psf/black.git black
-pipx --spec git+https://github.com/psf/black.git@branch-name black
-pipx --spec git+https://github.com/psf/black.git@git-hash black
-pipx --spec https://github.com/psf/black/archive/18.9b0.zip black --help
-pipx https://gist.githubusercontent.com/cs01/fa721a17a326e551ede048c5088f9e0f/raw/6bdfbb6e9c1132b1c38fdd2f195d4a24c540c324/pipx-demo.py
+pipx run pycowsay --version  # prints pycowsay version
+pipx run --python pythonX pycowsay
+pipx run --spec pycowsay==2.0 pycowsay --version
+pipx run --spec git+https://github.com/psf/black.git black
+pipx run --spec git+https://github.com/psf/black.git@branch-name black
+pipx run --spec git+https://github.com/psf/black.git@git-hash black
+pipx run --spec https://github.com/psf/black/archive/18.9b0.zip black --help
+pipx run https://gist.githubusercontent.com/cs01/fa721a17a326e551ede048c5088f9e0f/raw/6bdfbb6e9c1132b1c38fdd2f195d4a24c540c324/pipx-demo.py
 ```
 
 ## `pipx inject` example
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -3,11 +3,11 @@
 pipx install pycowsay
 pipx install --python python3.6 pycowsay
 pipx install --python python3.7 pycowsay
-pipx install --spec git+https://github.com/ambv/black black
-pipx --spec git+https://github.com/ambv/black.git@branch-name black
-pipx --spec git+https://github.com/ambv/black.git@git-hash black
-pipx install --spec https://github.com/ambv/black/archive/18.9b0.zip black
-pipx install --spec black[d] black
+pipx install git+https://github.com/ambv/black
+pipx install git+https://github.com/ambv/black.git@branch-name
+pipx install git+https://github.com/ambv/black.git@git-hash
+pipx install https://github.com/ambv/black/archive/18.9b0.zip
+pipx install black[d]
 pipx install --include-deps jupyter
 ```
 
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -3,10 +3,10 @@
 pipx install pycowsay
 pipx install --python python3.6 pycowsay
 pipx install --python python3.7 pycowsay
-pipx install git+https://github.com/ambv/black
-pipx install git+https://github.com/ambv/black.git@branch-name
-pipx install git+https://github.com/ambv/black.git@git-hash
-pipx install https://github.com/ambv/black/archive/18.9b0.zip
+pipx install git+https://github.com/psf/black
+pipx install git+https://github.com/psf/black.git@branch-name
+pipx install git+https://github.com/psf/black.git@git-hash
+pipx install https://github.com/psf/black/archive/18.9b0.zip
 pipx install black[d]
 pipx install --include-deps jupyter
 ```
@@ -27,10 +27,10 @@ pipx --version  # prints pipx version
 pipx run pycowsay  --version  # prints pycowsay version
 pipx --python pythonX pycowsay
 pipx --spec pycowsay==2.0 pycowsay --version
-pipx --spec git+https://github.com/ambv/black.git black
-pipx --spec git+https://github.com/ambv/black.git@branch-name black
-pipx --spec git+https://github.com/ambv/black.git@git-hash black
-pipx --spec https://github.com/ambv/black/archive/18.9b0.zip black --help
+pipx --spec git+https://github.com/psf/black.git black
+pipx --spec git+https://github.com/psf/black.git@branch-name black
+pipx --spec git+https://github.com/psf/black.git@git-hash black
+pipx --spec https://github.com/psf/black/archive/18.9b0.zip black --help
 pipx https://gist.githubusercontent.com/cs01/fa721a17a326e551ede048c5088f9e0f/raw/6bdfbb6e9c1132b1c38fdd2f195d4a24c540c324/pipx-demo.py
 ```
 
--- a/src/pipx/commands/commands.py
+++ b/src/pipx/commands/commands.py
@@ -3,9 +3,10 @@
 import logging
 import sys
 import textwrap
+import tempfile
 from pathlib import Path
 from shutil import which
-from typing import List
+from typing import List, Optional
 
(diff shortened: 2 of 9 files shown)


itsayellow:
@cs01 feel free to modify code if it seems fun to you.  Otherwise I'm happy to do the normal review process...

itsayellow:
Note, I may be out of communication for the next few days, but I may check in occasionally.

itsayellow:
Oh and before I forget, I was also thinking about making the `Venv` member variable `_python` public, meaning just changing its name to `python`.  Because I use it in commands.py in this PR.

cs01:
The _python name change is fine with me.

clbarnes:
For compatibility, `_python` could be made a property which throws a DeprecationWarning but still returns the right thing, then it can be removed in a couple of release's time.

itsayellow:
> For compatibility in the extremely unlikely event that anyone else is using it (was it in any previous releases?), `_python` could be made a property which throws a DeprecationWarning but still returns the right thing, then it can be removed in a couple of release's time.

AFAIK it was just an internal variable inside of Venv (hence the underscore).  I assume that the Venv author just assumed it would never be used outside.

I'm guessing that nobody was using it, and they were already "warned" when it had the starting underscore that they shouldn't.  😄

src/pipx/commands/commands.py
```diff
@@ -60,18 +70,47 @@ def install(
...
+            include_dependencies,
+            force=force,
         )
     except (Exception, KeyboardInterrupt):
         print("")
         venv.remove_venv()
         raise
 
 
+def _package_name_from_spec(package_spec: str, python: str) -> str:
```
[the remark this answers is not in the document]
itsayellow: Yes!  I used to use the normal `Venv.install_package()` which automatically did the animation, but when I refactored a bit I didn't realize I lost animation.

itsayellow:
Regarding the animation:
* I unquoted the package name, because other non-animated pipx messages refer to the package name unquoted and I was trying to stay consistent with them.  (e.g. black not 'black')
* The package specification __is__ quoted, because that made a lot of sense to me.
* I made a `util.py` function to return a string of the package name plus specification (if any) so that the format of this could be standardized across multiple different messages.

itsayellow:
I added your archived black zip-file to the list of tested install packages.  In the process I discovered and fixed a bug in `abspath_if_local()`.

itsayellow:
I also note (and not sure if this is a future PR or this one) that  `pipx list` currently does not mention a package specification and it would be helpful if they did.

itsayellow:
I think I addressed all of your bullet-point TODOs, ready for another review.

itsayellow:
I earmarked a testing TODO with newly-created Issue #303 that would be helpful to test `git+https://...` package specifications.

src/pipx/venv.py
```diff
@@ -212,6 +215,25 @@ def install_package(
         if self.package_metadata[package].package_version is None:
             raise PackageInstallFailureError
 
+    def install_package_no_deps(
+        self, package_or_url: str, pip_args: List[str]
+    ) -> Optional[str]:
+        package: Optional[str]
+
+        with animate(f"resolving {package_or_url!r}", self.do_animation):
```
cs01: `resolving` seems ambiguous to me. What do you think of something more descriptive like

* `determining package name from {...}`
* `resolving package name from {...}`
* `getting package name from {...}`

src/pipx/venv.py
```diff
@@ -292,11 +314,11 @@ def run_app(self, app: str, app_args: List[str]) -> int:
...
             return 130  # shell code for Ctrl-C
 
-    def _upgrade_package_no_metadata(
-        self, package_or_url: str, pip_args: List[str]
-    ) -> None:
-        with animate(f"upgrading package {package_or_url!r}", self.do_animation):
-            self._run_pip(["install"] + pip_args + ["--upgrade", package_or_url])
+    def _upgrade_package_no_metadata(self, package: str, pip_args: List[str]) -> None:
+        with animate(
+            "upgrading " + full_package_description(package, package), self.do_animation
```
cs01: nit: can f-strings be used instead of adding strings together (applies elsewhere in this PR as well)?

src/pipx/util.py
```diff
@@ -127,3 +128,17 @@ def run(cmd: Sequence[Union[str, Path]], check=True) -> int:
...
+
+def valid_pypi_name(package_name: str) -> bool:
+    # https://www.python.org/dev/peps/pep-0508/#names
+    return bool(
+        re.search(r"^([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])$", package_name, re.I)
+    )
+
+
+def full_package_description(package, package_spec):
+    name_str = f"{package}"
```
cs01: Why is this necessary?

src/pipx/util.py
```diff
@@ -127,3 +128,17 @@ def run(cmd: Sequence[Union[str, Path]], check=True) -> int:
...
+    if package != package_spec:
+        name_str += f" from spec. {package_spec!r}"
```
cs01: Remove `.` after `spec`

src/pipx/venv.py
```diff
@@ -173,21 +176,21 @@ def install_package(
...
+            pip_args = []
+        if package is None:
+            # If no package name is supplied, install only main package
+            #   first in order to see what its name is
+            package = self.install_package_no_deps(package_or_url, pip_args)
+
+        install_message = f"installing {package}"
+        if package != package_or_url:
+            install_message += f" from specification {package_or_url!r}"
+
```
cs01: Looks like this can be replaced with a call to `full_package_description()`?

docs/examples.md
```diff
@@ -27,10 +27,10 @@ pipx --version  # prints pipx version
...
 pipx --python pythonX pycowsay
 pipx --spec pycowsay==2.0 pycowsay --version
-pipx --spec git+https://github.com/ambv/black.git black
-pipx --spec git+https://github.com/ambv/black.git@branch-name black
-pipx --spec git+https://github.com/ambv/black.git@git-hash black
-pipx --spec https://github.com/ambv/black/archive/18.9b0.zip black --help
+pipx --spec git+https://github.com/psf/black.git black
+pipx --spec git+https://github.com/psf/black.git@branch-name black
+pipx --spec git+https://github.com/psf/black.git@git-hash black
+pipx --spec https://github.com/psf/black/archive/18.9b0.zip black --help
```
cs01: Not introduced in this PR but I just noticed these are all missing the `run` argument.

itsayellow:
> Just want to confirm that we are leaving the `run` command alone in this PR? i.e. it still has the optional `--spec` argument?

Confirmed, I left `run` alone.  Because the only required argument is app/package_name, the extra `--spec` option still makes sense as an additional option.

src/pipx/util.py
itsayellow: Ah you're right, I don't need the f-string if that's what you're referring to.  It fell out in the refactoring that way.

src/pipx/venv.py
itsayellow: Oops, actually those three lines can be deleted, I replaced install_message with `full_package_description()` in animate later on!

src/pipx/util.py
itsayellow: Done.

src/pipx/util.py
itsayellow: I simplified the function full_package_description and removed this.

docs/examples.md
itsayellow: True, and earlier examples in this block do not actually work (it won't work to put run options before `run`.) 

I'll fix them too.

src/pipx/venv.py
itsayellow: Yes, and "resolving" might be a little more loaded, it's probably too associated with package resolution.

Maybe determining package name?

src/pipx/venv.py
itsayellow: Done.

src/pipx/venv.py
itsayellow: Done.  Changed to "determining package name from "

src/pipx/venv.py
```diff
@@ -173,21 +176,20 @@ def install_package(
...
-        with animate(f"installing package {package_or_url!r}", self.do_animation):
-            if pip_args is None:
-                pip_args = []
+        if pip_args is None:
+            pip_args = []
+        if package is None:
+            # If no package name is supplied, install only main package
+            #   first in order to see what its name is
+            package = self.install_package_no_deps(package_or_url, pip_args)
+            # TODO 20191228: should we just fail if we cannot determine package?
```
itsayellow: I have a TODO here we should resolve.  I'm thinking if we cannot determine package name here we should just fail at this point before we try to install anything else.  Does that sound good?

itsayellow: Not sure if I need to start a review or not (github mechanics) but mostly I'm asking @cs01 and any other interested parties.

itsayellow:
I made install_package error out if package name cannot be determined.  I think this should be the wisest thing to do.

src/pipx/venv.py
```diff
@@ -173,32 +176,25 @@ def install_package(
...
-                installed_packages = self.list_installed_packages() - old_package_set
-                if len(installed_packages) == 1:
-                    package = installed_packages.pop()
-                    logging.info(f"Determined package name: '{package}'")
-                else:
-                    package = None
+                logging.warning(
+                    f"Cannot determine package name for spec {package_or_url!r}. "
+                )
+                raise PackageInstallFailureError
```
cs01: I agree an error should be raised if the package name cannot be determined. However, when a `PackageInstallFailureError` is raised, we could propagate more specific failure information to the user instead of unconditionally printing this message:
```
        except PackageInstallFailureError:
            venv.remove_venv()
            raise PipxError(
                f"Could not install package {package_name}. Is the name or spec correct?"
            )
```

I think it would be better if we could raise `PackageInstallFailureError` with specific failure text, and print that text directly to the user. i.e. `raise PackageInstallFailureError( f"Cannot determine package name for spec {package_or_url!r}.")` and then in `main.py` do

```
    except (PipxError, PackageInstallFailureError) as e:
```

src/pipx/commands/commands.py
```diff
@@ -60,18 +70,47 @@ def install(
...
+        return package_name
+
+    with tempfile.TemporaryDirectory() as temp_venv_dir:
+        venv = Venv(Path(temp_venv_dir), python=python)
+        venv.create_venv(venv_args=[], pip_args=[])
+        package_name = venv.install_package_no_deps(
+            package_or_url=package_spec, pip_args=[]
+        )
+        if package_name is None:
+            raise PipxError(f"Unable to validate {package_spec}")
```
cs01: I just ran into this locally while testing. 
```
>> pipx install gdbgui
Unable to validate gdbgui
```
This error message is technically correct, but I think a little more context and next steps would be helpful. Maybe something like
```
raise PipxError(f"Package was found and installed, but it is invalid. Verify the package  {package_spec} can be installed with pip and then try again.
```

itsayellow: I'm going to have to investigate that, because it shouldn't be getting to that error message because gdbgui is a valid pypi name and presumably not a local path for your.

itsayellow:
I revamped the package errors.  I subclassed `PackageInstallationFailureError` from `PipxError`.  Now I'm not sure if there's a value to having the separate `PackageInstallationFailureError`?

Currently if `pip install` subprocess fails, it will blab all of the pip errors to stderr.  It's a little raw, but I think useful, so I left it.  I could try and intercept this and send it to `logging.info()` if we wanted to clean things up a bit.

itsayellow:
I removed the specific `PackageInstallFailureError` because we were never specifically using it instead of the more general `PipxError`.

src/pipx/commands/commands.py
```diff
@@ -48,30 +58,51 @@ def install(
...
     except (Exception, KeyboardInterrupt):
         print("")
         venv.remove_venv()
         raise
 
 
+def _package_name_from_spec(package_spec: str, python: str) -> str:
+    start_time = time.time()
+
+    package_name: Optional[str]
```
cs01: I believe this is not Optional?

itsayellow: Yes!  thanks for pointing it out, I had meant to delete that.

PR #306, opened 2019-12-31
remove python_requires from setup.py

### LINKED ISSUE #237: publish version to PyPI for <= Python 3.5
**Describe the bug**
When users run `pip install pipx` with anything < Python 3.6, they will get the old pipx package, which is not related to this one. It is completely different, and very confusing that this occurs. Noted in https://twitter.com/mgedmin/status/1183682386103558144.

**How to reproduce**
Run
```
python3.5 -m pip install pipx
pipx --help
```

**Expected behavior**
An error should occur at install time saying that pipx is only supported for 3.6+. 

Removing the bad versions from PyPI is not a good idea since others may be depending on that package still.

cs01:
Test that python2 fails to install and prints a helpful error message
```
>> pip install .
Processing /home/csmith/git/pipx
    Complete output from command python setup.py egg_info:
    pipx requires Python 3.6 or later
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-XFhHFn-build/
```

fixes #237

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,8 +4,11 @@ import sys
 from setuptools import find_packages, setup  # type: ignore
 
 if sys.version_info < (3, 6, 0):
-    print("pipx requires Python 3.6 or later")
-    exit(1)
+    sys.exit(
+        "Python 3.6 or later is required. "
+        "See https://github.com/pipxproject/pipx "
+        "for installation instructions."
+    )
 
 import io  # noqa E402
 import os  # noqa E402
--- a/setup.py
+++ b/setup.py
@@ -4,8 +4,9 @@ import sys
 from setuptools import find_packages, setup  # type: ignore
 
 if sys.version_info < (3, 6, 0):
-    print("Python 3.6+ is required")
+    print("pipx requires Python 3.6 or later")
     exit(1)
+
 import io  # noqa E402
 import os  # noqa E402
 from pathlib import Path  # noqa E402
@@ -15,13 +16,13 @@ import re  # noqa E402
 
 CURDIR = Path(__file__).parent
 
-REQUIRED: List[str] = ["userpath", "argcomplete>=1.9.4, <2.0"]
+REQUIRED = ["userpath", "argcomplete>=1.9.4, <2.0"]  # type: List[str]
 
 with io.open(os.path.join(CURDIR, "README.md"), "r", encoding="utf-8") as f:
     README = f.read()
 
 
-def get_version() -> str:
+def get_version():
     main_file = CURDIR / "src" / "pipx" / "main.py"
     _version_re = re.compile(r"__version__\s+=\s+(?P<version>.*)")
     with open(main_file, "r", encoding="utf8") as f:
@@ -47,7 +48,6 @@ setup(
     scripts=[],
     entry_points={"console_scripts": ["pipx = pipx.main:cli"]},
     zip_safe=False,
-    python_requires=">=3.6",
     install_requires=REQUIRED,
     test_suite="tests.test_pipx",
     classifiers=[
--- a/src/pipx/__init__.py
+++ b/src/pipx/__init__.py
@@ -1,6 +1,7 @@
 import sys
 
-assert sys.version_info >= (3, 6, 0), (
-    "Python 3.6+ is required. See https://github.com/pipxproject/pipx "
-    "for installation instructions"
-)
+if sys.version_info < (3, 6, 0):
+    exit(
+        "Python 3.6+ is required. See https://github.com/pipxproject/pipx "
+        "for installation instructions."
+    )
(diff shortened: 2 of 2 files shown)


uranusjr:
Does pipx not publish a wheel? The message would only be shown when installing from sdist.

cs01:
It publishes a wheel for python3 only (https://pypi.org/project/pipx/#files), so in the case where < python 3.6 is used, https://github.com/pipxproject/pipx/blob/master/src/pipx/__init__.py#L3 should raise an assertion error. I just tested this locally:
```
>> python3.5 -m pipx -h
  File "/usr/local/lib/python3.5/runpy.py", line 183, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "/usr/local/lib/python3.5/runpy.py", line 142, in _get_module_details
    return _get_module_details(pkg_main_name, error)
  File "/usr/local/lib/python3.5/runpy.py", line 109, in _get_module_details
    __import__(pkg_name)
  File "/home/csmith/.local/lib/python3.5/site-packages/pipx/__init__.py", line 4, in <module>
    "Python 3.6+ is required. See https://github.com/pipxproject/pipx "
AssertionError: Python 3.6+ is required. See https://github.com/pipxproject/pipx for installation instructions
```

In the case where python2 is used to install pipx, I think pip will fall back to the sdist since there is no applicable wheel, and the error in this PR will be hit.

cs01:
I didn't like that a Python stack trace is presented to the user, so I changed the assertion to exit instead. So now the error looks like this:

```
>> python3.5 -m pipx -h
Python 3.6+ is required. See https://github.com/pipxproject/pipx for installation instructions.
```

uranusjr:
Ah OK, thanks for the clarification. LGTM.

gaborbernat:
I don't think this is a good Idea 🤔

cs01:
Could you elaborate?

gaborbernat:
Wouldn't this be detrimental for a dependency resolver? The whole point of this tag is that tools can find out if a package supports a python version without needing to build it.

gaborbernat:
I thought on the issue we agreed to release a patch for the old version, from a feature branch, not to do it on master 🤔

cs01:
> Wouldn't this be detrimental for a dependency resolver? The whole point of this tag is that tools can find out if a package supports a python version without needing to build it.

This is primarily so pip will install this version of pipx rather than the old one. We don't want any resolvers using the pipx app prior to the PyPI name transfer.

> I thought on the issue we agreed to release a patch for the old version, from a feature branch, not to do it on master

I think I misunderstood your comment on that issue. What do you mean by patch for the old version? 

After publishing this to PyPI, people who install pipx with < Python 3.6 will get this pipx version and see the error message. I can then revert this change on master, publish to PyPI and new versions will go back to only supporting >= 3.6, but users who erroneously install with < 3.6 will still get the version with the error message.

gaborbernat:
> I think I misunderstood your comment on that issue. What do you mean by patch for the old version? AFAIK PyPI doesn't support anything like that. IMO making the change on a branch from an old commit and publishing makes the git history messier 

What's the old version number causing issues?

> and doesn't really add anything over doing it on master.

It does all these:

> Wouldn't this be detrimental for a dependency resolver? The whole point of this tag is that tools can find out if a package supports a python version without needing to build it.

itsayellow:
> After publishing this to PyPI, people who install pipx with < Python 3.6 will get this pipx version and see the error message. I can then revert this change on master, publish to PyPI and new versions will go back to only supporting >= 3.6, but users who erroneously install with < 3.6 will still get the version with the error message.

I like this 2-step pypi solution.  
* It leaves the old code (from another project) untouched and able to be used by people specifically trying to find it.
* It causes naive users to get a version of the modern pipx that will point them in the right direction.
* It will the put back into place normal `python_requires` from that point on so all versions except the one interim one supports static information for dependency resolution.

cs01:
The old version numbers that cause problems are 0.1.0 and 0.1.1.

I think what you're saying is to fork the [old pipx](https://github.com/khalti/pipx) and change it somehow, but I'm sill not clear on what you're saying to do with it. If we branch it and remove support for certain python versions in it, then publish to 0.1.2, the dependency resolver will still use the last available 0.1.0 or 0.1.1, which is what is currently causing issues now. On top of that, modifying that project's code and re-publishing with error messages related to modern pipx seems like it will just cause confusion.

gaborbernat:
I was thinking of option 2 from above. We really only need to modify setup.py, it doesn't even have to be the old project just a single setup.py file that raises that error. Sure we could also alter master, release and then backout but that might throw off dependency resolvers 🤔 I was thinking of a single step solution that just releases a setup.py that always fails with 0.1.2 and notes one should use Python 3 🤷‍♂️

cs01:
I still don't see how that is any better for dependency resolvers and prefer merging this PR as-is.

PR #308, opened 2020-01-03
[bugfix] fix 'delay' in animate api

cs01:
Delay is an argument to functions, but wasn't used. This PR actually implements the delay.

### DIFF
--- a/tests/test_animate.py
+++ b/tests/test_animate.py
@@ -30,8 +30,8 @@ def test_delay_suppresses_output(capsys, monkeypatch):
 
     test_string = "asdf"
 
-    with pipx.animate.animate(test_string, do_animation=True, delay=0.1):
-        time.sleep(0.05)
+    with pipx.animate.animate(test_string, do_animation=True, delay=0.9):
+        time.sleep(0.5)
     captured = capsys.readouterr()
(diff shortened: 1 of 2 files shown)


itsayellow:
The code LGTM.

I would think adding at least one test is nice.  Could you just make a test that delays the start of animation, and the code inside the context finishes before that delay, so that you can verify that there was no animation printed?  I think that might be the best we can do, given threads/timing/precision constraints.

tests/test_animate.py
```diff
@@ -25,6 +25,17 @@ def check_animate_output(
     assert captured.err[:chars_to_test] == expected_string[:chars_to_test]
 
 
+def test_delay_suppresses_output(capsys, monkeypatch):
+    monkeypatch.setattr(pipx.animate, "stderr_is_tty", True)
+
+    test_string = "asdf"
+
+    with pipx.animate.animate(test_string, do_animation=True, delay=0.1):
```
itsayellow: I think you might need more time buffer here.  I started out with only 0.1 extra time in `check_animate_output`, and occasionally (especially on Travis' Windows testers) I would get spurious failures.  I think you may need a `delay` of a few tenths of a second more than the `sleep()` time at least.

PR #310, opened 2020-01-03
update changelog; remove build dir before publishing

cs01:
* Add other changes since 0.14 to changelog
* Remove two spaces after periods
* Remove build dir when publishing (discovered this was causing errors when I was about to publish 0.15rc)


cs01:
You have been doing all the impactful stuff so all I have left is to nitpick grammar, haha 😆

PR #314, opened 2020-01-05
Add Python 3.8 to classifiers in setup.py.

itsayellow:
I noticed that the latest release doesn't specifically list Python 3.8 alongside other python versions.

Since we actually test for it I figured it was worth adding.


itsayellow:
Oh, my mistake, now that I look I don't see python3.8 in our tests.  Maybe we should add it?

itsayellow:
Oh, now I see my confusion, noxfile.py has python 3.8 tests, but .travis,.yml does not.

PR #316, opened 2020-01-05
Add upgrade from pre-0.15.0.0 upgrade help, move Installation Options.

itsayellow:
The primary motivation was to add the pre- to post- 0.15.0.0 upgrade special instructions to the detailed install page.

In addition, I moved Installation Options (h3) to inside of the Install (h2) section.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -26,7 +26,7 @@ python3 -m pip install -U pipx
 ### Note: Upgrading pipx from a pre-0.15.0.0 version to 0.15.0.0 or later
 When upgrading to pipx 0.15.0.0 or above from a pre-0.15.0.0 version, you must re-install all packages to take advantage of the new persistent pipx metadata files introduced in the 0.15.0.0 release. These metadata files store pip specification values, injected packages, any custom pip arguments, and more in each main package's venv.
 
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -13,15 +13,22 @@ python3 -m pip install --user pipx
...
 python3 -m pip install -U pipx
 ```
 
-### Installation Options
-pipx's default binary location is `~/.local/bin`. This can be overriden with the environment variable `PIPX_BIN_DIR`.
+### Note: Upgrading pipx from a pre-0.15.0.0 version to 0.15.0.0 or later
+When upgrading to pipx 0.15.0.0 or above from a pre-0.15.0.0 version, you must re-install all packages to take advantage of the new persistent pipx metadata files introduced in the 0.15.0.0 release. These metadata files store pip specification values, injected packages, any custom pip arguments, and more in each main package's venv.
 
-pipx's default virtual environment location is `~/.local/pipx`. This can be overridden with the environment variable `PIPX_HOME`.
+If you have no packages installed from a VCS source and no venvs with injected packages, you can do this by running `pipx reinstall-all`
```
cs01: Add period at the end. Other than that LGTM, feel free to merge.

itsayellow:
I changed "VCS..." to talking about using a `--spec` because it encompasses more situations and is more accurate.  Look good?

PR #317, opened 2020-01-06
allow direct invocation from src directory

cs01:
After looking at pip's instructions (https://pip.pypa.io/en/stable/development/getting-started/) I thought pipx could benefit from a similar contributor workflow. 

With this change, pipx can be run from source with `python src/pipx --version` with no setup, and no ambiguity about what is being run. The contributing instructions were updated to reflect this.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -1,46 +1,27 @@
 Thanks for your interest in contributing to pipx!
 
+## Running pipx From Source Tree
+To run the pipx executable from your source tree during development, run pipx from the src directory:
 
-## Dependencies
+```
+python src/pipx --version
+```
+
+## Running Tests
+
+### Setup
 pipx uses an automation tool called [nox](https://pypi.org/project/nox/) for development, continuous integration testing, and various tasks.
 
 `nox` defines tasks or "sessions" in `noxfile.py` which can be run with `nox -s SESSION_NAME`. Session names can be listed with `nox -l`.
 
 Install nox for pipx development:
 ```
-pip install --user git+https://github.com/cs01/nox.git@5ea70723e9e6 nox
+python -m pip install --user git+https://github.com/cs01/nox.git@5ea70723e9e6 nox
 ```
 
 !!! note
     A specific version of nox must be used for pipx development until [nox issue 233](https://github.com/theacodes/nox/issues/233) is resolved.
 
-
-## Developing pipx
-Clone pipx, then enter the root of the pipx repository. Run this command to see which commands you can run with nox.
-```
-nox -l
-```
-
-You should see some options that look like this
-```
-- develop-3.6
-- develop-3.7
-- develop-3.8
-```
-
-Choose the Python version you want to develop in and run the following commands, modifying with your version as necessary.
-```
-nox -s develop-3.8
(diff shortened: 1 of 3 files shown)


itsayellow:
I'm looking at "Files changed", and I'm wondering why it doesn't show the docs/installation.md commits as being already in master?  They show up as new in this PR.

docs/contributing.md
```diff
@@ -119,14 +100,16 @@ nox -s watch_docs
 nox -s publish_docs
 ```
 
-## Release New `pipx` Version
+## Releasing New `pipx` Versions
 To create a new release
 
+* update pipx's version in `__main__.py` and regenerate documentation
```
itsayellow: Where is the version specified in `__main__.py`?  And why not `setup.py`?

cs01:
Not sure what the cause of that was, but it went away after pulling from origin/master.

docs/contributing.md
```diff
@@ -1,46 +1,27 @@
...
-```
-
-A virtual environment with required dependencies is now sandboxed and activated!
-
-Go ahead and make your changes now. You can confirm they still work by running unit tests (see the next section).
-
-To leave the environment, type `deactivate`.
-
-## Testing pipx locally
 Tests are defined as `nox` sessions. You can see all nox sessions with
```
itsayellow: Why did you delete the section describing the develop nox sessions?

docs/installation.md
```diff
@@ -13,15 +13,22 @@ python3 -m pip install --user pipx
...
+
 ## Upgrade pipx
 ```
 python3 -m pip install -U pipx
 ```
 
-### Installation Options
-pipx's default binary location is `~/.local/bin`. This can be overriden with the environment variable `PIPX_BIN_DIR`.
+### Note: Upgrading pipx from a pre-0.15.0.0 version to 0.15.0.0 or later
+When upgrading to pipx 0.15.0.0 or above from a pre-0.15.0.0 version, you must re-install all packages to take advantage of the new persistent pipx metadata files introduced in the 0.15.0.0 release. These metadata files store pip specification values, injected packages, any custom pip arguments, and more in each main package's venv.
```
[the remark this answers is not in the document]
itsayellow: This comment got placed as outdated because docs/installation.md is now not marked as changed, but I think it's changing this would still be good, either here or with another PR.

docs/contributing.md
itsayellow: Oh I think maybe you mean `main.py` ?

itsayellow: And I see the fact that `setup.py` no longer has a hard-coded version.

docs/contributing.md
cs01: Because the code can be tested directly without creating a virtual environment (with exceptions around dependent packages userpath and argcomplete, though I assume those are installed since I assume the user has already installed pipx).

docs/contributing.md
cs01: Yup I meant `main.py`, thanks.

itsayellow:
I like it.  I think this will make it easier for the maintainers to test bugs (i.e. switch between the current dev and the release version that a bug report is referring to.)

PR #321, opened 2020-01-06
pass pip_args and verbose flag when determining package name

### LINKED ISSUE #320: pip-args ignored for pip install that determines package name
**Describe the bug**

When installing a package from a source such as a git repository, pipx now does a pip install dance to figure out the package name.

Unfortunately, it fails to propagate `--pip-args` specified in the `install` command when doing this. If the pip-args are used for something like specifying the `--extra-index-url` for a dependency this will cause installation to fail.

This seems to occur due to this line: https://github.com/pipxproject/pipx/blob/fa570d50049e0f60789ad631d0c7d505b171d82b/src/pipx/commands/commands.py#L97

**How to reproduce**

`pipx install --pip-args <arguments> <vcs url>`

**Expected behavior**

`<arguments>` are passed to the pip install command that is used to infer the package name.

cs01:
This passes pip arguments to the command that determines the package name. It also passes the verbose flag to the venv creation which suppresses the animation output when running in verbose mode. These were regressions in the 0.15 release.

cc @tsiq-rob, @tsiq-alberto

fixes #320

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ See Contributing for how to update this file.
 <a href="https://travis-ci.org/pipxproject/pipx"><img src="https://travis-ci.org/pipxproject/pipx.svg?branch=master" /></a>
 
 <a href="https://pypi.python.org/pypi/pipx/">
-<img src="https://img.shields.io/badge/pypi-0.15.0.0-blue.svg" /></a>
+<img src="https://img.shields.io/badge/pypi-0.15.0.1-blue.svg" /></a>
 <a href="https://github.com/ambv/black"><img alt="Code style: black" src="https://img.shields.io/badge/code%20style-black-000000.svg"></a>
 </p>
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,3 +1,7 @@
+0.15.0.1
+
+- [bugfix] pass pip arguments to pip when determining package name
+
 0.15.0.0
 
 Upgrade instructions: When upgrading to 0.15.0.0 or above from a pre-0.15.0.0 version, you must re-install all packages to take advantage of the new persistent pipx metadata files introduced in this release. These metadata files store pip specification values, injected packages, any custom pip arguments, and more in each main package's venv. You can do this by running `pipx reinstall-all` or `pipx uninstall-all`, then reinstalling manually.
--- a/docs/index.md
+++ b/docs/index.md
@@ -21,7 +21,7 @@ See Contributing for how to update this file.
 <a href="https://travis-ci.org/pipxproject/pipx"><img src="https://travis-ci.org/pipxproject/pipx.svg?branch=master" /></a>
 
 <a href="https://pypi.python.org/pypi/pipx/">
-<img src="https://img.shields.io/badge/pypi-0.15.0.0-blue.svg" /></a>
+<img src="https://img.shields.io/badge/pypi-0.15.0.1-blue.svg" /></a>
 <a href="https://github.com/ambv/black"><img alt="Code style: black" src="https://img.shields.io/badge/code%20style-black-000000.svg"></a>
 </p>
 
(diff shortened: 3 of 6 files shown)


cs01:
Since this is a serious bug, I also updated the pipx version to 0.15.0.1 and will publish a new version to PyPI once this is reviewed.

tsiq-rob:
Thank you for such a swift follow up :)

cs01:
Until we set up a test server with deterministic package content (which will have the nice side effect of speeding up all of pipx's tests), I'm not sure that adds any value to our unit tests. If you disagree, could you go into more detail on how you are proposing the test work?

`_package_name_from_spec()` now requires the arguments, so if they are not passed lint tests will fail. I guess I could write unit tests for that function?

itsayellow:
Mostly I just somebody to try out the code (aka integration test) to make sure that it will fix the bug.  

Does the OP of the bug (@tsiq-rob) have an example we could use to verify that the new code works?

It doesn't have to be a unit test (although there existed another index it would be nice.)

cs01:
Oh I already did that

```
> pipx install git+https://github.com/psf/black  --verbose --force --pip-args='--extra-index-url=https://test.pypi.org/simple/'
pipx > (run_subprocess:107): running /home/csmith/git/pipx/venv/bin/python3.7 -m venv --without-pip /tmp/tmpd759jwj2
pipx > (run_subprocess:107): running /tmp/tmpd759jwj2/bin/python --version
pipx > (run_subprocess:107): running /tmp/tmpd759jwj2/bin/python -m pip list --format=json
pipx > (run_subprocess:107): running /tmp/tmpd759jwj2/bin/python -m pip install --extra-index-url=https://test.pypi.org/simple/ --no-dependencies git+https://github.com/psf/black
Looking in indexes: https://pypi.org/simple, https://test.pypi.org/simple/
Collecting git+https://github.com/psf/black
  Cloning https://github.com/psf/black to /tmp/pip-req-build-6o1xlp2e
  Running command git clone -q https://github.com/psf/black /tmp/pip-req-build-6o1xlp2e
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
Building wheels for collected packages: black
  Building wheel for black (PEP 517) ... done
  Created wheel for black: filename=black-19.10b1.dev16+g1ab87a3-cp37-none-any.whl size=101532 sha256=4a2b27e6e44236e7b444ecb9761792f1e5e65edcbdf3dc07524ccbe2f6de7be8
  Stored in directory: /tmp/pip-ephem-wheel-cache-yvn3dbm9/wheels/80/17/07/88c1dba6753a7314e671700d8ff6d15770a8f1f61a925b181f
Successfully built black
Installing collected packages: black
Successfully installed black-19.10b1.dev16+g1ab87a3
pipx > (run_subprocess:107): running /tmp/tmpd759jwj2/bin/python -m pip list --format=json
pipx > (install_package_no_deps:232): Determined package name: 'black'
pipx > (_package_name_from_spec:104): Package name determined in 9.7s
```

itsayellow:
OK, that's all I was thinking.

Although, unfortunately this example also works with our current code (just tried it) because black is both on regular pypi as well as the `--extra-index-url` test.pypi.org/simple.

But I guess at least everything seems to still be working.  I'm glad that --verbose works properly now, I noticed something weird in the formatting before but I didn't register it was my code's fault...

LGTM

cs01:
If you inspect the commands it ran, you can see it passed the new arguments when determining the package name. 

I just thought of a way to test this: Pass an invalid argument to pip, then confirm `Cannot determine package name from spec` is in the output. Then remove the invalid argument and confirm the installation succeeds.

itsayellow:
For future reference (not gating this PR's merge) this uses a Github Pages URL to mock a simple pypi index.  If all we need is a separate index online.
https://www.freecodecamp.org/news/how-to-use-github-as-a-pypi-server-1c3b0d07db2/

Issue #322, opened 2020-01-07
pipx install --editable fails during package determination


itsayellow:
**Describe the bug**
Regression from #302.  I think significant.

Background: `pip install --editable` needs either a local directory or VCS spec right after the `--editable` switch.

The #302 code added `--no-dependencies` after all pip arguments when determining the package name.  In editable pipx installs, this has the effect of putting `--no-dependencies` between `--editable` and the package specification in the `pip install` arguments, which is illegal for pip, causing package determination to fail.

In addition, `main.py` code puts `pip_args` after `--editable` which would cause the same bug if both `--editable` and `--pip_args` are used.

**How to reproduce**
```
carvel:pipx pipx install --verbose -e git+https://github.com/itsayellow/datadiary
pipx > (run_subprocess:108): running /usr/local/opt/python/bin/python3.7 -m venv --without-pip /var/folders/5n/rx81rh0d35nbq92pjf3jpl2r0000gn/T/tmpu4hpo780
pipx > (run_subprocess:108): running /var/folders/5n/rx81rh0d35nbq92pjf3jpl2r0000gn/T/tmpu4hpo780/bin/python --version
pipx > (run_subprocess:108): running /var/folders/5n/rx81rh0d35nbq92pjf3jpl2r0000gn/T/tmpu4hpo780/bin/python -m pip list --format=json
pipx > (run_subprocess:108): running /var/folders/5n/rx81rh0d35nbq92pjf3jpl2r0000gn/T/tmpu4hpo780/bin/python -m pip install --editable --no-dependencies git+https://github.com/itsayellow/datadiary
ERROR: --no-dependencies is not a valid editable requirement. It should either be a path to a local project or a VCS URL (beginning with svn+, git+, hg+, or bzr+).
pipx > (install_package_no_deps:222): '/var/folders/5n/rx81rh0d35nbq92pjf3jpl2r0000gn/T/tmpu4hpo780/bin/python -m pip install --editable --no-dependencies git+https://github.com/itsayellow/datadiary' failed
Cannot determine package name from spec 'git+https://github.com/itsayellow/datadiary'. Check package spec for errors.
carvel:pipx 
```

**Expected behavior**

Package should have been installed normally

itsayellow:
Fixed by #323

PR #323, opened 2020-01-07
pipx install --editable: fix regression

itsayellow:
This is for #322 

There was a regression that occurred with #302 which made editable installations fail.

`--editable` needs either a local directory or VCS spec right after it to be valid for pip.

This PR puts the `--no-dependencies` option FIRST in the pip args inside of `venv.py` `install_package_no_deps()`, so that `--editable` is still last in options and right before the package specification.

This PR also reorders pip args in `main.py` `get_pip_args()` so that `--editable` is sure to be last (formerly, it used to precede pipx's `--pipx_args`.)

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,3 +1,7 @@
+dev
+
+- [bugfix] fix regression that caused installing with --editable flag to fail package name determination.
+
 0.15.1.0
 
 - Add Python 3.8 to PyPI classifier and travis test matrix
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -102,11 +102,13 @@ def get_pip_args(parsed_args: Dict) -> List[str]:
     if parsed_args.get("index_url"):
         pip_args += ["--index-url", parsed_args["index_url"]]
 
-    if parsed_args.get("editable"):
-        pip_args += ["--editable"]
-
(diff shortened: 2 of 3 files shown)


cs01:
Do you have a PyPI account? I can add you as a maintainer so you can publish new versions. This can be published in a minor version update, 0.15.1.1.

cs01:
FYI @gaborbernat now has maintainer rights to push new versions to PyPI in case I'm not available.

itsayellow:
I do have a pypi account (itsayellow) and would be happy to publish (although it's been a while since I published to pypi).

Reading the Contributing page, the steps are (correct if I'm wrong):
* update version in `main.py`
* nox -s publish
* nox -s publish_docs

That's it, correct?

cs01:
Great! That is correct. The only thing is you need to generate documentation after changing the version in `main.py` to propagate the version to the docs.

itsayellow:
Can I publish from my fork?

cs01:
Oh I guess there is a little more to it:

* Always publish from master
* After publishing go to GitHub -> releases -> Draft new release

itsayellow:
Ah darn, I already did it from my fork/branch.

itsayellow:
Fixes #322

PR #326, opened 2020-01-08
In upgrade-all, skip --editable packages

### LINKED ISSUE #307: Editable pipx venvs get pip error during upgrade-all
**Describe the bug**
Running `pipx upgrade-all`, any venvs using the `--editable` flag set cause an error during pip upgrade because in pip `--editable` and `--upgrade` conflict.

**How to reproduce**
```
git clone https://github.com/itsayellow/pytivo
pipx install -e pytivo/
pipx upgrade-all
```

produces:
```
pipx upgrade-all
ERROR: --upgrade is not a valid editable requirement. It should either be a path to a local project or a VCS URL (beginning with svn+, git+, hg+, or bzr+).
Error encountered when upgrading pytivo
```

**Expected behavior**
`pipx upgrade-all` should detect that a venv is in `--editable` mode and just skip over it for.

itsayellow:
Fixes #307 

Change to skip any packages with `--editable` switch during `upgrade-all`

### DIFF
--- a/src/pipx/commands/upgrade.py
+++ b/src/pipx/commands/upgrade.py
@@ -97,7 +97,7 @@ def upgrade_all(
         num_packages += 1
         package = venv_dir.name
         venv = Venv(venv_dir, verbose=verbose)
-        if package in skip:
+        if package in skip or "--editable" in venv.pipx_metadata.main_package.pip_args:
             continue
         try:
             packages_upgraded += upgrade(
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,6 +1,7 @@
 dev
 
 - [bugfix] Fix recursive search of dependencies' apps so no apps are missed.
+- `upgrade-all` now skips editable packages, because pip disallows upgrading editable packages.
 
 0.15.1.1
 


src/pipx/commands/upgrade.py
```diff
@@ -97,7 +97,7 @@ def upgrade_all(
         num_packages += 1
         package = venv_dir.name
         venv = Venv(venv_dir, verbose=verbose)
-        if package in skip:
+        if package in skip or "--editable" in venv.pipx_metadata.main_package.pip_args:
```
cs01: Should probably check for the short version of `--editable`, `-e`

itsayellow: I thought of that, but I think it's ok.  In `main.py` `get_pip_args()` specifically adds only `--editable` to pip args:
```
    # make sure --editable is last because it needs to be right before
    #   package specification
    if parsed_args.get("editable"):
        pip_args += ["--editable"]
```

So for both `-e` or `--editable` sent to pipx, only `--editable` ends up in `pip_args`.

cs01: Sounds good. I guess the only other way would be if a if a user manually passes the `-e` flag to pipx's `--pip-args` argument, effectively sidestepping pipx's editable flag. I don't think that's likely though.

cs01:
Please update the changelog before merging

src/pipx/commands/upgrade.py
itsayellow: Ah right about the `--pip_args`.

PR #327, opened 2020-01-08
0.15.1.2

itsayellow:
fixes app search in dependencies, fixes minor bug in upgrade-all

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ See Contributing for how to update this file.
 <a href="https://travis-ci.org/pipxproject/pipx"><img src="https://travis-ci.org/pipxproject/pipx.svg?branch=master" /></a>
 
 <a href="https://pypi.python.org/pypi/pipx/">
-<img src="https://img.shields.io/badge/pypi-0.15.1.1-blue.svg" /></a>
+<img src="https://img.shields.io/badge/pypi-0.15.1.2-blue.svg" /></a>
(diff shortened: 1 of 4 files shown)


itsayellow:
I could successfully `publish_docs` after:
* change url for remote `upstream` from `https://github.com/pipxproject/pipx.git` to `git@github.com:pipxproject/pipx.git`
* `git checkout --track upstream/gh-pages`
* checkout `master`, run `nox -s publish_docs`
* checkout `gh-pages`, manually push

`nox -s publish_docs` says it is pushing, but somehow nothing showed up on the pipxproject/pipx gh-pages branch until I manually pushed.  Not sure why.

cs01:
Is it using the mkdocs.yml configuration? If not, maybe that will fix it. If so, maybe it can be tweaked?

cs01:
Congrats on doing a pipx release on your own 🎉

itsayellow:
> Is it using the mkdocs.yml configuration? If not, maybe that will fix it. If so, maybe it can be tweaked?

From my reading of the mkdocs documentation, it should use it if it's in the project dir.  But I don't see anything in pipx's `mkdocs.yml` that pertains to gh-deploy, so maybe we need to configure more?

itsayellow:
Ah I see what happened.  Even though I specifically made my gh-pages branch track upstream/gh-pages, `mkdocs gh-deploy` still seems to have specifically pushed to origin/gh-pages.

Ah and I see why.  https://www.mkdocs.org/user-guide/configuration/#remote_name  `remote_name` defaults to origin for gh-deploy.  Let me make a pull request to fix mkdocs.yml.

PR #335, opened 2020-01-13
Fix problems in install with user-specified python involving `__PYVENV_LAUNCHER__`

### LINKED ISSUE #334: FileNotFoundError for pipx_shared.pth with `pipx install --python` and pyenv
**Describe the bug**

I have pipx installed from Homebrew (yay!), as well as pyenv. When I run `pipx install black`, the resulting venv using the python3.7 interpreter from Homebrew. When I run `pipx install --python python3.8 black` (or any other version, with any other package), I get the error mentioned in the issue title and reproduced below.

**How to reproduce**

```
$ pipx install --verbose --python python3.8 black
pipx > (_package_name_from_spec:93): Determined package name: black
pipx > (_package_name_from_spec:94): Package name determined in 0.0s
pipx > (run_subprocess:108): running python3.8 -m venv --without-pip /Users/brian/.local/pipx/venvs/black

pipx > (rmdir:18): removing directory /Users/brian/.local/pipx/venvs/black
  File "/usr/local/Cellar/pipx/0.15.1.1/libexec/bin/pipx", line 11, in <module>
    load_entry_point('pipx==0.15.1.1', 'console_scripts', 'pipx')()
  File "/usr/local/Cellar/pipx/0.15.1.1/libexec/lib/python3.7/site-packages/pipx/main.py", line 590, in cli
  File "/usr/local/Cellar/pipx/0.15.1.1/libexec/lib/python3.7/site-packages/pipx/main.py", line 172, in run_pipx_command
    include_dependencies=args.include_deps,
  File "/usr/local/Cellar/pipx/0.15.1.1/libexec/lib/python3.7/site-packages/pipx/commands/commands.py", line 62, in install
    venv.create_venv(venv_args, pip_args)
  File "/usr/local/Cellar/pipx/0.15.1.1/libexec/lib/python3.7/site-packages/pipx/venv.py", line 139, in create_venv
    pipx_pth.write_text(f"{shared_libs.site_packages}\n", encoding="utf-8")
  File "/usr/local/Cellar/python/3.7.6_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/pathlib.py", line 1235, in write_text
    with self.open(mode='w', encoding=encoding, errors=errors) as f:
  File "/usr/local/Cellar/python/3.7.6_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/pathlib.py", line 1203, in open
    opener=self._opener)
  File "/usr/local/Cellar/python/3.7.6_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/pathlib.py", line 1058, in _opener
    return self._accessor.open(self, flags, mode)
FileNotFoundError: [Errno 2] No such file or directory: '/Users/brian/.local/pipx/venvs/black/lib/python3.7/site-packages/pipx_shared.pth'

$ pipx --version
0.15.1.1

$ pyenv version
3.8.0 (set by /Users/brian/.pyenv/version)
2.7.15 (set by /Users/brian/.pyenv/version)
3.7.4 (set by /Users/brian/.pyenv/version)
3.6.9 (set by /Users/brian/.pyenv/version)

$ pyenv which python3.8
/Users/brian/.pyenv/versions/3.8.0/bin/python3.8
```

**Expected behavior**

pipx should install a package using the python3.8 interpreter from pyenv.

I get the same error using an absolute path with `--python $(pyenv which python3.8)`, but `--python $(brew --prefix python@3.8)/bin/python3` works.

itsayellow:
Fixes #334

This removes `__PYVENV_LAUNCHER__` from the env of subprocess commands.  `__PYVENV_LAUNCHER__` is used for the python main process to communicate to a subprocess it spawns, and is not meant to be in the environment for a user running python.  It was set in pipx's inherited env for certain python builds, and it sometimes caused any argument-supplied python binary to be ignored in favor of the python interpreter in this environment variable.

In the process of fixing this bug, I converted `get_script_output()` to use `run_subprocess()`, so `run_subprocess()` is the only function needing to carefully control environment variables (and simplifying the `get_script_output()` code in the process).

There is only one other situation in `util.py` that uses `subprocess.run()`, and that is in `run_pypackage_bin()`.  I left it alone, but it may also suffer from similar issues.  It is so infrequently used (only with running an app in a local `__pypackages__` directory), and not tested, so I was a little afraid to mess with it.

### DIFF
--- a/src/pipx/util.py
+++ b/src/pipx/util.py
@@ -75,7 +75,7 @@ else:
         return bin_path, python_path
 
 
-def get_script_output(interpreter: Path, script: str, *args) -> str:
+def get_script_output(interpreter: Path, script: str, *args: Union[str, Path]) -> str:
     proc = run_subprocess([interpreter, "-c", script, *args], capture_stderr=False)
     return proc.stdout
 
--- a/src/pipx/util.py
+++ b/src/pipx/util.py
@@ -95,12 +95,11 @@ def run_subprocess(
     """Run arbitrary command as subprocess, capturing stderr and stout"""
     env = dict(os.environ)
 
-    # Remove PYTHONPATH because some platforms (macOS with Homebrew) add
-    #   pipx directories to it, and can make it appear to venvs as though
-    #   pipx dependencies are in the venv path (#233)
-    # Remove __PYVENV_LAUNCHER__ because it is only meant to be communication
-    #   between main python binary and its subprocess and can cause the wrong
-    #   python binary to be used (#334)
+    # Remove PYTHONPATH because some platforms (macOS with Homebrew) add pipx
+    #   directories to it, and can make it appear to venvs as though pipx
+    #   dependencies are in the venv path (#233)
+    # Remove __PYVENV_LAUNCHER__ because it can cause the wrong python binary
+    #   to be used (#334)
     env_blacklist = ["PYTHONPATH", "__PYVENV_LAUNCHER__"]
     for env_to_remove in env_blacklist:
         env.pop(env_to_remove, None)
--- a/src/pipx/util.py
+++ b/src/pipx/util.py
@@ -93,16 +93,18 @@ def run_subprocess(
     capture_stderr: bool = True,
 ) -> subprocess.CompletedProcess:
     """Run arbitrary command as subprocess, capturing stderr and stout"""
-
     env = dict(os.environ)
+
     # Remove PYTHONPATH because some platforms (macOS with Homebrew) add
     #   pipx directories to it, and can make it appear to venvs as though
     #   pipx dependencies are in the venv path (#233)
-    env.pop("PYTHONPATH", None)
     # Remove __PYVENV_LAUNCHER__ because it is only meant to be communication
     #   between main python binary and its subprocess and can cause the wrong
     #   python binary to be used (#334)
-    env.pop("__PYVENV_LAUNCHER__", None)
+    env_blacklist = ["PYTHONPATH", "__PYVENV_LAUNCHER__"]
+    for env_to_remove in env_blacklist:
+        env.pop(env_to_remove, None)
+
     env["PIP_DISABLE_PIP_VERSION_CHECK"] = "1"
     cmd_str = " ".join(str(c) for c in cmd)
     logging.info(f"running {cmd_str}")
--- a/src/pipx/util.py
+++ b/src/pipx/util.py
@@ -76,14 +76,8 @@ else:
 
 
 def get_script_output(interpreter: Path, script: str, *args) -> str:
-    # Make sure that Python writes output in UTF-8
-    env = os.environ.copy()
-    env["PYTHONIOENCODING"] = "utf-8"
-    env.pop("__PYVENV_LAUNCHER__", None)
-    output = subprocess.run(
-        [str(interpreter), "-c", script, *args], stdout=subprocess.PIPE, env=env
-    ).stdout.decode(encoding="utf-8")
-    return output
+    proc = run_subprocess([str(interpreter), "-c", script, *args], capture_stderr=False)
+    return proc.stdout
 
 
 def get_site_packages(python: Path) -> Path:
--- a/src/pipx/util.py
+++ b/src/pipx/util.py
@@ -76,7 +76,7 @@ else:
 
 
 def get_script_output(interpreter: Path, script: str, *args) -> str:
-    proc = run_subprocess([str(interpreter), "-c", script, *args], capture_stderr=False)
+    proc = run_subprocess([interpreter, "-c", script, *args], capture_stderr=False)
     return proc.stdout
 
 
--- a/src/pipx/util.py
+++ b/src/pipx/util.py
@@ -100,10 +100,14 @@ def run_subprocess(
 ) -> subprocess.CompletedProcess:
     """Run arbitrary command as subprocess, capturing stderr and stout"""
 
-    # Null out PYTHONPATH because some platforms (macOS with Homebrew) add
+    env = dict(os.environ)
+    # Remove PYTHONPATH because some platforms (macOS with Homebrew) add
     #   pipx directories to it, and can make it appear to venvs as though
     #   pipx dependencies are in the venv path (#233)
-    env = {k: v for k, v in os.environ.items() if k.upper() != "PYTHONPATH"}
+    env.pop("PYTHONPATH", None)
+    # Remove __PYVENV_LAUNCHER__ because it is only meant to be communication
+    #   between main python binary and its subprocess and can cause the wrong
+    #   python binary to be used (#334)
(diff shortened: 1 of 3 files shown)


src/pipx/util.py
```diff
@@ -75,14 +75,9 @@ def get_venv_paths(root: Path) -> Tuple[Path, Path]:
         return bin_path, python_path
 
 
-def get_script_output(interpreter: Path, script: str, *args) -> str:
-    # Make sure that Python writes output in UTF-8
-    env = os.environ.copy()
-    env["PYTHONIOENCODING"] = "utf-8"
```
cs01: This should be applied in the `run_subprocess` function

src/pipx/util.py
```diff
@@ -75,14 +75,9 @@ def get_venv_paths(root: Path) -> Tuple[Path, Path]:
...
-    output = subprocess.run(
-        [str(interpreter), "-c", script, *args], stdout=subprocess.PIPE, env=env
-    ).stdout.decode(encoding="utf-8")
-    return output
+def get_script_output(interpreter: Path, script: str, *args: Union[str, Path]) -> str:
+    proc = run_subprocess([interpreter, "-c", script, *args], capture_stderr=False)
+    return proc.stdout
```
cs01: This is a sequence of bytes that needs to be decoded isn't it?

src/pipx/util.py
```diff
@@ -98,11 +93,18 @@ def run_subprocess(
...
+    env = dict(os.environ)
 
-    # Null out PYTHONPATH because some platforms (macOS with Homebrew) add
+    # Remove PYTHONPATH because some platforms (macOS with Homebrew) add
     #   pipx directories to it, and can make it appear to venvs as though
     #   pipx dependencies are in the venv path (#233)
-    env = {k: v for k, v in os.environ.items() if k.upper() != "PYTHONPATH"}
+    # Remove __PYVENV_LAUNCHER__ because it is only meant to be communication
+    #   between main python binary and its subprocess and can cause the wrong
+    #   python binary to be used (#334)
```
cs01: nit: this is a little confusing. Maybe change to
```
    # Remove __PYVENV_LAUNCHER__ because it can cause the wrong
    #   python binary to be used (#334)
```

src/pipx/venv.py
```diff
@@ -242,7 +242,7 @@ def install_package_no_deps(self, package_or_url: str, pip_args: List[str]) -> s
     def get_venv_metadata_for_package(self, package: str) -> VenvMetadata:
         data = json.loads(
             get_script_output(
-                self.python_path, VENV_METADATA_INSPECTOR, package, str(self.bin_path)
+                self.python_path, VENV_METADATA_INSPECTOR, package, self.bin_path
```
cs01: I always cast to string because we've run into problems on windows when passing `Path` type arguments and assuming they will be auto-cast to strings. The CI tests passed so I guess this is okay.

src/pipx/util.py
itsayellow: if `subprocess.run()` is called with the `universal_newlines` option set to true (as it is in `run_subprocess()`) then stdout and stderr are already decoded text.  I think the PYTHONIOENCODING stuff is only needed to decode bytes, but with the `universal_newlines` this is not necessary.

https://docs.python.org/3.6/library/subprocess.html#subprocess.run
> If encoding or errors are specified, or universal_newlines is true, file objects for stdin, stdout and stderr are opened in text mode using the specified encoding and errors or the io.TextIOWrapper default. Otherwise, file objects are opened in binary mode.

src/pipx/util.py
itsayellow: See the above comment.  `proc.stdout` is text not bytes with `universal_newlines` set in `run_subprocess()`.

src/pipx/venv.py
itsayellow: Inside of `run_subprocess()`, line 112 manually and explicitly casts every item in the command list to a string.  This was another convenience of `run_subprocess()`.

src/pipx/util.py
itsayellow: Done

src/pipx/venv.py
itsayellow: I guess now it's line 111 of `util.py`:
```
cmd_str_list = [str(c) for c in cmd]
```

src/pipx/util.py
cs01: > or the io.TextIOWrapper default

If the encoding isn't specified, then the default is `locale.getpreferredencoding` (https://docs.python.org/3/library/locale.html#locale.getpreferredencoding)

There are various "fun" nuances with this:

> Return the encoding used for text data, according to user preferences. User preferences are expressed differently on different systems, and might not be available programmatically on some systems, so this function only returns a guess.
On some systems, it is necessary to invoke setlocale() to obtain the user preferences, so this function is not thread-safe. If invoking setlocale is not necessary or desired, do_setlocale should be set to False.
On Android or in the UTF-8 mode (-X utf8 option), always return 'UTF-8', the locale and the do_setlocale argument are ignored.
Changed in version 3.7: The function now always returns UTF-8 on Android or if the UTF-8 mode is enabled.

The `PYTHONIOENCODING` env variable is described like this:

> If this is set before running the interpreter, it overrides the encoding used for stdin/stdout/stderr, in the syntax encodingname:errorhandler. Both the encodingname and the :errorhandler parts are optional and have the same meaning as in str.encode().

I think using it defensively to prevent any edge cases from popping up makes sense.

src/pipx/util.py
cs01: Ah I missed that we are using `universal_newlines`. Thanks.

src/pipx/util.py
itsayellow: OK.  I wasn't worried about the performance hit--my thought was that getting the text decoded by a standard library instead of us doing it by hand might be simpler and more reliable.  And I was trying to disturb the environment variables less.

But I can add the PYTHONIOENCODING back.

cs01: We are letting it be decoded by the standard library since `universal_newlines` is true. But we are making extra sure the bytes are decoded to text the right format by specifying the format rather than leaving it up to the system's configuration.

cs01: We should also specify the encoding on the `subprocess.run` call to make sure it's decoded properly.

itsayellow: I wish the `subprocess` docs were clearer.  My assumption was that subprocess would use the encoding that is being used to make the bytes in the first place, but I guess one can't tell from the docs.

Yes, I was alreading setting the `encoding` to "utf-8" in `subprocess.run()` :+1:

cs01: > I wish the subprocess docs were clearer. My assumption was that subprocess would decode using the encoding that is being used to make the bytes in the first place, but I guess one can't tell from the docs.

Totally agree

cs01:
Thank you for putting this together

PR #340, opened 2020-01-20
Windows: bugfix, correctly inventorying apps

itsayellow:
Fixes #217 .

This properly finds and inventories the correct apps and app paths on Windows.
* It includes app names with a `.exe` at the end of the entry_point name if they exist in the venv bin_dir
* It excludes apps without the `.exe` at the end of the name if they don't exist in the venv bin_dir
* It looks for companion apps that end in `-script.py` that are needed for editable installs on Windows, and adds these to app_paths but not to the list of apps.  That way the `-script.py` files are copied to PIPX_BIN_DIR but do not show up during `pipx list`.

Tests were updated, especially to reflect correct list of apps.  Also, tests were enabled for previously skipped test packages ("cloudtoken", "awscli", "shell-functools") that now work on Windows.  "ansible" is still skipped because it officially is not installable on Windows.

A pre-existing test involving symlinks was changed to be excluded on Windows (test_existing_symlink_points_to_nothing).

PR #341, opened 2020-01-20
Also add *.exe.manifest on windows if present.

itsayellow:
This is an addendum to PR #340 .

To be safe, on Windows if a `.exe.manifest` file has been generated in the venv bin_dir, we will copy that over as well to PIPX_BIN_DIR.  (Along with `<file>.exe`, and `<file>-script.py`.

The `.exe.manifest` file ensures that 32-bit Windows will not mistake an `.exe` file in the same dir as an installer, for example if the main `.exe` file contains the words "install", or "setup", etc.

More Info:
https://github.com/pypa/setuptools/blob/682b6511ac67e021b542e74ce30e13fe52bc2da9/setuptools/command/easy_install.py#L2243
https://setuptools.readthedocs.io/en/latest/history.html#id627
https://bitbucket.org/tarek/distribute/issues/143
https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-vista/cc709628(v=ws.10)?redirectedfrom=MSDN

### DIFF
--- a/src/pipx/venv_metadata_inspector.py
+++ b/src/pipx/venv_metadata_inspector.py
@@ -102,9 +102,14 @@ def _dfs_package_apps(
 def _windows_extra_app_paths(app_paths: List[Path]) -> List[Path]:
     # In Windows, editable package have additional files starting with the
     #   same name that are required to be in the same dir to run the app
+    # Add "*-script.py", "*.exe.manifest" only to app_paths to make
+    #   execution work; do not add them to apps to ensure they are not listed
     app_paths_output = app_paths.copy()
     for app_path in app_paths:
         win_app_path = app_path.parent / (app_path.stem + "-script.py")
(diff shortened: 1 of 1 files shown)


cs01:
Are these be cleaned up on uninstall?

itsayellow:
> Are these be cleaned up on uninstall?

Yes.  I verified that.  

But also in the code: Thanks to pipx_metadata.json, and the fact that these Windows files are recorded in `app_paths` and `app_paths_of_dependencies`--when we uninstall we iterate through these lists and remove all the names from them we find inside of PIPX_BIN_DIR.

itsayellow:
For the record of this PR, @uranusjr said:
> I believe the manifest file can have a positive effect to anti virus things, no harm to copy them.

https://github.com/pipxproject/pipx/issues/217#issuecomment-576142377

Issue #345, opened 2020-01-26
[packaging] pipx-0.15.1.2/src/pipx/.main.py.swp: Failed to restore metadata


AlJohri:
Looks like there's some `swp` file in the `0.15.1.2` released tar file that is causing errors with installing the homebrew package. I'll take a look next week but flagging here in case anyone any ideas:

https://github.com/Homebrew/homebrew-core/pull/48955
https://github.com/Homebrew/homebrew-core/pull/49062

itsayellow:
Thanks @AlJohri .  btw, we're at 0.15.1.3 now

itsayellow:
So `.main.py.swp` is a vim swapfile that is created whenever vim is editing a file (in this case when vim would be editing `main.py`.

There's no `.main.py.swp` in the 0.15.1.2 tagged repo, but when I download the [tar.gz file from pypi.org](https://files.pythonhosted.org/packages/55/d7/0f29f5b369af46e8d33f49a47720f7d419da132706bb550560f32504b266/pipx-0.15.1.2.tar.gz) , it IS in the version submitted to pypi.  I'm guessing this is my fault, since I did the 0.15.1.2 and I use vim.

This problem does not exist in 0.15.1.3 (no .swp files found).

It seems we need to re-submit 0.15.1.2 to pypi with the proper files.  I'm not sure how to do that.  Does anybody know how to overwrite pypi versions?

itsayellow:
It seems you cannot fix an old pypi version.  You can delete it, but you cannot upload a new version to replace the old one.
https://stackoverflow.com/questions/44517306/how-can-i-re-upload-package-to-pypi
https://github.com/pypa/packaging-problems/issues/74

@AlJohri , that file (`.main.py.swp`) is a mistake and should be deleted.  If you can do (despite the difference with the pypi tar.gz) then that would be the solution.

Another question, since it is the only solution possible, is if we should delete the 0.15.1.2 version on pypi that has the bogus additional file.

uranusjr:
One way to sort of amend this with the least disturbance is to issue a post release (i.e. 0.15.1.2.post1) with the same code as 0.15.1.2, but with the swap file removed.

itsayellow:
So @uranusjr , would the idea be EXACTLY everything the same as the 0.15.1.2 release, just without the swap file?  For example, I would not change any metadata or version info, I would leave all of that just saying "0.15.1.2"?

itsayellow:
[...]

PR #346, opened 2020-01-28
Dev: Add a global-exclude for vim swap files.

itsayellow:
This is for pipx development.

Recently I made the mistake of including a `.*.swp` file (vim swap file) in a release sent to pypi (#345).  This PR is adding a global-exclude to MANIFEST.in so that won't happen again.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -9,3 +9,4 @@ prune scripts
 
 global-exclude *.py[cod]
 global-exclude __pycache__
+global-exclude .*.swp


uranusjr:
GitHub has a pattern for VIM swap files that might be useful, but this also LGTM in the immediate term.

https://github.com/github/gitignore/blob/master/Global/Vim.gitignore

cs01:
> GitHub has a pattern for VIM swap files that might be useful, but this also LGTM in the immediate term.

A gitinore file won't change what's in the build though, will it? Only the MANIFEST.in file defines that IIRC.

cs01:
This is good for the .swp files, but only for those. We should err on the side of including only things we need rather than only excluding things we don't want. If we still include everything it seems like we are setting ourselves up for a similar situation with a .emacs file or something.

itsayellow:
> GitHub has a pattern for VIM swap files that might be useful, but this also LGTM in the immediate term.
> 
> https://github.com/github/gitignore/blob/master/Global/Vim.gitignore

@uranusjr , the problem is not with the repo (there is no swp-file in the repo), the problem was in my working dir there was a swap file when I submitted to pypi.

> LGTM but long-term it will be better to change from "include by default" to "exclude by default".

@cs01 I totally agree.  This was just a quick fix to eliminate one problem.  Possibly automating the submission using what files are actually checked into the repo as opposed to what is in the working directory is another fix.

cs01:
>  Possibly automating the submission using what files are actually checked into the repo as opposed to what is in the working directory is another fix.

This is already done in the `lint` session with a call to `check-manifest`. It verifies everything tracked in source control is accounted for in MANIFEST.in. Of course in this case we are grafting the `src` directory, so `check-manifest` is happy, yet we are potentially including things we shouldn't be.

uranusjr:
@itsayellow Yeah I know, I was just suggesting maybe it’d help include a well-known and comprehensive pattern if we’re doing this. But erring on the side of being conservative (and improve as needed) is fine for me as well.

itsayellow:
Ah @uranusjr I see, sorry was slow on the uptake.  

One thing is that the MANIFEST.in documentation was pretty vague on what kind of globbing was used (for example, there are subtle differences between python's glob and pathlib Path globbing, such as how they handle files that start with a ".").  So yes, I tried to stay pretty conservative with the actual pattern.

PR #358, opened 2020-02-07
Add markdown max version for mkdocs-material.

itsayellow:
Can't wait for that advanced pip version resolution! :smile:

Our nox docs session is now failing because of a version conflict.  It installs the latest Markdown version (3.2), which then conflicts with mkdocs-material which has a requirement `Markdown<3.2`.

This PR has a workaround to force install of Markdown<3.2 until mkdocs-material will accept the latest Markdown version.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -27,7 +27,9 @@ else:
     nox.options.sessions = ["tests", "lint", "docs"]
 
 
-doc_dependencies = [".", "jinja2", "mkdocs", "mkdocs-material"]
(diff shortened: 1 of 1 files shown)


itsayellow:
https://github.com/squidfunk/mkdocs-material/issues/1450

cs01:
> Can't wait for that advanced pip version resolution! 

I think pip-tools already does this with `pip-compile`. I see some projects make a dev_requirements.txt to install in their test suite, which would have avoided this issue.

PR #361, opened 2020-02-07
[WIP] Remove pkg_resources

clbarnes:
See #339 .

Below py38, this depends on a backport of `importlib.metadata` (`importlib_metadata`).

Mypy refuses to look whichever arm is functional for the given python version, so both have to be ignored (see https://github.com/python/mypy/issues/1153 ). However, it also refuses to accept any imports from the module. Not sure how to get around that.

I think I captured the spirit of what venv_metadata_inspector was doing before, although it's a bit hairy. Plenty of comments in there - would love for someone with more experience in packaging internals to set me straight if I've missed anything!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,6 +16,7 @@ from pathlib import Path  # noqa E402
 from typing import List  # noqa E402
 import ast  # noqa E402
 import re  # noqa E402
+import runpy
 
 CURDIR = Path(__file__).parent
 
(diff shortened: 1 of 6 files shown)


clbarnes:
The <py38 tests are failing because importlib_metadata isn't being installed (is the addition to setup.py incorrect?). The py38 tests are failing because the VenvMetadata constructor arguments aren't getting populated properly.

layday:
The <py38 tests are failing because the metadata inspector is run in the shared library environment.  You need to add `importlib_metadata` (unconstrained) to and `packaging` to the shared library's dependencies in `shared_lib.py`.  You do need to add them to `setup.py`. See #344 where this was discussed extensively.

cs01:
In https://github.com/pipxproject/pipx/issues/339#issuecomment-583532860 I was interested in a PR with the simple change of defining and tuple and computing the version string based on the tuple. Along with that, the removal of the top-level dependency in setup.py on `packaging` could be done, since that is only used in one place for the version tuple.

cs01:
What is the status of this PR?

clbarnes:
Sorry I didn't get round to finishing this, I think it can be closed now https://github.com/pipxproject/pipx/issues/339#issuecomment-645089617

PR #363, opened 2020-02-08
ensure clean master branch on release

cs01:
Ensure git checkout is on master branch and is clean before publishing. Also publish docs after publishing release to PyPI.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -89,19 +89,35 @@ def build(session):
 
 def has_changes():
     status = (
-        subprocess.run("git status --porcelain", shell=True, stdout=subprocess.PIPE)
+        subprocess.run(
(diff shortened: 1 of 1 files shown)


noxfile.py
```diff
@@ -86,11 +88,25 @@ def build(session):
     session.run("python", "setup.py", "--quiet", "sdist", "bdist_wheel")
 
 
+def has_changes():
+    status = (
+        subprocess.run("git status --porcelain", shell=True, stdout=subprocess.PIPE)
+        .stdout.decode()
+        .strip()
+    )
+    return len(status) > 0
```
uranusjr: Maybe check the status code well? Otherwise it’d be reporting no changes when Git fails without output.

PR #366, opened 2020-02-13
update readme

cs01:
Update readme with

* more comparisons to existing tools
* instructions on how to make a package pipx-compatible
* instructions on installing from source in the readme
* try to more clearly differentiate install vs. run. I saw a case where someone was installing a package, and thought they needed to `pipx run` it after installing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,9 +53,11 @@ For more details, see the [installation instructions](https://pipxproject.github
 
 ## Overview: What is `pipx`?
 
-pipx is a tool to help you install and run end-user applications written in Python. It's roughly similar to macOS's `brew`, JavaScript's [npx](https://medium.com/@maybekatz/introducing-npx-an-npm-package-runner-55f7d4bd282b), and Linux's `apt`. It's closely related to `pip`. In fact, it uses pip, but is focused installing and managing packages that have entry points. For library installation you should use pip, not pipx. But for application installation, consider pipx.
+pipx is a tool to help you install and run end-user applications written in Python. It's roughly similar to macOS's `brew`, JavaScript's [npx](https://medium.com/@maybekatz/introducing-npx-an-npm-package-runner-55f7d4bd282b), and Linux's `apt`.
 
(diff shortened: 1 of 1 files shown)


itsayellow:
This is a lot clearer in general, nice.

I think we need to find a gentler introduction to the idea of "entry points".  I'd be willing to bet that's an unfamiliar term to many people who would like to use python applications via pipx.

For example, in __Overview: What is pipx?__, instead of "In fact, it uses pip, but is focused installing and managing packages that have entry points." maybe instead "In fact, it uses pip, but is focused on installing and managing packages that contain applications."  It could be some other text (not sure what's the clearest and most accurate), but at least in the overview we shouldn't require people to know what entry points are.  (Also, note there's a missing "on" in the original sentence.)

Later you have a very nice explanation of what an entry point is.

itsayellow:
Another typo:

If you're a poetry user, use these instruction.

should be

If you're a poetry user, use these instructions.

cs01:
Good feedback, thanks. I updated it again. It still feels like it could use some more smoothing out, but I think it's a little better. LMK what you think.

itsayellow:
As of now:

> It's closely related to `pip`. In fact, it uses `pip`, but is focused installing and managing packages that can be run from the command line directly as applications. `pip` is a general-purpose package installer for both libraries and apps with no environment isolation. For library installation you should use pip (or a similar tool), not `pipx`. But for application installation, consider `pipx` since it adds a layer of isolation on top of pip specifically for command-line tools.

How about adding "Python" to the first sentence, just to be clear we are not a generic package manager (like brew, apt suggestions in paragraph above):
Before
* "but is focused installing and managing packages"

After
* "but is focused on installing and managing Python packages"

How about explaining in a phrase what isolation means:
Before
* "consider `pipx` since it adds a layer of isolation on top of pip specifically for command-line tools."

After
* "consider `pipx` since it also adds isolation: `pipx` creates a separate virtual environment for each application and its associated packages."

Issue #208, opened 2019-08-12
Installation instructions don't work (pyenv problems)


anentropic:
```
 ~ $ python3 -m pip install --user pipx
Collecting pipx
  Downloading https://files.pythonhosted.org/packages/bd/16/4dd481c913d748928df79a78dfbf6ce3f42ec5382b630d78f3fd0ef0d8bd/pipx-0.14.0.0-py3-none-any.whl
Collecting userpath (from pipx)
  Downloading https://files.pythonhosted.org/packages/8c/71/72611cc9aed89374326645a4069787328db573c842e4e37b78c022b18146/userpath-1.2.0-py2.py3-none-any.whl
Collecting argcomplete<2.0,>=1.9.4 (from pipx)
  Downloading https://files.pythonhosted.org/packages/4d/82/f44c9661e479207348a979b1f6f063625d11dc4ca6256af053719bbb0124/argcomplete-1.10.0-py2.py3-none-any.whl
Collecting distro (from userpath->pipx)
  Downloading https://files.pythonhosted.org/packages/ea/35/82f79b92fa4d937146c660a6482cee4f3dfa1f97ff3d2a6f3ecba33e712e/distro-1.4.0-py2.py3-none-any.whl
Collecting click (from userpath->pipx)
  Using cached https://files.pythonhosted.org/packages/fa/37/45185cb5abbc30d7257104c434fe0b07e5a195a6847506c074527aa599ec/Click-7.0-py2.py3-none-any.whl
Installing collected packages: distro, click, userpath, argcomplete, pipx
  WARNING: The script distro is installed in '/Users/anentropic/Library/Python/3.7/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
  WARNING: The script userpath is installed in '/Users/anentropic/Library/Python/3.7/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
  WARNING: The script pipx is installed in '/Users/anentropic/Library/Python/3.7/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
Successfully installed argcomplete-1.10.0 click-7.0 distro-1.4.0 pipx-0.14.0.0 userpath-1.2.0
 ~ $ python3 -m pipx ensurepath
[...]

PR #370, opened 2020-02-21
Update installation instructions with macOS homebrew specifics

itsayellow:
Via #208, I see that `docs/installation.md` does not have the same macOS install and upgrade instructions that `README.md` does.  This PR is to fix that by integrating macOS homebrew instructions into `docs/installation.md`.


cs01:
LGTM

PR #375, opened 2020-02-24
removed modules io and os

CMeza99:
Since `pathlib.Path()` is already being used, removing `io.open()` for consistency and cleaner code.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,8 +10,6 @@ if sys.version_info < (3, 6, 0):
         "for installation instructions."
     )
 
-import io  # noqa E402
-import os  # noqa E402
 from pathlib import Path  # noqa E402
 from runpy import run_path  # noqa E402
 from typing import List  # noqa E402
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -38,7 +33,7 @@ def get_version():
     author="Chad Smith",
     author_email="grassfedcode@gmail.com",
     description="Install and Run Python Applications in Isolated Environments",
-    long_description=README,
+    long_description=Path(CURDIR, "README.me").read_text(),
```
cs01: suggest
```
(CURDIR / README.md).read_text(encoding="utf-8", )
```

uranusjr: Personally I prefer `CURDIR.joinpath("README.md")` in this kind of situations (but ultimately they are all implemented the same anyway, so it’s just a style issue).

CMeza99: `git commit --amend --no-edit` `!=` `git commit setup.py --amend --no-edit` :man_facepalming: 
@uranusjr that is what I have on my workstation. #git-fail

Issue #376, opened 2020-02-27
Remove emojis by default


traverseda:
They don't really have any benefit, support is platform, specific, and they look unprofessional. Also half the time I get broken characters because I don't have the right emoji fonts installed. I don't really need any cutesy icons in my tooling.

I recognize that this is a very subjective thing, and having emojis doesn't really hurt anything, I just don't like them.

I'm not expecting any action to be taken here, just wanted to mention that I think they look bad. Maybe people can vote on this issue to give a better idea of how many people also think they look bad/good?

If you like the emoji please give this comment a thumbs down.

----

The environment variable `USE_EMOJI` will allow you to opt-out of emoji usage.

Mattwmaster58:
I agree. In Windows they look terrible.

cs01:
>  In Windows they look terrible.

They should not be appearing in Windows

https://github.com/pipxproject/pipx/blob/8e6f8b798dd6c2ea0e59eb39f01ae1a4e7da4427/src/pipx/constants.py#L26

If they are, we need to determine a more reliable way to set `WINDOWS` or find places where the `emoji_support` isn't being used but should be.

Mattwmaster58:
Well, I'm assuming their emoji's or at the very least fancy Unicode characters:
![image](https://user-images.githubusercontent.com/26337069/81132061-5bc05e00-8f0a-11ea-8867-7ce465511724.png)

cs01:
> Well, I'm assuming their emoji's or at the very least fancy Unicode characters:

Ah those are probably the terminal codes to show/hide the cursor.
```
HIDE_CURSOR = "\033[?25l"
SHOW_CURSOR = "\033[?25h"
```

The goal is to do nothing fancy at all in windows. The windows animation uses a standard `.`.

I wonder if those should be taken out for windows too. Maybe we can't even clear lines.  Which terminal are you using?

Mattwmaster58:
@cs01 I'm using CMD - latest windows insider fast ring preview. Weirdly I thought a remember changing the code page to utf8 which in theory sold it anything fix things.

imba-tjd:
I also see `[?25l[?25h` in cmd and pwsh, but pipx works well in Windows Terminal and VS Code's terminal, which both support emoji either.

buhtz:
See also #1057

PR #377, opened 2020-02-27
Disable emoji unless explicitly enabled by the user

traverseda:


traverseda:
As per #376

cs01:
Is `USE_EMOJI` a common environment variable used more broadly?

traverseda:
Nope, I just tried to pick a name that expressed the sentiment well, and then realized that other projects might arrive at a similar name, and that we'd want a way to deal with it on a per-project basis.

traverseda:
Like I say, this is a personal preference thing. It might be more in-line with the projects goals if it defaulted there being emoji, a `NO_EMOJI` flag instead...

That also isn't a flag that's in use anywhere, and I worry that it's a bit of a stronger case. To me `NO_EMOJI` says "absolutely no emoji under any circumstances", where as `USE_EMOJI` doesn't have that same connotation. Maybe a `NO_EXTRA_EMOJI` flag if you want to default to there being emoji?

itsayellow:
Personally I like the emojis we use, so my vote would be for default-on emojis.

That being said, I have seen cases (like using a less-capable shell on Windows) where they are broken, and it would be nice if we disabled emojis automatically in more situations and/or had a user way of opting out.

itsayellow:
A conservative approach might be `PIPX_NO_EMOJI` since there is no overall standard and no indication that other apps would find such an env variable useful.

traverseda:
There's never going to be a formal standard for this, the best we're going to get is if tools start settling on something. It might as well start with this tool, so I'd *strongly* prefer a generic env var so if I bug other projects about removing emoji from their cli tool I can point at this and say "look, here's an informal standard". 

Well let's look at some other projects that use emoji, to get an idea of the kind of tone that using emoji in your project sets?

https://www.google.com/search?q="🔥"%20site:github.com

https://www.google.com/search?q="🚀"%20site:github.com

traverseda:
Use of emoji is very much a part of the javascript culture, and I have a *strong* dislike of javascript tooling, so 90% of the reason I don't like emoji in my python tooling is because it's subconsciously making me think this tool is doing things the "javascript way".

If I had to give an actual defense, beyond just "I don't like it", I'd say that using emoji in cli output is not idiomatic python. It's writing python tools like they're javascript tools.

But ultimately it's an aesthetic thing that doesn't matter too much.

traverseda:
How about defaulting to true?

`emoji_support = os.environ.get('PIPX_USE_EMOJI',os.environ.get('USE_EMOJI', True))`

So that `USE_EMOJI=""` will be our false value, but an unset `USE_EMOJI` variable is still defaults to the current behavior.

I'd have to add back in support for the default being whatever the platform default is, so in reality it would be 

```python
platform_emoji_support = not WINDOWS and sys.getdefaultencoding() == "utf-8"
# The "USE_EMOJI" environment variable allows you to override default emoji behavior.
# Optionally over-ride `USE_EMOJI` with the more specific `PIPX_USE_EMOJI` value.
emoji_support = os.environ.get('PIPX_USE_EMOJI',os.environ.get('USE_EMOJI', platform_emoji_support))
```

cs01:
I like the idea of emoji by default (`USE_EMJOI` is true), but can be turned off with the env var. I don't see a need to add in any other env vars. If you update the PR with that behavior I will merge it.

traverseda:
Done, `USE_EMOJI` defaults to platform default. Setting the env variable to an empty string disabled emoji regardless of platform default, and setting it to anything else force enabled emoji.

clbarnes:
As we're defaulting to using emoji where possible, would it make more sense for the environment variable, which will only ever be defined to prevent theuse of emoji, to be `DISABLE_EMOJI`?

traverseda:
I do like `DISABLE_EMOJI` more than `NO_EMOJI`, I'd also consider `LESS_EMOJI`. Obviously I have the bias that I'd like to eventually move projects towards an opt-in standard.

cs01:
`DISABLE_EMOJI` is fine with me too. Maybe make it plural, `DISABLE_EMOJIS`.

clbarnes:
Consensus on the internet seems to be that either is an acceptable plural, leaning slightly in favour of "emoji" in written contexts. I say go for fewer characters if it's equally valid.

clbarnes:
Consensus on the internet seems to be that either is an acceptable plural, leaning slightly in favour of "emoji" in written contexts. I say go for fewer characters if it's equally valid.

cs01:
> Personally I like the emojis we use, so my vote would be for default-on emojis.
> 
> That being said, I have seen cases (like using a less-capable shell on Windows) where they are broken, and it would be nice if we disabled emojis automatically in more situations and/or had a user way of opting out.

Let's choose a set of rules to turn on emoji's in only the most conservative of use cases. Something like: os is not windows, and `$LANG` ends with `.UTF-8`.

cs01:
This is basically ready to merge. Just need to run `black` on pipx/src/pipx/constants.py.

traverseda:
>This is basically ready to merge. Just need to run black on pipx/src/pipx/constants.py.

Sorry, I somehow missed this message or I would have done the formatting myself.

cs01:
I have a branch I'm trying to update this PR with but I am getting some permission errors. Want to try to re-push to this PR with the contents of the branch cs01/disable-emoji?

traverseda:
Yep

traverseda:
Well it is done for what that's worth, doesn't seem to be updating. Maybe open the PR back up and try again?

cs01:
Now travis isn't building. New PR here https://github.com/pipxproject/pipx/pull/398.

Issue #378, opened 2020-03-02
Pipx crashes during editable install


AliGhahraei:
**Describe the bug**
Pipx 0.15.1.3 crashes during editable install.

**How to reproduce**

```bash
$ pipx install --verbose -e ~/g/scripts/maintenance
pipx > (run_subprocess:112): running /usr/bin/python -m venv --without-pip /tmp/tmpqvyi9c_8
pipx > (run_subprocess:112): running /tmp/tmpqvyi9c_8/bin/python -c import sysconfig; print(sysconfig.get_path('purelib'))
pipx > (run_subprocess:112): running /home/ali/.local/pipx/shared/bin/python -c import sysconfig; print(sysconfig.get_path('purelib'))
pipx > (run_subprocess:112): running /tmp/tmpqvyi9c_8/bin/python --version
pipx > (run_subprocess:112): running /tmp/tmpqvyi9c_8/bin/python -m pip list --format=json
pipx > (run_subprocess:112): running /tmp/tmpqvyi9c_8/bin/python -m pip install --no-dependencies --editable /home/ali/g/scripts/maintenance
Obtaining file:///home/ali/g/scripts/maintenance
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
Installing collected packages: maintenance
  Running setup.py develop for maintenance
Successfully installed maintenance
pipx > (run_subprocess:112): running /tmp/tmpqvyi9c_8/bin/python -m pip list --format=json
pipx > (install_package_no_deps:231): Determined package name: 'maintenance'
pipx > (_package_name_from_spec:104): Package name determined in 7.0s
pipx > (run_subprocess:112): running /usr/bin/python -m venv --without-pip /home/ali/.local/pipx/venvs/maintenance
pipx > (run_subprocess:112): running /home/ali/.local/pipx/venvs/maintenance/bin/python -c import sysconfig; print(sysconfig.get_path('purelib'))
pipx > (run_subprocess:112): running /home/ali/.local/pipx/venvs/maintenance/bin/python --version
pipx > (run_subprocess:112): running /home/ali/.local/pipx/venvs/maintenance/bin/python -m pip install --editable /home/ali/g/scripts/maintenance
Obtaining file:///home/ali/g/scripts/maintenance
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
[...]

PR #380, opened 2020-03-06
Fix `venv_metadata_inspector.py` error handling/reporting

itsayellow:
This fixes a bug with bad empty json data returned during a caught exception in `venv_metadata_inspector`.  It also captures and communicates the traceback of an Exception that occurred in `venv_metadata_inspector.py` to the caller.  It is inspired by #378 and the difficulty of assessing what was going wrong inside of `venv_metadata_inspector.py`.
                                                                                
Changes                                                                         
* `venv_metadata_inspector.py` now returns proper empty data json (added required key `"apps_of_dependencies"`
* Pass string of traceback for a caught Exception in `venv_metadata_inspector.py` in returned json
* ~~Change invocation of `venv_metadata_inspector.py`: instead of passing file text to `python -c` use filename passed to `python`.  This prevents entire text of `venv_metadata_inspector.py` from showing up in pipx verbose output.~~
* `venv.get_venv_metadata_for_package()` logs a warning ("Internal error [...]") and logs info of `venv_metadata_inspector.py` traceback string.
* Removed `pipx.utils.get_script_output()` because it was only used in one place in `utils.py`.
* Added extra changelog for #373

### DIFF
--- a/src/pipx/util.py
+++ b/src/pipx/util.py
@@ -75,15 +75,11 @@ else:
         return bin_path, python_path
 
 
-def get_script_output(interpreter: Path, script: str, *args: Union[str, Path]) -> str:
-    proc = run_subprocess([interpreter, "-c", script, *args], capture_stderr=False)
-    return proc.stdout
-
-
 def get_site_packages(python: Path) -> Path:
-    output = get_script_output(
-        python, "import sysconfig; print(sysconfig.get_path('purelib'))"
-    )
+    output = run_subprocess(
+        [python, "-c", "import sysconfig; print(sysconfig.get_path('purelib'))"],
+        capture_stderr=False,
+    ).stdout
     path = Path(output.strip())
     path.mkdir(parents=True, exist_ok=True)
     return path
--- a/src/pipx/util.py
+++ b/src/pipx/util.py
@@ -89,7 +89,7 @@ def run_subprocess(
     cmd: Sequence[Union[str, Path]],
     capture_stdout: bool = True,
     capture_stderr: bool = True,
-    log_cmd_str: Optional[str] = None
+    log_cmd_str: Optional[str] = None,
 ) -> subprocess.CompletedProcess:
     """Run arbitrary command as subprocess, capturing stderr and stout"""
     env = dict(os.environ)
--- a/src/pipx/util.py
+++ b/src/pipx/util.py
@@ -5,7 +5,7 @@ import shutil
 import subprocess
 import sys
 from pathlib import Path
-from typing import List, Sequence, Tuple, Union
+from typing import List, Sequence, Tuple, Union, Optional
 
 from pipx.constants import WINDOWS
 
@@ -89,6 +89,7 @@ def run_subprocess(
     cmd: Sequence[Union[str, Path]],
     capture_stdout: bool = True,
     capture_stderr: bool = True,
+    log_cmd_str: Optional[str] = None
 ) -> subprocess.CompletedProcess:
     """Run arbitrary command as subprocess, capturing stderr and stout"""
     env = dict(os.environ)
@@ -106,8 +107,9 @@ def run_subprocess(
     # Make sure that Python writes output in UTF-8
     env["PYTHONIOENCODING"] = "utf-8"
 
-    cmd_str = " ".join(str(c) for c in cmd)
-    logging.info(f"running {cmd_str}")
+    if log_cmd_str is None:
+        log_cmd_str = " ".join(str(c) for c in cmd)
+    logging.info(f"running {log_cmd_str}")
     # windows cannot take Path objects, only strings
     cmd_str_list = [str(c) for c in cmd]
     return subprocess.run(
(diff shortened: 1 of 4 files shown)


src/pipx/util.py
```diff
@@ -75,15 +75,11 @@ def get_venv_paths(root: Path) -> Tuple[Path, Path]:
...
-
-
 def get_site_packages(python: Path) -> Path:
-    output = get_script_output(
-        python, "import sysconfig; print(sysconfig.get_path('purelib'))"
-    )
+    output = run_subprocess(
+        [python, "-c", "import sysconfig; print(sysconfig.get_path('purelib'))"],
+        capture_stderr=False,
+    ).stdout
```
uranusjr: Is there a behaviour change in this part? I assumed there is because there’s significant code diff, but couldn’t figure out what changed reading it multiple times.

src/pipx/venv.py
```diff
@@ -241,10 +232,24 @@ def install_package_no_deps(self, package_or_url: str, pip_args: List[str]) -> s
...
-            )
+            run_subprocess(
+                [
+                    self.python_path,
+                    VENV_METADATA_INSPECTOR_FILE,
+                    package,
+                    self.bin_path,
+                ],
+                capture_stderr=False,
+            ).stdout
```
uranusjr: Would this work if pipx is run in a non-flat context? e.g. zipapp or py2exe.

src/pipx/util.py
itsayellow: In the original, we submitted the entire contents of `venv_metadata_inspector.py` as one giant string as the argument after `python -c`.

The version I changed it to uses the full-path to `venv_metadata_inspector.py` file to run it like a normal script, e.g. `python venv_metadata_inspector.py`.

src/pipx/venv.py
itsayellow: The code here is exactly what used to be there inside of `get_script_output()`, the only exception being no longer using the `-c` switch after the python interpreter (i.e. submitting the name of the script to the python interpreter instead of the entire string contents of the script.)  So I assume it should work the same way.

Is there an example of a zipapp or py2exe that I can test to double-check?

src/pipx/util.py
itsayellow: Sorry, I made a previous comment referring to the wrong code.

No, this particular code should be exactly the same.  I just took the code that used to be wrapped inside of get_script_output.  It should be functionally the same, just not wrapped in another function.

itsayellow: This was the only remaining place using `get_script_output` so I thought it might be simpler just to show the code instead of using one more level of function.

src/pipx/venv.py
uranusjr: The problem here is the usage of `__file__`, which is [not available for py2exe](https://stackoverflow.com/questions/24140593) (and other bundling utilities, since they generate do no maintain the source code tree).

A simple example with zipapp:

```python
# foo/__main__.py
import subprocess
import sys

from mod import __file__ as mod_file

subprocess.check_call([sys.executable, mod_file])
```

```python
# foo/mod.py
if __name__ == "__main__":
    print("Executing mod")
```

```console
$ python3 foo/__main__.py   # This works.
Executing mod
```

```console
$ python3 -m zipapp foo
$ python3 foo.pyz  # But this does not.
python3: can't find '__main__' module in 'foo.pyz/mod.py'
  File "lib/runpy.py", line 192, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "lib/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "foo.pyz/__main__.py", line 6, in <module>
  File "lib/subprocess.py", line 364, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['python3', 'foo.pyz/mod.py']' returned non-zero exit status 1.
```

src/pipx/util.py
uranusjr: Cool, I just wanted to be sure I don’t miss any changes :+1:

src/pipx/venv.py
itsayellow: So to make sure I understand, this would be a problem if we bundle `pipx` itself, not so much a problem for the client apps?

uranusjr: Yes, that is correct.

itsayellow:
Ok, I reverted back to the original version using the text inside of `venv_metadata_inspector.py` using `pkgutil.get_data()`.  I also made an optional `log_cmd_str` argument for run_subprocess so in verbose mode pipx will now print and log `<contents of venv_metadata_inspector.py>` in place of the entire text of `venv_metadata_inspector.py` when it prints/logs the command.

For example, now:
```
pipx > (run_subprocess:112): running /Users/myuser/.local/pipx/venvs/pylint/bin/python -c <contents of venv_metadata_inspector.py> pylint /Users/myuser/.local/pipx/venvs/pylint/bin
```

src/pipx/venv.py
itsayellow: OK, I reverted back to using the text of `pipx_metadata_inspector.py` as before.

cs01:
Looks good to me!

PR #382, opened 2020-03-27
Separate commands into separate files (#255)

jackno:
Hopefully this satisfies #255 

All commands have been moved to separate files. Please note the following:
- `uninstall` and `uninstall_all` were moved into the same file to follow the convention in a previous commit for `upgrade` command
- `_package_name_from_spec()` and `_run_post_install_actions()` were moved to the `common.py` file
- `ensurepath` was renamed to `ensure_path` because all the other commands follow snake case, no documentation has been changed about this yet


src/pipx/commands/common.py
```diff
@@ -172,3 +178,104 @@ def _get_list_output(
             f"    - {red(name)} (symlink missing or pointing to unexpected location)"
         )
     return "\n".join(output)
+
+
+def _package_name_from_spec(
```
itsayellow: This should probably no longer begin with an underscore since we are referencing it from outside of this module.

src/pipx/commands/common.py
```diff
@@ -172,3 +178,104 @@ def _get_list_output(
...
+        venv.create_venv(venv_args=[], pip_args=[])
+        package_name = venv.install_package_no_deps(
+            package_or_url=package_spec, pip_args=pip_args
+        )
+
+    logging.info(f"Package name determined in {time.time()-start_time:.1f}s")
+    return package_name
+
+
+def _run_post_install_actions(
```
itsayellow: This should probably no longer begin with an underscore since we are referencing it from outside of this module.

src/pipx/commands/common.py
```diff
@@ -172,3 +178,104 @@ def _get_list_output(
...
+    if include_dependencies:
+        for _, app_paths in package_metadata.app_paths_of_dependencies.items():
+            expose_apps_globally(local_bin_dir, app_paths, package, force=force)
+
+    print(get_package_summary(venv_dir, package=package, new_install=True))
+    _warn_if_not_on_path(local_bin_dir)
+    print(f"done! {stars}", file=sys.stderr)
+
+
+def _warn_if_not_on_path(local_bin_dir: Path):
```
itsayellow: This should probably no longer begin with an underscore since we are referencing it from outside of this module.

jackno: 6cda48b

src/pipx/commands/common.py
jackno: 6cda48b

src/pipx/commands/common.py
jackno: 6cda48b

cs01:
I did a quick review and it looks good to me. @itsayellow did you have any other concerns?

itsayellow:
It looks good to me. 👍

docs/changelog.md
```diff
@@ -1,3 +1,7 @@
+0.15.2.0
```
itsayellow, line 1: Actually the entry for 0.15.1.4 is unreleased, these changes and other changes in the changelog for 0.15.1.4 should be combined into one section

itsayellow:
@cs01 I feel like we need a new placeholder for "the next release" in the changelog instead of a concrete version number, maybe something like "dev".  Maybe not important for this PR but something to think about for the future.

docs/changelog.md
jackno: Done in b677c5d and conflict merged in fa5397b

src/pipx/commands/__init__.py
```diff
@@ -23,5 +20,5 @@
     "reinstall_all",
     "list_packages",
     "run_pip",
-    "ensurepath",
+    "ensure_path",
```
jackno, line 23: Before merging, notice this command has been renamed from `ensurepath` to `ensure_path` (see first comment). A decision has to be made:

- Keep `ensurepath` or update docs to reflect `ensure_path`.

Let me know.

jackno:
Before merging, notice `ensurepath` has been changed to `ensure_path` (see first comment). A decision has to be made:

- [ ] Keep `ensurepath` or update docs to reflect `ensure_path`.

Let me know.

itsayellow:
It looked to me like only the function name has been changed, is that correct?

As long as the user-facing subcommand has not changed I don't have a problem with the internal function name changing.

> Before merging, notice `ensurepath` has been changed to `ensure_path` (see first comment). A decision has to be made:
> 
>     * [ ]  Keep `ensurepath` or update docs to reflect `ensure_path`.
> 
> 
> Let me know.

itsayellow:
Ah you bring up an important point that needs to be fixed.

You need to change `main.py` to reference the new broken-out commands.

I would prefer we keep `ensurepath` as the CLI subcommand, but it's not a problem to change the function name.

itsayellow:
Basically the 'ensurepath' subcommand can function like the 'runpip' subcommand does currently.  The function that is called from `main.py` will be `commands.ensure_path` even though the parsed subcommand is still 'ensurepath'.

jackno:
> You need to change `main.py` to reference the new broken-out commands.

Good catch.

Now the codebase command is `ensure_path` and CLI command remains `ensurepath`.

docs/changelog.md
```diff
@@ -1,7 +1,10 @@
 0.15.1.4
+
 - Improved error reporting during venv metadata inspection.
 - [bugfix] Fixed incompatibility with pypy as venv interpreter (#372).
 - [bugfix] Replaced implicit dependency on setuptools with an explicit dependency on packaging (#339).
+- [refactor] Moved all commands to separate files within the commands module (#255).
+- [refactor] Command `ensurepath` renamed to `ensure_path`.
```
itsayellow, line 7: Remove this so we don't confuse users (the subcommand is remaining 'ensurepath'.

itsayellow: (Just remove the last bullet item referring to ensurepath.)

jackno: Removed in c663d90

itsayellow:
Thanks @jackno !

itsayellow:
@cs01 if you're done looking at it I can merge, I'll wait a day or so.

cs01:
I agree the changelog version should always be `dev` when adding changes on top of the last release. For now this is fine, but in the future we can to keep that convention. 

I also agree that the CLI command for `ensurepath` should remain `ensurepath` (not `ensure_path`), but I do not have a preference for the way the function is named in the codebase, so this is all fine with me. Thank you for the PR, @jackno.

cs01:
> > You need to change `main.py` to reference the new broken-out commands.
> 
> Good catch.
> 
> Now the codebase command is `ensure_path` and CLI subcommand remains `ensurepath`.

Were there any failures on CI when the wrong module name was used? I would have expected something to fail, maybe mypy or flake8.

jackno:
I would've expected mypy to fail as well. There were no failures in CI before the module name was fixed in `main.py`. [Here's the line it should've caught](https://github.com/pipxproject/pipx/blob/f1062f69962ea1f70dba53f18130f69f63ab8d43/src/pipx/main.py#L202).

itsayellow:
Hmmm you're right.  I just ran an experiment locally and neither flake8 nor mypy complained if I put a non-existent command.<function>() in.

> Were there any failures on CI when the wrong module name was used? I would have expected something to fail, maybe mypy or flake8.

jackno:
I don't think anything is linted in the whole [`run_pipx_command()`](https://github.com/pipxproject/pipx/blob/f1062f69962ea1f70dba53f18130f69f63ab8d43/src/pipx/main.py#L106) because of [`# noqa: C901`](https://flake8.pycqa.org/en/2.5.5/extensions.html?highlight=noqa%20c901). Even if it is removed, `mypy` likely wouldn't lint it because `run_pipx_command()` [is not annotated](https://mypy.readthedocs.io/en/stable/common_issues.html#no-errors-reported-for-obviously-wrong-code).

`flake8` [doesn't check for things like this](https://stackoverflow.com/a/37101171).

itsayellow:
Bingo.  Thanks @jackno!
If I replace:
```
def run_pipx_command(args):  # noqa: C901
```
with:
```
def run_pipx_command(args: argparse.Namespace):  # noqa: C901
```

Then it will correctly identify an incorrect `commands.<function>()` call.

@jackno, can you replace that function definition for `run_pipx_command`?  Then I think we can merge this.

In the future I think `main.py` needs more checking, I'm seeing more than one function that could be annotated.

jackno:
Added. Thanks for the code

itsayellow:
@cs01 I'm ready to merge this unless you have any objections.

In the future we should annotate `main.py` more.

cs01:
> @cs01 I'm ready to merge this unless you have any objections.

Nope, this is great. Thank you both!

> In the future we should annotate main.py more.

Totally agree. I will put together a PR today.

itsayellow:
Fixes #255

PR #392, opened 2020-04-14
list: Add --clean for portable output

SanketDG:

### DIFF
--- a/src/pipx/commands/list_packages.py
+++ b/src/pipx/commands/list_packages.py
@@ -18,14 +18,15 @@ from pipx.emojies import sleep
 from pipx.venv import VenvContainer
 
 
-def list_packages(venv_container: VenvContainer):
+def list_packages(venv_container: VenvContainer, clean: bool):
     dirs = list(sorted(venv_container.iter_venv_dirs()))
     if not dirs:
         print(f"nothing has been installed with pipx {sleep}")
         return
 
-    print(f"venvs are in {bold(str(venv_container))}")
-    print(f"apps are exposed on your $PATH at {bold(str(constants.LOCAL_BIN_DIR))}")
+    if not clean:
+        print(f"venvs are in {bold(str(venv_container))}")
+        print(f"apps are exposed on your $PATH at {bold(str(constants.LOCAL_BIN_DIR))}")
 
     venv_container.verify_shared_libs()
 
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -176,7 +176,7 @@ def run_pipx_command(args: argparse.Namespace):  # noqa: C901
             venv_dir, package, pip_args, verbose, upgrading_all=False, force=args.force
         )
     elif args.command == "list":
-        return commands.list_packages(venv_container)
+        return commands.list_packages(venv_container, args.clean)
     elif args.command == "uninstall":
         return commands.uninstall(venv_dir, package, constants.LOCAL_BIN_DIR, verbose)
     elif args.command == "uninstall-all":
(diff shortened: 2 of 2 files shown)


cs01:
Thanks for the PR. I am curious, is there a technical reason this needs to be added or is this more of a preference thing? I think I would lean toward the flag being named `--portable`.

PratikBhusal:
Rather than clobber `list` with more options, I'd suggest making a new command `export` (name is a work in progress) that would work similar to  `pix runpip MAIN_APP freeze`.  Futhermore, it should be easy to export the output into either standard `pip freeze` output or what could be stored inside a `Pipfile` for example. 

If we go into exporting in `Pipfile` format, I'd suggest only exporting the main app name and the name of all the injected apps. This might also be a good idea to do with a psedudo `pip-freeze` aproach where we only show the bear minimum to recreate the virtual enviroment.

uranusjr:
For some context, there’s actually talk that pip may deprecate `freeze` as a standalone command, since it is too similar with `list`, and it’s difficult to tell their differences from the command names.

It does make logical sense to combine this export functionality with `list`. `--clean` is likely not the best way to do it, however, since it does not explain the intention well—why is a clean output related to machine friendliness and portability? It may be more desirable to have something akin to `pip list --format`, e.g.

* `pipx list` with no flags: Human readable, table-ish format
* `--format=export`: Machine-friendly output
* `--format=json`: Human readable, JSON format
* `--format=export,json`: Machine-friendly, JSON format

There are other ways to achieve similar results, but IMO it’s important to avoid using *flags*, but value switches instead.

PratikBhusal:
@uranusjr do you have a source about the deprecation? That feeling like an interesting read and could be a nice guide for something like this.

uranusjr:
@PratikBhusal I don’t have a canonical source, sorry. It’s mainly informal conversations scattered around in the issue tracker. pypa/pip#7157 is one of them, but there are others. IIRC there’s one proposing a new `pip inspect` command to replace both `list` and `freeze` as well, but I can’t seem to find it now. Or maybe I remembered wrong.

bcatubig:
I'll chime in and add that I like to do a list on an individual venv. Having a `--json` flag as well for parsing would also be a nice add.

```
pipx list --include-injected --json ansible
```

mjpieters:
I was looking for similar commands. More generally, being able to introspect pipx with other tools, being able to list pipx metadata without having to parse the human-readable output of the `pipx` command-line tools or having to rely on the `pipx_metadata.json` files would be a huge boon.

itsayellow:
We, the pipx team, are working on a freezing/unfreezing solution.  The interest has definitely been noted!

cs01:
Since there are a few use cases for various output formats, I like the idea of a `--format` flag with choices like @uranusjr mentioned. If this `clean` format could be changed to `--format=export`, I will merge this PR as it will also set us up for `--format=json`.

SanketDG:
> Since there are a few use cases for various output formats, I like the idea of a `--format` flag with choices like @uranusjr mentioned. If this `clean` format could be changed to `--format=export`, I will merge this PR as it will also set us up for `--format=json`.

Sure!

I should be able to get back to this soon.

itsayellow:
If this is meant for some sort of future import command, we should also list the versions of the injected packages.  I wonder if that should be the default behavior anyway for `--include-injected` so it matches the how the main package is displayed.

itsayellow:
Also if this is meant to be able to import in the future, shouldn't we just dump the entirety of each venv's `PipxMetadata` into the json?  Like a list of all the `PipxMetadata` instances.

itsayellow:
Related: #109

gaborbernat (changes requested):
Comments above needs addressed and merge conflicts resolved.

gaborbernat:
Closing due to inactivity, @SanketDG if you want to pick this up again we can reopen.

PR #388, opened 2020-04-10
Handle debundled pip._vendor, fixes #386

ivanov:
Some distrbutions (Debian, <s>Redhat</s>) break up pip's _vendor directory
into its constituent wheels, in which case we want to ensure that
the pth file we write in each virtual environment points to these
shared environment wheels.

*Edit*: if I understood the Redhat patches, they do something different

This will fix #386

Issue #387, opened 2020-04-09
Allow package to specify Python version


BrandonLWhite:
Background: I'm supporting a CLI tool built on Python3.6, that is used by non-developers.  As such, pipx is very attractive for making this hassle free for the users of the tool.  At this time, pipx is using Python3.8 when installed with `brew install pipx`.  This leads to an error of `ERROR: No matching distribution found for dataclasses==0.7` when subsequently using pipx install on my CLI application.

The particular error I'm facing is probably irrelevant.  I can imagine that there are other snags that could come with a surprise Python3.8 runtime.

I see that `pipx install` allows for a `--python` version switch.  That seems like it will get me out of the woods for now.  However, it occurs to me that requiring my end users to add this switch (and then down the road, probably updating it to 3.7, 3.8, etc) is not quite in line with the mission of pipx.

This request is to propose a way for a CLI application to expose its required Python version and for pipx to install and run it CLI with that version.  That would make it easier for my end-users now and in the future when we are able and want to use a later Python version.

uranusjr:
The challenge of this feature, from what I can tell, would be to match Python versions (which is what packages can specify with `requires-python`) cleanly to Python executable paths (which is what pipx needs to create virtual environments). pipx will also need to detect all installed Python versions, so for (say) `>=3.6` it can choose 3.8 over 3.7.

There are existing means to do this (e.g. `virtualenv` has an implementation to map version to executable path; the Python Launcher on Windows can collect versions from registry), but everything is quite ad-hoc in this area. I would say this is a very sensible feature to have, but the Python ecosystem does not have enough means yet to get it done.

johnklehm:
Would it be reasonable to build this out with pyenv/asdf in mind?  That would allow a large number of python users to make use of this feature.

uranusjr:
That goes back to the ad-hoc nature of Python version detection; both pyenv and asdf introduces even more tool-specific behaviour that needs to be special-cased, and that approach won’t scale well. IMO the design should be the other way around—someone should propose a central registration format (like PEP 514 but for POSIX systems), make pyenv etc. register to it, and pipx can then just read that registry to know where to find what versions. This needs to involve much more than just pipx; I would suggest raising a topic on https://discuss.python.org, I think some people there would be very interested in this.

johnklehm:
I agree that's the right move long term, it's likely a long process though.

uranusjr:
[...]

Issue #393, opened 2020-04-24
pipx run fails with json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)


mgedmin:
**Describe the bug**

```
$ pipx run pipx
⚠️  pipx is already on your PATH and installed at /home/mg/bin/pipx. Downloading and running anyway.
Traceback (most recent call last):
  File "/home/mg/.local/bin/pipx", line 8, in <module>
    sys.exit(cli())
  File "/home/mg/.local/pipx/venvs/pipx/lib/python3.8/site-packages/pipx/main.py", line 590, in cli
    return run_pipx_command(parsed_pipx_args)
  File "/home/mg/.local/pipx/venvs/pipx/lib/python3.8/site-packages/pipx/main.py", line 150, in run_pipx_command
    return commands.run(
  File "/home/mg/.local/pipx/venvs/pipx/lib/python3.8/site-packages/pipx/commands/run.py", line 89, in run
    retval = _download_and_run(
  File "/home/mg/.local/pipx/venvs/pipx/lib/python3.8/site-packages/pipx/commands/run.py", line 122, in _download_and_run
    venv.install_package(
  File "/home/mg/.local/pipx/venvs/pipx/lib/python3.8/site-packages/pipx/venv.py", line 179, in install_package
    package = self.install_package_no_deps(package_or_url, pip_args)
  File "/home/mg/.local/pipx/venvs/pipx/lib/python3.8/site-packages/pipx/venv.py", line 218, in install_package_no_deps
    old_package_set = self.list_installed_packages()
  File "/home/mg/.local/pipx/venvs/pipx/lib/python3.8/site-packages/pipx/venv.py", line 300, in list_installed_packages
    pip_list = json.loads(cmd_run.stdout.strip())
  File "/usr/lib/python3.8/json/__init__.py", line 357, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.8/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.8/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

**How to reproduce**

`pipx run pipx`

> If possible, include output of `pipx --verbose ...` 

```
$ pipx --verbose run pipx
usage: pipx [-h] [--version]
[...]

PR #394, opened 2020-04-24
Avoid Debian-patched pip in the shared venv

### LINKED ISSUE #386: ModuleNotFoundError: No module named 'pip._vendor.six' when running "pipx install"
**Describe the bug**

I have installed a new Debian (unstable) system. On one of my installations pipx runs without problems. The newly installed system behaves differently and I do not know where to look.

**How to reproduce**

```
$ pipx install --verbose youtube-dl
pipx > (_package_name_from_spec:93): Determined package name: youtube-dl
pipx > (_package_name_from_spec:94): Package name determined in 0.0s
pipx > (run_subprocess:112): running /usr/bin/python3 -m venv --without-pip /home/kai/.local/pipx/venvs/youtube-dl
pipx > (run_subprocess:112): running /home/kai/.local/pipx/venvs/youtube-dl/bin/python -c import sysconfig; print(sysconfig.get_path('purelib'))
pipx > (run_subprocess:112): running /home/kai/.local/pipx/shared/bin/python -c import sysconfig; print(sysconfig.get_path('purelib'))
pipx > (run_subprocess:112): running /home/kai/.local/pipx/venvs/youtube-dl/bin/python --version
pipx > (run_subprocess:112): running /home/kai/.local/pipx/venvs/youtube-dl/bin/python -m pip install youtube-dl
  File "/usr/lib/python3.8/runpy.py", line 193, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/kai/.local/pipx/shared/lib/python3.8/site-packages/pip/__main__.py", line 16, in <module>
    from pip._internal.cli.main import main as _main  # isort:skip # noqa
  File "/home/kai/.local/pipx/shared/lib/python3.8/site-packages/pip/_internal/cli/main.py", line 10, in <module>
    from pip._internal.cli.autocompletion import autocomplete
  File "/home/kai/.local/pipx/shared/lib/python3.8/site-packages/pip/_internal/cli/autocompletion.py", line 9, in <module>
    from pip._internal.cli.main_parser import create_main_parser
  File "/home/kai/.local/pipx/shared/lib/python3.8/site-packages/pip/_internal/cli/main_parser.py", line 7, in <module>
    from pip._internal.cli import cmdoptions
  File "/home/kai/.local/pipx/shared/lib/python3.8/site-packages/pip/_internal/cli/cmdoptions.py", line 24, in <module>
    from pip._internal.exceptions import CommandError
  File "/home/kai/.local/pipx/shared/lib/python3.8/site-packages/pip/_internal/exceptions.py", line 10, in <module>
    from pip._vendor.six import iteritems
ModuleNotFoundError: No module named 'pip._vendor.six'
pipx > (install_package:189): '/home/kai/.local/pipx/venvs/youtube-dl/bin/python -m pip install youtube-dl' failed

pipx > (rmdir:18): removing directory /home/kai/.local/pipx/venvs/youtube-dl
Error installing youtube-dl.
```

If I run python3 from the command line I can import the module:

```
$ python3 -c "import pip._vendor.six"
$
```

I have no idea what could be different. I already checked the list of python3-* packages on both systems but that seems to be the same. Any idea what to look for or how to debug the issue?

**Expected behavior**

The package should be installed.

mgedmin:
Should fix #386.

This fix should be more lightweight than #387, and it is simpler than #388.  It tries to find an acceptable middle ground between explicitly supporting Debian's changes to their pip and causing pain to users in order to make a statement.

This patch needs testing.  I haven't come up with a test plan yet (I don't want to mess up my currently-working actual pipx setup), but I'll try to do something with docker.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -26,6 +26,7 @@ dev
 - [bugfix] Hide cursor while pipx runs
 - [feature] Add environment variable `USE_EMOJI` to allow enabling/disabling emojies (#376)
 - [refactor] Moved all commands to separate files within the commands module (#255).
+- [bugfix] Ignore system shared libraries when installing shared libraries pip, wheel, and setuptools. This also fixes an incompatibility with Debian/Ubuntu's version of pip (#386).
 
 0.15.1.3
 
--- a/src/pipx/shared_libs.py
+++ b/src/pipx/shared_libs.py
@@ -32,7 +32,12 @@ class _SharedLibs:
         if not self.is_valid:
             with animate("creating shared libraries", not verbose):
                 run([DEFAULT_PYTHON, "-m", "venv", "--clear", self.root])
-            self.upgrade(pip_args, verbose)
+            # Note: -I is needed here to replace the patched Debian pip with
+            # the upstream one, even when the Debian pip version matches the
+            # current upstream pip version.  Otherwise there's a flood of
+            # bug reports about things not working, like
+            # https://github.com/pipxproject/pipx/issues/386
+            self.upgrade(["-I"] + pip_args, verbose)
 
     @property
     def is_valid(self):
--- a/src/pipx/shared_libs.py
+++ b/src/pipx/shared_libs.py
@@ -32,12 +32,9 @@ class _SharedLibs:
         if not self.is_valid:
             with animate("creating shared libraries", not verbose):
                 run([DEFAULT_PYTHON, "-m", "venv", "--clear", self.root])
-            # Note: -I is needed here to replace the patched Debian pip with
-            # the upstream one, even when the Debian pip version matches the
-            # current upstream pip version.  Otherwise there's a flood of
-            # bug reports about things not working, like
-            # https://github.com/pipxproject/pipx/issues/386
-            self.upgrade(["-I"] + pip_args, verbose)
+            # ignore installed packages to ensure no unexpected patches from the OS vendor
+            # are used
+            self.upgrade(["--ignore-installed"] + pip_args, verbose)
 
     @property
     def is_valid(self):


mgedmin:
Test plan:

- docker run --rm -ti ubuntu:focal bash
- apt update -qq && apt install -y python3-{pip,venv} git
- pip3 install --user git+https://github.com/mgedmin/pipx@debian-what-you-doin
- ~/.local/bin/pipx run pipx

This confirms that my fix works!

Replacing step 3 with `pip3 install --user pipx` confirms that the test plan can recreate #393, which is a manifestation of #386.

Now to fix unit tests.

src/pipx/shared_libs.py
```diff
@@ -32,7 +32,12 @@ def create(self, pip_args: List[str], verbose: bool = False):
         if not self.is_valid:
             with animate("creating shared libraries", not verbose):
                 run([DEFAULT_PYTHON, "-m", "venv", "--clear", self.root])
-            self.upgrade(pip_args, verbose)
+            # Note: -I is needed here to replace the patched Debian pip with
+            # the upstream one, even when the Debian pip version matches the
+            # current upstream pip version.  Otherwise there's a flood of
+            # bug reports about things not working, like
+            # https://github.com/pipxproject/pipx/issues/386
+            self.upgrade(["-I"] + pip_args, verbose)
```
gaborbernat, line 40: Will this be a NO OP when running outside of Debian? Would not want to make all platforms slower to patch a Debian bug.

docs/changelog.md
```diff
@@ -5,6 +5,7 @@ dev
 - [bugfix] Replaced implicit dependency on setuptools with an explicit dependency on packaging (#339).
 - [refactor] Moved all commands to separate files within the commands module (#255).
 - [bugfix] Continue reinstalling packages after failure
+- [bugfix] Fixed incompatibility with Debian/Ubuntu changes made to their version of pip (#386).
```
gaborbernat, line 8: Considering the bug is within Debian packaging I consider this more an enhancement than bug fix; something along the line (Support Debian 20s pip patching).

src/pipx/shared_libs.py
mgedmin: I don't know.  Some measurements would be nice to have.  I don't have any non-Debian-based systems to test against, unless maybe I could come up with something using Docker and Alpine?

docs/changelog.md
mgedmin: I've checked the "Allow edits from maintainers" box.

I'm a bit confused about the phrase "Support Debian 20s pip".  What's "20s"?  pip version?

I haven't mentioned specific Debian/Ubuntu/pip versions because I haven't done the research to figure out where the incompatibility got introduced.  I know things worked in Ubuntu 19.10 and broke in 20.04.  I don't run Debian myself and have trouble remembering its version numbers and codenames.

src/pipx/shared_libs.py
mgedmin: lol, -I makes this step _faster_

```
$ docker run -ti --rm python:3.8-alpine sh
... some experiments to get the commands right ...
... which incidentally populate ~/.pip/cache so the measurements are independent of network speed ...
/ # rm -rf env && time python3 -m venv env --clear && time env/bin/python -m pip --disable-pip-version-check install --upgrade pip setuptools wheel
real	0m 1.75s
user	0m 1.59s
sys	0m 0.15s
Collecting pip
  Using cached https://files.pythonhosted.org/packages/54/0c/d01aa759fdc501a58f431eb594a17495f15b88da142ce14b5845662c13f3/pip-20.0.2-py2.py3-none-any.whl
Collecting setuptools
  Using cached https://files.pythonhosted.org/packages/a0/df/635cdb901ee4a8a42ec68e480c49f85f4c59e8816effbf57d9e6ee8b3588/setuptools-46.1.3-py3-none-any.whl
Collecting wheel
  Using cached https://files.pythonhosted.org/packages/8c/23/848298cccf8e40f5bbb59009b32848a4c38f4e7f3364297ab3c3e2e2cd14/wheel-0.34.2-py2.py3-none-any.whl
Installing collected packages: pip, setuptools, wheel
  Found existing installation: pip 19.2.3
    Uninstalling pip-19.2.3:
      Successfully uninstalled pip-19.2.3
  Found existing installation: setuptools 41.2.0
    Uninstalling setuptools-41.2.0:
      Successfully uninstalled setuptools-41.2.0
Successfully installed pip-20.0.2 setuptools-46.1.3 wheel-0.34.2
real	0m 2.53s
user	0m 2.05s
sys	0m 0.19s
/ # rm -rf env && time python3 -m venv env --clear && time env/bin/python -m pip --disable-pip-version-check install -I --upgrade pip setuptools wheel
real	0m 1.77s
user	0m 1.61s
sys	0m 0.15s
Collecting pip
  Using cached https://files.pythonhosted.org/packages/54/0c/d01aa759fdc501a58f431eb594a17495f15b88da142ce14b5845662c13f3/pip-20.0.2-py2.py3-none-any.whl
Collecting setuptools
  Using cached https://files.pythonhosted.org/packages/a0/df/635cdb901ee4a8a42ec68e480c49f85f4c59e8816effbf57d9e6ee8b3588/setuptools-46.1.3-py3-none-any.whl
Collecting wheel
  Using cached https://files.pythonhosted.org/packages/8c/23/848298cccf8e40f5bbb59009b32848a4c38f4e7f3364297ab3c3e2e2cd14/wheel-0.34.2-py2.py3-none-any.whl
Installing collected packages: pip, setuptools, wheel
Successfully installed pip-20.0.2 setuptools-46.1.3 wheel-0.34.2
real	0m 1.83s
user	0m 1.47s
sys	0m 0.18s
```

without -I: 2.53s
with -I: 1.83s

overhead of python -m venv env --clear: 1.75s (same in both cases)

docs/changelog.md
mgedmin: Would 
```suggestion
- [enhancement] Insulate from problems caused by Debian/Ubuntu changes made to their version of pip (#386).
```
work for you?

gaborbernat: Debian decided lately to package pip by trying to devendor six from within pip... the problem is the way they did this unvendoring does not work with venvs... Not to mention the pip within the venv is supposed to be the latest pip (or the bundled version), but either way not a reference to the operating system pip, but a full copy on its own that install/uninstallable. 

Anyways I've purposefully did not do any changes here as IMHO this should be fixed on Debian side and not our end. Just throwing my ideas out there, however not planning to accept this PR myself. Maybe other maintainers are more agreeable to it, in which case I'll not stop it 👍

cs01: The optimal solution is for Debian to fix this issue. But from a user perspective, it looks like pipx is broken, and there is no way for us to control Debian. Is there a timeline for the fix on the Debian side?

mgedmin: Do you have a link to the bug in the Debian bug tracker?  (This assumes anyone reported it to Debian already.)

Once it's fixed in Debian, it should be possible to backport the fix to Ubuntu 20.04 LTS by creating a SRU.  It's just tedious work.

In the meantime I suspect users will keep filing duplicate bug reports or looking for alternative projects...

src/pipx/shared_libs.py
ivanov: > which incidentally populate ~/.pip/cache so the measurements are independent of network speed ...

Unless I'm misunderstanding something - this comment highlights that time measured avoids one of the most relevant impacts - all users will have to download pip. Here are the docs for `-I`

```
-I, --ignore-installed      Ignore the installed packages, overwriting them. This can 
                              break your system if the existing package is of a different version
                              or was installed with a different package manager!
```
So this would *always* ignore the system pip (possibly re-downloading it) for everyone.

ivanov:
I'm reposting my code comment reply here:

@mgedmin wrote:
>    $ docker run -ti --rm python:3.8-alpine sh
... some experiments to get the commands right ...
... which incidentally populate ~/.pip/cache so the measurements are independent of network speed ....

Unless I'm misunderstanding something - this comment highlights that time measured avoids one of the most relevant impacts - all users will have to download pip. Here are the docs for -I
```
-I, --ignore-installed      Ignore the installed packages, overwriting them. This can 
                              break your system if the existing package is of a different version
                              or was installed with a different package manager!
```
So this would always ignore the system pip (possibly re-downloading it) for everyone. I don't think that's great. I think the approach in #388 is sufficient for getting a workable pip out of how it is packaged in Debian that does not force everyone else to reinstall pip.

docs/changelog.md
ivanov: Doesn't look like anyone's reported it to Debian, so I went ahead and did just now: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=958764

src/pipx/shared_libs.py
mgedmin: That is a good point, but as you saw in my terminal log, pip is downloaded and installed irrespective of the -I flag.

I'm not familiar with venv and I don't know where it found the old outdated versions of pip/setuptools that got installed in my experimental venv.  Do you know how I could reproduce (in Docker?) a situation where venv installs up-to-date versions if pip/setuotools that don't require an upgrade?

cs01:
Where are we on this? 

Debian has fixed its pip installation thanks to @ivanov (https://github.com/pipxproject/pipx/pull/388#issuecomment-625560424). Is this still needed or can it be closed? I'm have no problem with it if it causes a slight slowdown because it only affects the shared libraries, which are installed rarely, and thus the impact to the user will be minimal.

I am curious if adding this would sidestep various inconsistencies in the environment though, which might be nice since it would ensure the same pip source code is used for everyone. In that case we may want to consider adding this even though Debian has fixed it's vendoring issue upstream.

mgedmin:
Rebased on master to resolve the changelog merge conflict.

src/pipx/shared_libs.py
cs01: > pip is downloaded and installed irrespective of the -I flag.
> That is a good point, but as you saw in my terminal log, pip is downloaded and installed irrespective of the -I flag.

The log shows pip was cached, but in any case, the shared libs directory is created rarely, and installing a version of pip that is guaranteed to be good is appealing. Being able to install packages from the internet is required for pipx to work, so I don't think pip, setuptools, and wheel being downloaded every once in a while will have a significant negative impact on user experience.

Is this actually overwriting the system pip or just the pip in the venv? If it's the former I don't think we should land it because we could be breaking OS's. If it's the latter, I think we should merge this since it will make pipx hermetic to whatever weird system things are going on. Does anyone know?

cs01: Another q: should `--force-reinstall` be preferred? https://stackoverflow.com/questions/51913361/difference-between-pip-install-options-ignore-installed-and-force-reinstall indicated the follwing

> "Otherwise, you may get obscure import errors after reinstall due to stale modules still available in sys.path."

cs01: There is a scary warning in the pip docs:

> Ignore the installed packages, overwriting them. This can break your system if the existing package is of a different version or was installed with a different package manager!

https://pip.pypa.io/en/stable/reference/pip_install/

I don't know enough about the relationship between OS packages vs how `--ignore-installed` or `--force-reinstall` affects them, so unfortunately I don't feel comfortable merging this.

PR #398, opened 2020-05-04
add env var USE_EMOJI

### LINKED ISSUE #376: Remove emojis by default

cs01:
fixes #376

cc @traverseda 

We had some trouble with https://github.com/pipxproject/pipx/pull/377 so I just made a new PR.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -6,6 +6,7 @@ dev
 - [refactor] Moved all commands to separate files within the commands module (#255).
 - [bugfix] Continue reinstalling packages after failure
 - [bugfix] Hide cursor while pipx runs
+- [feature] Add environment variable `USE_EMOJI` to allow enabling/disabling emojies (#376)
 
 0.15.1.3
 
--- a/src/pipx/commands/uninstall.py
+++ b/src/pipx/commands/uninstall.py
@@ -6,7 +6,7 @@ from pipx.util import WINDOWS, rmdir
 
 
 from pipx import constants
-from pipx.emojies import hazard, stars
+from pipx.emojies import hazard, stars, sleep
 from pipx.venv import Venv, VenvContainer
 
 
@@ -15,7 +15,7 @@ def uninstall(venv_dir: Path, package: str, local_bin_dir: Path, verbose: bool):
     packages.
     """
     if not venv_dir.exists():
-        print(f"Nothing to uninstall for {package} 😴")
+        print(f"Nothing to uninstall for {package} {sleep}")
         app = which(package)
         if app:
             print(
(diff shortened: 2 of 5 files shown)


itsayellow:
I'd still be happy with emoji-on by default, but I can live with it as an opt-in.  

My guess is that if emoji are opt-in, nobody will opt in and we almost may as well remove emoji support.

As a counterpoint to "emoji are a javascript thing" I'd like to point out that `black` uses emoji in its status messages.

cs01:
This is emoji-on by default

itsayellow:
> This is emoji-on by default

Ah ok!  Sorry I misunderstood.

traverseda:
@itsayellow It would be nice if you could give a thumbs down on #376 since it would help maintainers get a better idea of what users/contributors actual preferences are? Not that I expect this project to be a straight democracy or be run by some kind of mob rule, just that it's good to gather data on it.

itsayellow:
> @itsayellow It would be nice if you could give a thumbs down on #376 since it would help people get a better idea of what users/contributors actual preferences are? Not that I expect this project to be a straight democracy or be run by some kind of mob rule, just that it's good to gather data on it.

It's nice that you're soliciting broader response, thanks.  I put a thumbs down there, as requested, but in general my thumbs-down is mostly only for making emoji default off.  I don't have a problem with a mechanism that allows people to disable emoji.  Nor do I have a problem disabling it for platforms like Windows where the implementation can be problematic.

traverseda:
> I put a thumbs down there, as requested, but in general my thumbs-down is mostly only for making emoji default off.

Of course, yeah. An option to disable emoji is completely non-controversial. Personally I'm happy with this solutions, and I'm not going to push very hard to default to disabling emoji. Originally that ticket was just "remove all emoji", I've edited it to instead be "default to no emoji".

Right now "remove all emoji" had overwhelming support, and I suspect that's not realistic. I'd like to get a better sample size so I can actually have relevant data on what people think of emoji. Of course selection bias would still be a major issue, because probably only people that have issues with emoji will seek out that ticket...

Still, worth noting down your feedback and I encourage anyone else who has an opinion do the same, so we can get slightly better data about it.

cs01:
> Implementation looks fine to me. Could we have some tests?

Sure, I will add some

> Also I can use this to make the USE_EMOJI env variable the beginning of a standard, which is nice.

Sounds like you are on a quest to eradicate emojies! Haha. Glad to play a part in it.

PR #401, opened 2020-05-05
Use symlinks on Windows if they are available

pfmoore:
This checks if symlinks are usable on Windows (they are on Windows 10, with "developer mode" enabled, on Python 3.8 and later) and uses them if they are. The main advantage is that with symlinks, there's no problem with overwriting the command executables when upgrading - so it's possible to use pipx to manage itself :-)

The checking approach is basically taken from virtualenv, so it's had some real-world use in that.

### DIFF
--- a/src/pipx/commands/common.py
+++ b/src/pipx/commands/common.py
@@ -9,7 +9,7 @@ import userpath  # type: ignore
 from pathlib import Path
 from shutil import which
 from tempfile import TemporaryDirectory
-from typing import List
+from typing import Dict, List
 
 from pipx import constants
 from pipx.colors import bold, red
@@ -27,7 +27,7 @@ def expose_apps_globally(
         _symlink_package_apps(local_bin_dir, app_paths, package, force=force)
 
 
-_can_symlink_cache = {}
+_can_symlink_cache: Dict[Path, bool] = {}
 
 
 def _can_symlink(local_bin_dir: Path) -> bool:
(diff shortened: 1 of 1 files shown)


cs01:
This is cool, thanks!

There are some other places that will need to be updated
* _get_exposed_app_paths_for_package
* uninstall

src/pipx/commands/common.py
```diff
@@ -20,12 +21,36 @@
...
     local_bin_dir: Path, app_paths: List[Path], package: str, *, force: bool
 ):
-    if WINDOWS:
+    if not _can_symlink(local_bin_dir):
         _copy_package_apps(local_bin_dir, app_paths, package)
     else:
         _symlink_package_apps(local_bin_dir, app_paths, package, force=force)
 
 
+_can_symlink_cache = {}
```
cs01, line 30: would be nice to add mypy types, mostly for documentation so people reading the code can guess how the dict is used

pfmoore: Thanks, will do. I forgot you could annotate variables.

pfmoore:
> There are some other places that will need to be updated

I'll take a look - thanks for the review.

pfmoore:
I've updated `_get_exposed_app_paths_for_package`. If I'm reading the code correctly, the only place in uninstall that assumes Windows == no symlinks is when we don't have a `pipx_metadata.json` file. And since anything like that will have been created before symlink support was added, I think that assumption remains valid.

I could fix the uninstall code [here](https://github.com/pipxproject/pipx/blob/master/src/pipx/commands/uninstall.py#L59) to just use the symlink code everywhere (on Windows with no symlinks, I'm pretty sure it will just return nothing, so will behave the same as the current code), but I'm concerned that I might break some edge case so I'd prefer not to unless you think it's important.

cs01:
>  If I'm reading the code correctly, the only place in uninstall that assumes Windows == no symlinks is when we don't have a pipx_metadata.json file. And since anything like that will have been created before symlink support was added, I think that assumption remains valid.

You're right, I forgot about the new metadata.

The only other thing is changelog should be updated, but I can do that. Thanks again.

PR #404, opened 2020-05-11
increment version to 0.15.1.4

cs01:
New release 0.15.1.4. When this is approved I will publish to PyPI. 

There isn't a particular reason I'm making this now other than I want to keep a fast publishing cadence to keep the project healthy and prevent any cobwebs from forming :smile: .


cs01:
Abandoning in favor of https://github.com/pipxproject/pipx/pull/408

PR #408, opened 2020-05-17
update docs and increment version

cs01:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,12 +1,15 @@
 dev
 
-- Improved error reporting during venv metadata inspection.
+0.15.2.0
+
+- [feature] Use symlinks on Windows when symlinks are available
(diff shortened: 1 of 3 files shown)


itsayellow:
Sorry coming late...
Should this have a github release as well?

cs01:
No problem. Yes there should already be one. Is there not?

itsayellow:
I see 15.3.1 but no 15.2.0 (which I assume is this version?)

itsayellow:
Oh there are mini references to 15.2.0 and 15.3.0 in the Releases tab.  Not sure why they're different than other releases.  Does that mean they have a tag, but were not specifically added as a "Release"?

cs01:
I pushed a tag directly but didn't make the release through github.

> Does that mean they have a tag, but were not specifically added as a "Release"?

Looks like that is the case.

Issue #400, opened 2020-05-04
Remove the "Assuming app is 'xxx.exe' (Windows only)" message


pfmoore:
**How would this feature be useful?**
This message doesn't really add any value (at least in my opinion) and it clutters up the output of the actual command being run. Removing it makes `pipx run xxx` look essentially identical to running an installed copy of `xxx` directly.

**Describe the solution you'd like**
Omit the message.

**Describe alternatives you've considered**
The message could be suppressed with a `--quiet` option, but in practice I'd just use that option all the time, and there would be a question of how it interacts with `--verbose`.

itsayellow:
I agree.  There's no option for pipx to use a non-exe-ending app on Windows, so maybe while being well-intentioned to let the user know what's happening, it's not really necessary or all that helpful.

pfmoore:
Raised #415 to address this (it makes the message info-level, so it still shows with `-v`, rather than completely removing it).

uranusjr:
#415 is merged.

PR #415, opened 2020-05-31
Set assuming app is xxx.exe message to info level

pfmoore:
Rather than completely remove the message, set it to "info" level, so it will show if `-v` is specified.


pfmoore:
Fixes #400

PR #418, opened 2020-06-07
expanded pipx code from single markdown

metinsenturk:
## Sample code Separated to Single Markdown Items

- Separated sample package installs/ runs to single markdown items. Easier to copy from the documentation.
- If you ever enable `toc.permalink` for MKDocs, separating each code will have nice anchor links for each sample package and could potentially help search results.

PS: I checked the `changelog.md` file for any documentation update logs, but couldn't find any. That's why I am not adding an entry, but let me know otherwise.


cs01:
Thanks! Nice tip about `toc.permalink`, I'll have to check that out.

PR #425, opened 2020-06-17
Clean up package specifiers before saving to metadata

### LINKED ISSUE #422: Do not store version specifiers in pipx metadata
Currently we record verbatim to `package_or_url` whatever package specification the user submits to `install`.  This has the effect of making a package specification with a version specifier not able to be upgraded in the future.

This change to pipx behavior would allow a specific version to be specified and installed, but still allow it to be upgraded later.

This also would be one step to enabling useful behavior for a future pipx freeze/unfreeze functionality that specifies all versions during an "unfreeze" installation.

During `pipx install`, install based on whatever requirement specifier the user submits, but strip off any version specifiers before storing `package_or_url` to metadata.

There could also be a `--force` flag in `upgrade`, or `upgrade` could simply strip off the version specifier present in the metadata itself (always ignoring the version specifier.)

itsayellow:
Fixes #422 

I'm posting this to also look for comments on the code.

It adds a pipx dependency on `packaging`.  I tried to think of a good way to avoid adding a dependency, but couldn't.  My reasons for adding `packaging`:
* We need to parse the package specifier in order to remove the version specifier
* By parsing, we can remove any markers (e.g. `python_version>3.6`) from the package specifier which seem to me to be incompatible with use in pipx
* I was able to get rid of `abs_path_if_local` which frankly seemed hacky to me (despite me originally writing it). The new implementation doesn't need to do a pip search (like `abs_path_if_local` did) which should make it faster too.
* I was able to get rid of the old version of `valid_pypi_name` which on examination I realized would fail anyway if there was any specifier beyond a bare name (e.g. it would fail in the case of a version specifier)
* Finally, parsing the package specifier is so easy and clean if we have a dependency do it.  Otherwise we'd be reinventing the wheel.

I made the dependency requirement `packaging>=20.0` so we can have type hints for mypy.

I believe the code works well.  I tested it by installing a specific version `pipx install black==18.9b0`.  With this PR I could successfully upgrade that package in the future.  With the current pipx this is not possible.

Again, this functionality I believe is very useful for a freeze/unfreeze operation so that unfreezing to specific versions that would still allow upgrading them in the future.


src/pipx/util.py
```diff
@@ -135,13 +134,6 @@ def run(cmd: Sequence[Union[str, Path]], check=True) -> int:
     return returncode
 
 
-def valid_pypi_name(package_name: str) -> bool:
-    # https://www.python.org/dev/peps/pep-0508/#names
-    return bool(
-        re.search(r"^([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])$", package_name, re.I)
-    )
```
uranusjr, line 142: I believe this is equivalent?

```python
return re.match(r"^([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])$", package_name, re.I)
```

cs01:
Awesome work. Can some unit tests be added?

itsayellow:
@cs01, just to be completely explicit about one thing I'm doing here: I'm stripping off the version specifier from `package_or_url` before saving it to the metadata.  This is to allow upgrades later.

Just wanted to make sure that effect didn't accidentally get hidden among all the other cleanups I'm doing.

itsayellow:
@cs01, I think this should be good to go now.  Not sure if you wanted to give it another once-over look before merging.

With the new unit tests, I found and fixed at least one bug. (Huh!  Funny that!)

There is something flaky that I can't figure out going on with the last `test_animate.py` test.  I thought just adding more extra time to the `sleep()` statement should fix it, but too much extra time seems to break it, which I don't understand.  It probably needs looking at.

itsayellow:
Oops!, wait, I made a mistake I just saw.  `valid_pypi_name` is stripping everything except the name from package_or_url, so now we cannot specify versions, extras, etc.

Stop the presses, I need to fix this.

itsayellow:
OK, my deleted comment was a false alarm.  I think the code is fine.

@cs01 let me know if you want to take a look at the new code.  Otherwise I'm ready to merge.

itsayellow:
I also fixed `test_animate`, it should be more robust.  `animate` is a little tricky to test because it fires off another thread.  In the macos tests especially, we needed two separate wait times (instead of just the original one extra time.)
1. Extra time for the animate thread to write out all of the text frames (in case it started late.)
2. Extra time after the `animate` context is finished before sampling `capsys.readouterr()`.  We needed to wait before sampling the pytest `capsys` because the `animate` thread might shut down a little late.  If it printed some extra characters after we read `capsys`, they would end up in the next invocation of `check_animate_output` because that function is called multiple times during the same test.

PR #426, opened 2020-06-18
Consistent use of PACKAGE and APP in README examples

spazm:
Consistent use of PACKAGE and APP in examples
    
* Guideline: use ALL CAPS to denote an item that the user will provide
* Removes `$` before PACKAGE and APP in spec examples for consistency across the document.   
* e.g.:
     ```
    pipx run APP [ARGS...]
    ```
    
    ```
    pipx install PACKAGE
    ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,12 +195,12 @@ Re-running the same app is quick because pipx caches Virtual Environments on a p
 
 If the app name does not match that package name, you can use the `--spec` arguemnt:
 ```
-pipx run --spec $PACKAGE $APP
+pipx run --spec PACKAGE APP
 ```
 
 You can also use the `--spec` argument to run a specific version, or use any other `pip`-specifier:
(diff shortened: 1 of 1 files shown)


itsayellow:
Thanks for the help with the docs!  As you point out, the whole README is inconsistent, and I think it would be better to go completely with a no-dollar-sign approach to the all-caps arguments, so it matches the appearance of `pipx install --help`, `pipx run --help`, etc.

Could you instead remove this and the other dollar signs?  By my count there are just 3 current dollar signs on arguments right in this location (i.e. lines 198-203)

spazm:
Will do

spazm:
Force pushed an update to this PR.  

I can't edit the PR title to remove the $ since we went the other way.

> Consistent use of PACKAGE and APP in readme exampes.

PR #428, opened 2020-06-19
Log pipx version at info level (#423)

### LINKED ISSUE #423: include the pipx version when the --verbose flag is passed

Version will be available in bug reports without running a separate command

User logger to log the version

spazm:
Log pipx version at info level, ensuring it is in the logs when running verbosely.

closes #423

### DIFF
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -543,6 +543,7 @@ def setup(args):
     else:
         logging.basicConfig(level=logging.WARNING, format="%(message)s")
 
+    logging.info(f"pipx version is {__version__}")
     mkdir(constants.PIPX_LOCAL_VENVS)
     mkdir(constants.LOCAL_BIN_DIR)
     mkdir(constants.PIPX_VENV_CACHEDIR)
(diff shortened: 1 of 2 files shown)


src/pipx/main.py
```diff
@@ -543,6 +543,7 @@ def setup(args):
     else:
         logging.basicConfig(level=logging.WARNING, format="%(message)s")
 
+    logging.info(f"Pipx version is {__version__}")
```
cs01, line 546: this is a total nit, but could `pipx` be changed to lowercase?

spazm:
Will do.

I copied the usage of what --version produces.

src/pipx/main.py
spazm: Done!  
I thought I'd copied the casing from somewhere else, but apparently I did not.  :)

cs01:
Looks like there are merge conflicts that need to be fixed before I can land

spazm:
Fixed the conflict in `docs/changelog.md`.

PR #430, opened 2020-06-21
Use `source` instead of `.` (#429)

### LINKED ISSUE #429: `.` command is deprecated as for fish shell
It may be a little from a bug, but currently `. ` is deprecated (https://github.com/fish-shell/fish-shell/issues/310).
Therefore, `pipx completions` command should suggest to use `source` command instead of `.` command.

> . (a single period) is an alias for the source command. The use of . is deprecated in favour of
source, and . will be removed in a future version of fish.

(This is quoted from `source --help`)

**Describe the bug**

```
$ pipx completions

:

fish:
    register-python-argcomplete --shell fish pipx | .
```

**Expected behavior**

```
$ pipx completions

:

fish:
    register-python-argcomplete --shell fish pipx | source
```

eduidl:
I fixed #429.

### DIFF
--- a/src/pipx/constants.py
+++ b/src/pipx/constants.py
@@ -71,7 +71,7 @@ tcsh:
     eval `register-python-argcomplete --shell tcsh pipx`
 
 fish:
-    register-python-argcomplete --shell fish pipx | .
+    register-python-argcomplete --shell fish pipx | source
 
 """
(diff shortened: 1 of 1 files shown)


itsayellow:
Thanks @ediuidl !

PR #436, opened 2020-06-25
Make `ensurepath` also ensure pip-user-installed pipx script is in path

### LINKED ISSUE #427: ensurepath should ensure pipx's binary path is in PATH
**Describe the bug**

Our install directions in the absence of Homebrew direct the user to install pipx using pip to the python user install directory.  But this directory's `bin/` or `Scripts/` subdirectory may not already be in a user's path.

We should use `python3 -m pipx ensurepath` to also ensure that the path to the `pipx` script itself is in the user's path.

Ref: https://github.com/pipxproject/pipx/issues/421#issuecomment-646243397

itsayellow:
Fixes #427 

This PR changes `pipx ensurepath` to test to see if pipx has been installed using `pipx install --user` and if it has, then ensure that the path to the pipx script itself is in the user's PATH along with pipx's packages bin directory.

It reports on a per-path basis the status ("successfully added", "already added", "needs shell restart").  It also prints summary messages after all paths have been handled similar to the previous `ensurepath`.

### DIFF
--- a/makefile
+++ b/makefile
@@ -1,7 +1,10 @@
-.PHONY: test docs develop build publish publish_docs
+.PHONY: test docs develop build publish publish_docs lint
 
 develop:
 	pipx run --spec=git+https://github.com/cs01/nox.git@5ea70723e9e6 nox -s develop
+
+lint:
+	pipx run --spec=git+https://github.com/cs01/nox.git@5ea70723e9e6 nox -s lint
 
 test:
 	# TODO use `pipx run nox` when nox supports venv creation (and thus
--- a/makefile
+++ b/makefile
@@ -4,7 +4,7 @@ develop:
 	pipx run --spec=git+https://github.com/cs01/nox.git@5ea70723e9e6 nox -s develop
 
 lint:
-	pipx run --spec=git+https://github.com/cs01/nox.git@5ea70723e9e6 nox -s lint
+	pipx run nox -s lint
 
 test:
 	# TODO use `pipx run nox` when nox supports venv creation (and thus
--- a/src/pipx/commands/ensure_path.py
+++ b/src/pipx/commands/ensure_path.py
@@ -1,6 +1,7 @@
 import logging
 from pathlib import Path
 import site
+import sys
 import textwrap
 from typing import Tuple, Optional
 
@@ -117,6 +118,7 @@ def ensure_pipx_paths(force: bool):
             + "\n"
         )
     elif not need_shell_restart:
+        sys.stdout.flush()
         logging.warning(
             textwrap.fill(
                 "All pipx binary directories have been added to PATH. "
--- a/src/pipx/commands/ensure_path.py
+++ b/src/pipx/commands/ensure_path.py
@@ -114,7 +114,7 @@ def ensure_pipx_paths(force: bool):
         logging.warning(
             textwrap.fill(
                 "All pipx binary directories have been added to PATH. "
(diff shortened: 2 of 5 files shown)


src/pipx/commands/ensure_path.py
```diff
@@ -1,42 +1,140 @@
...
+        test_paths = (
+            userbase_path / "bin" / "pipx",
+            Path(site.getusersitepackages()).resolve().parent / "Scripts" / "pipx.exe",
+        )
+        for test_path in test_paths:
+            if test_path.exists():
+                pipx_bin_path = test_path.parent
+                break
+
+    return pipx_bin_path
```
uranusjr, line 59:
```python
import os
import sysconfig

if os.name == "posix":
    scheme = "posix_user"
elif os.name == "nt":
    scheme = "nt_user"
else:
    raise RuntimeError(f"unsupported operating system {os.name!r}")

dirpath = Path(sysconfig.get_path("scripts", scheme=scheme))
for name in ["pipx", "pipx.exe"]:
    if Path(dirpath, name).exists():
        return dirpath
```

itsayellow: This seems good to me also.  I can modify the code to use sysconfig if people would like me to.

itsayellow:
BTW I've tested this on Windows and macOS, using as many different install possibilities as I could (user, non-user, rebooted shell, non-rebooted shell)

makefile
```diff
@@ -1,8 +1,11 @@
-.PHONY: test docs develop build publish publish_docs
+.PHONY: test docs develop build publish publish_docs lint
 
 develop:
 	pipx run --spec=git+https://github.com/cs01/nox.git@5ea70723e9e6 nox -s develop
 
+lint:
+	pipx run --spec=git+https://github.com/cs01/nox.git@5ea70723e9e6 nox -s lint
```
cs01, line 7: the fork of nox is only needed when creating venvs in the nox session. lint doesn't need it, it can just do `pipx run nox -s lint`.

src/pipx/main.py
```diff
@@ -491,7 +493,7 @@ def _add_ensurepath(subparsers):
         action="store_true",
         help=(
             "Add text to your shell's config file even if it looks like your "
-            f"PATH already has {str(constants.LOCAL_BIN_DIR)}"
+            "PATH already contains pipx binary path(s)."
```
cs01, line 496: I have been trying to avoid the word "binary" since it's kind of ambiguous. I try to use app or application instead. Is there something more precise that can be used here?

uranusjr: Maybe “PATH alreay contains the 'pipx' command”?

itsayellow: Maybe: `PATH already contains paths to pipx and pipx-installed apps.`

Does that sound good?

itsayellow: (Just committed the above change.)

src/pipx/commands/ensure_path.py
cs01: Is this resolved?

itsayellow: I wasn't sure if @uranusjr was just building awareness for `sysconfig` or if he wanted me to change it. 🙂  I was going to leave it, but can change it if people wish.

cs01: Either way is fine with me

PR #441, opened 2020-06-28
GitHub Actions for CI

### LINKED ISSUE #440: Use Github Actions instead of Travis on push

We could use native Github Actions to do the same push-triggered verifications that we currently do using Travis CI.

It would remove Travis from our flow, removing one more moving part, and keeping all committing actions within Github, which seems more reliable to me.

It would also make it easier for us to in the future integrate other actions, such as doing a lot of housekeeping on a release.  For example on a Github release, we could rebuild the docs and publish them, and submit to PyPI.

I have a branch in my repo up and running using Github Actions that currently replicates our Travis CI
push-triggered verifications and tests.
https://github.com/itsayellow/pipx/tree/github-actions

We don't strictly speaking **need** to do this, the current Travis verifications work.  But I thought this might conceptually be simpler, and thought I'd submit it for discussion since I have it completely implemented and it would be ready to drop in now as a PR.

itsayellow:
Closes #440 

This replicates our Travis CI on-push verifications and adds a macos target.


cs01:
Did the actions run? How do we see the results passed before merging?

itsayellow:
> Did the actions run? How do we see the results passed before merging?

I see the actions and results here, hopefully everyone else can too:
https://github.com/itsayellow/pipx/actions

In the end, it was only the tests-macos target that had issues--the last one was with in `test_animate.py` (`test_line_lengths_emoji()`), which needed more extra sleep time to ensure the animate thread would print everything it was supposed to for the assert statement.  I only added 0.1s (0.5 -> 0.6) to that extra sleep time, so it may need more to be always robustly passing.

itsayellow:
I had an inspiration and managed to simplify `on-push.yml` a lot because I finally understood more of the ins and outs of `strategy-matrix`, so I could make only one `tests` job that is just parameterized properly with the various os flavors and python versions.

itsayellow:
On question I had: I currently have `ubuntu-latest` but specific versions: `windows-2019` and `macos-10.15`.  I wasn't sure what the best route was, but I'm guessing they should either all be `-latest` or all to some specific OS version.  Or possibly `-latest` for `lint` and `docs` but specific OS versions for the others.

Any thoughts?

itsayellow:
Hmm, I just merged some upstream pipx commits and this error happened on macos.  It's not obvious to me how this could fail now and not before, or why it's failing in general:
```
=================================== FAILURES ===================================
____________ test_auto_update_shared_libs[1591476297.571466-False] _____________

capsys = <_pytest.capture.CaptureFixture object at 0x10c2bff50>
pipx_temp_env = None, mtime = 1591476297.571466, needs_upgrade = False

    @pytest.mark.parametrize(
        "mtime,needs_upgrade",
        [
            (now - shared_libs.SHARED_LIBS_MAX_AGE_SEC - 600, True),
            (now - shared_libs.SHARED_LIBS_MAX_AGE_SEC + 600, False),
        ],
    )
    def test_auto_update_shared_libs(capsys, pipx_temp_env, mtime, needs_upgrade):
        shared_libs.shared_libs.create([], verbose=True)
        shared_libs.shared_libs.has_been_updated_this_run = False
    
        access_time = now  # this can be anything
        os.utime(shared_libs.shared_libs.pip_path, (access_time, mtime))
    
>       assert shared_libs.shared_libs.needs_upgrade is needs_upgrade
E       assert True is False
E        +  where True = <pipx.shared_libs._SharedLibs object at 0x10c13c850>.needs_upgrade
E        +    where <pipx.shared_libs._SharedLibs object at 0x10c13c850> = shared_libs.shared_libs

tests/test_shared_libs.py:24: AssertionError
```

itsayellow:
Ah!  Actually I see how this failed.  `now` is defined at import, and the tests for macos on this particular commit took over 12 minutes, so that mocking mtime as `now - shared_libs.SHARED_LIBS_MAX_AGE + 600` is (12 minutes later) actually correctly needing to be upgraded (2 minutes after `SHARED_LIB_MAX_AGE`)

cs01:
This is really great. I just started learning and using GitHub actions in another project and I'm a fan.

> On question I had: I currently have ubuntu-latest but specific versions: windows-2019 and macos-10.15. I wasn't sure what the best route was, but I'm guessing they should either all be -latest or all to some specific OS version. Or possibly -latest for lint and docs but specific OS versions for tests. Any thoughts?

When python had pre-releases that were soon be real actual releases, I would make them run but have their success be optional. 

However, the OS's on github actions under "latest" are not pre-releases (https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on). In fact, for ubuntu it looks like it's the latest LTS release (18.04), rather than 20.04 which is also available. Thus I think we should prioritize things working on the latest release rather than arbitrary past OS's. So let's just use `-latest` for everything.

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,81 @@
+# https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Push Verifications
+
+on: [push]
```
cs01, line 5: I just found that this will only run *after* merging pull requests.

I think we want something more like this. The syntax is a little awkward, but it will trigger the job on all pull requests, pushes to the master branch, and releases.
```
on:
  pull_request:
  push:
    branches:
      - master
  release:
```

itsayellow: Would the sub-specification for push of `branches: master` prevent the workflow from running in the user's non-master branch (e.g. before submitting as a PR)?  Because right now it works fine in my local branch, it just doesn't show up in the PR.

cs01: > prevent the workflow from running in the user's non-master branch

Yes that is my understanding. If we change to `[push, pull_request]`, the tests will run on a user's non-master branch, but we'll get duplicate jobs for everything after a PR is made.

itsayellow: Hmm, so I tried just adding `pull_request` and `release` and leaving `push` alone, but I'm still not seeing it in the PR.  I am still seeing the workflow execute in my local branch.

So it's possible instead of duplicating workflow execution, it's only running in my local branch and not in the PR.  I'll add the `branches: master` and see if I see it then show up in the PR.

cs01: > Hmm, so I tried just adding pull_request and release and leaving push alone, but I'm still not seeing it in the PR. 

I suspect that is because it hasn't landed in the pipx source code yet

itsayellow: huh.  I committed your suggestion and am still not seeing it here in this PR.  And also not in my local branch.  I'll investigate.

itsayellow: Or maybe we won't see it in the pipx repo until we finally merge the PR?

cs01: > we won't see it in the pipx repo until we finally merge the PR?

Yes I think this is the case. Or if you make the branch in `pipxproject/github-actions` rather than your personal repo.

itsayellow: Do we need to prevent duplicate workflows with the `branches:` specification?  

I'm wondering if github would do the right thing and only execute one workflow but have it show up in two different places.

Mostly I would just be sad to lose the CI feedback when developing in my own local branch. 🙂

cs01: No we don't need to. It makes things a little noisier, possibly take longer, and use more GitHub resources (which are free for public repos... crazy). None of those are a dealbreaker. 

We can try merging this with all PRs and pushes turned on and go from there.

cs01:
We'll also have to update the build badge to include results from the github actions rather than travis tests. https://docs.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#adding-a-workflow-status-badge-to-your-repository

.github/workflows/on-push.yml
itsayellow: Maybe we can check in the future and attempt to verify if it's really duplicating workflow executions or just displaying the same workflow in different contexts.

README.md
```diff
@@ -13,7 +13,7 @@
 </p>
 
 <p align="center">
-<a href="https://travis-ci.org/pipxproject/pipx"><img src="https://travis-ci.org/pipxproject/pipx.svg?branch=master" /></a>
+<a href="https://github.com/pipxproject/pipx/actions?query=workflow%3A%22Push+Verifications%22"><img src="https://github.com/pipxproject/pipx/workflows/Push%20Verifications/badge.svg"></a>
```
cs01, line 16: The branch should be specified too

```
badge.svg?branch=master
```

itsayellow:
I changed the name of the workflow to `Test` so that the badge is more logical.  Do you prefer `build` instead?  (That's what the Travis badge says.)

cs01:
> I changed the name of the workflow to `Test` so that the badge is more logical. Do you prefer `build` instead? (That's what the Travis badge says.)

No, I just copy/pasted from their docs. `Test` is fine with me.

itsayellow:
I suppose the badge could use an `alt` tag if we fix the workflow in the future... :shrug:

itsayellow:
As some data points:  I've been poking around python packages with github workflows, and both zulip and black run their tests
```
on: [push, pull_request]
```
https://github.com/zulip/zulip/blob/master/.github/workflows/zulip-ci.yml
https://github.com/psf/black/blob/master/.github/workflows/test.yml

PR #445, opened 2020-07-08
Add support for --suffix option

joshbode:
Optional suffix can be used to install multiple versions of the same
tool side-by-side - e.g.
```
$ pipx install --suffix=_3.1 foo==3.1.2 --> foo_3.1
$ pipx install --suffix=_2.2 foo==2.2.0 --> foo_2.2
```


cs01:
Great work, thanks. @uranusjr @itsayellow this looks good to merge to me. Any thoughts you'd like to share before I do so?

joshbode:
Thank you for the merge :)

Issue #395, opened 2020-04-28
[Q] Default Python interpreter


jaklan:
Hi, I can't find it in docs - in what way the default Python interpreter path is resolved? I'm using Python 3.7 with Homebrew, but I also have Python@3.8 which was installed as a dependency by some library. It's Keg-only, so it's not visible in `/usr/local/bin`. But since then, the packages were installed with Python 3.8 by default and I had to manually use `--python` flag to switch it back to 3.7 each time. I even deleted `.local/pipx/shared` folder and tried to recreate with `reinstall-all`, but 3.8 was used anyway as long as I didn't use `--python` flag.

uranusjr:
The default Python interpreter is whatever `sys.executable` reports. This can cause problems with a few environments, and there are workarounds available, and my [pipx-standalone](https://github.com/uranusjr/pipx-standalone) catually carries a patch to use perform custom Python version detection. Also see #146, #294. I think this is an issue the Homebrew formula maintainers need to address, but maybe pipx can organise the code to make it easier for downstream packagers to patch the default Python inrepreter logic.

jaklan:
@uranusjr thanks for the answer. As I cen see, `pipx` formula in Homebrew depends on Keg-only `python@3.8` (for whatever reason), so as I understand that would be the default interpreter. Is there any way to install `pipx` via brew with another Python version, e.g. 3.7?

AlJohri:
@jaklan No, brew installed pipx always uses the latest python. Homebrew is slowly porting all formula to manually depend on python 3.8. When this is complete, the global brew python will also change from 3.7 to 3.8.

itsayellow:
Closing because #448 clears up documenting this (use `pipx --help` to see what pipx thinks is the default python executable) and also allows setting the default python (use `PIPX_DEFAULT_PYTHON` env variable.)

PR #448, opened 2020-07-13
Add env var to set DEFAULT_PYTHON (add Pyenv support)

### LINKED ISSUE #113: Add pyenv compatibility
I've got a CLI-tool (tmux-res) I built for handling some common tasks in my TMUX-environment which I built a while back. For fun I wanted to try installing it via pipx but it doesn't work.

This is the package:
https://pypi.org/project/tmux-res/

```sh
 $ pyenv global 3.7.2
 $ python --version
Python 3.7.2

 $ pipx install tmux-res
  Could not find a version that satisfies the requirement tmux-res (from versions: )
No matching distribution found for tmux-res
'/Users/hultner/.local/pipx/venvs/tmux-res/bin/python -m pip install tmux-res -q' failed

 $ pip install tmux-res
Collecting tmux-res
  Downloading https://files.pythonhosted.org/packages/09/80/a0294c244aebbdc9f13c06b182a7c1857618da20eabd1c88c4f449dbdf09/tmux_res-0.4.8-py3-none-any.whl
Collecting python-dotenv (from tmux-res)
  Downloading https://files.pythonhosted.org/packages/8c/14/501508b016e7b1ad0eb91bba581e66ad9bfc7c66fcacbb580eaf9bc38458/python_dotenv-0.10.1-py2.py3-none-any.whl
Collecting Click (from tmux-res)
  Using cached https://files.pythonhosted.org/packages/fa/37/45185cb5abbc30d7257104c434fe0b07e5a195a6847506c074527aa599ec/Click-7.0-py2.py3-none-any.whl
Installing collected packages: python-dotenv, Click, tmux-res
Successfully installed Click-7.0 python-dotenv-0.10.1 tmux-res-0.4.8
You are using pip version 18.1, however version 19.0.3 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
```

wren:
Fixes #113.

This keeps sys.executable as the default, but will use the PIPX_DEFAULT_PYTHON environment variable, if defined. This mean no change in functionality for any users until they define the variable.

I'm suggesting the var `PIPX_DEFAULT_PYTHON` (no `PYENV` even though the original issue asked for pyenv support), because this approach will work with any python manager, and I think it might be confusing to enshrine one manager in the var name over others.

Also, it didn't look like the constants are tested, so I didn't update any tests. But let me know if I missed something and I'll be happy to add.

### DIFF
--- a/src/pipx/interpreter.py
+++ b/src/pipx/interpreter.py
@@ -1,3 +1,4 @@
+import os
 import shutil
 import subprocess
 import sys
@@ -56,7 +57,23 @@ def _find_default_windows_python() -> str:
     return python  # This executable seems to work.
 
 
-if WINDOWS:
-    DEFAULT_PYTHON = _find_default_windows_python()
+def _get_sys_executable() -> str:
+    if WINDOWS:
+        return _find_default_windows_python()
+    else:
+        return sys.executable
+
+
+def _get_absolute_python_interpreter(env_python: str) -> str:
+    which_python = shutil.which(env_python)
+    if not which_python:
+        raise PipxError(f"Default python interpreter '{env_python}' is invalid")
+    return which_python
+
+
+env_default_python = os.environ.get("PIPX_DEFAULT_PYTHON")
+
+if not env_default_python:
+    DEFAULT_PYTHON = _get_sys_executable()
 else:
-    DEFAULT_PYTHON = sys.executable
+    DEFAULT_PYTHON = _get_absolute_python_interpreter(env_default_python)
--- a/src/pipx/interpreter.py
+++ b/src/pipx/interpreter.py
@@ -67,7 +67,7 @@ def _get_sys_executable() -> str:
 def _get_absolute_python_interpreter(env_python: str) -> str:
     which_python = shutil.which(env_python)
     if not which_python:
-        raise PipxError(f"Default python interpreter '{env_python}' is invalid")
+        raise PipxError(f"Default python interpreter {repr(env_python)} is invalid")
     return which_python
 
 
--- a/tests/test_constants.py
+++ b/tests/test_constants.py
@@ -4,8 +4,6 @@ from unittest import mock
 import sys
 import pytest  # type: ignore
 
-from helpers import run_pipx_cli
-
 
 @pytest.mark.parametrize(
     "windows, USE_EMOJI, encoding, expected",
@@ -42,10 +40,3 @@ def test_use_emjois(monkeypatch, windows, USE_EMOJI, encoding, expected):
         if USE_EMOJI is not None:
             monkeypatch.setenv("USE_EMOJI", USE_EMOJI)
         assert use_emjois() is expected
-
-
-def test_default_python(monkeypatch, capsys):
-    monkeypatch.setattr(constants, "DEFAULT_PYTHON", "bad_python")
-    run_pipx_cli(["install", "pycowsay"])
-    captured = capsys.readouterr()
-    assert "Default python interpreter 'bad_python' is invalid" in captured.err
--- a/tests/test_constants.py
+++ b/tests/test_constants.py
@@ -4,8 +4,6 @@ from unittest import mock
 import sys
 import pytest  # type: ignore
 
-from pipx.util import PipxError
-
 from helpers import run_pipx_cli
 
 
--- a/tests/test_constants.py
+++ b/tests/test_constants.py
@@ -4,6 +4,8 @@ from unittest import mock
 import sys
 import pytest  # type: ignore
 
+from helpers import run_pipx_cli
+
 
 @pytest.mark.parametrize(
     "windows, USE_EMOJI, encoding, expected",
@@ -40,3 +42,10 @@ def test_use_emjois(monkeypatch, windows, USE_EMOJI, encoding, expected):
         if USE_EMOJI is not None:
             monkeypatch.setenv("USE_EMOJI", USE_EMOJI)
         assert use_emjois() is expected
+
+
+def test_default_python():
+    with mock.patch.object(constants, "DEFAULT_PYTHON", "bad_python"), pytest.raises(
+        FileNotFoundError
+    ):
+        run_pipx_cli(["install", "pycowsay"])
--- a/tests/test_constants.py
+++ b/tests/test_constants.py
@@ -4,6 +4,8 @@ from unittest import mock
 import sys
 import pytest  # type: ignore
 
+from pipx.util import PipxError
+
 from helpers import run_pipx_cli
 
 
@@ -44,7 +46,8 @@ def test_use_emjois(monkeypatch, windows, USE_EMOJI, encoding, expected):
         assert use_emjois() is expected
 
 
-def test_default_python(monkeypatch):
+def test_default_python(monkeypatch, capsys):
     monkeypatch.setattr(constants, "DEFAULT_PYTHON", "bad_python")
-    with pytest.raises(FileNotFoundError):
-        run_pipx_cli(["install", "pycowsay"])
+    run_pipx_cli(["install", "pycowsay"])
+    captured = capsys.readouterr()
+    assert "Default python interpreter 'bad_python' is invalid" in captured.err
--- a/tests/test_constants.py
+++ b/tests/test_constants.py
@@ -44,8 +44,7 @@ def test_use_emjois(monkeypatch, windows, USE_EMOJI, encoding, expected):
         assert use_emjois() is expected
 
 
-def test_default_python():
-    with mock.patch.object(constants, "DEFAULT_PYTHON", "bad_python"), pytest.raises(
-        FileNotFoundError
-    ):
+def test_default_python(monkeypatch):
+    monkeypatch.setattr(constants, "DEFAULT_PYTHON", "bad_python")
+    with pytest.raises(FileNotFoundError):
         run_pipx_cli(["install", "pycowsay"])
--- a/.github/workflows/on-push.yml
+++ b/.github/workflows/on-push.yml
@@ -77,17 +77,8 @@ jobs:
         key: ${{ runner.os }}-pip${{ matrix.python-version }}
     - name: Upgrade pip
       run: python -m pip install --upgrade pip
-    # macOS: Install wheel to allow wheel-building for nox and regex
-    # Build regex wheel into pip cache, so future pipx installs don't compile.
-    # This avoids future compile errors in nox tests because of missing
-    #   header files.
-    - name: Install wheel and build wheel for regex to pip cache
-      if: ${{ runner.os == 'macOS' }}
-      run: |
-        python -m pip install wheel
-        python -m pip install regex
-    - name: Install native-venv version of nox
-      run: pip install git+https://github.com/cs01/nox.git@5ea70723e9e6#egg=nox
+    - name: Install nox
+      run: pip install nox
     - name: Execute Tests
       run: |
         nox --non-interactive --session tests-${{ matrix.python-version }}
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,6 +25,9 @@ repos:
   rev: 3.8.3
   hooks:
   - id: flake8
+    additional_dependencies: [
+        'flake8-bugbear==20.1.4'
+    ]
 - repo: https://github.com/pre-commit/mirrors-mypy
   rev: v0.782
   hooks:
--- a/README.md
+++ b/README.md
@@ -258,7 +258,7 @@ That's it! Those are the most important commands `pipx` offers. To see all of pi
 
 ## Credits
 
-pipx was inspired by [pipsi](https://github.com/mitsuhiko/pipsi) and [npx](https://github.com/zkat/npx). It was created by [Chad Smith](https://github.com/cs01/) and has had lots of help from [contributors](https://github.com/pipxproject/pipx/graphs/contributors). The logo was created by [@IrishMorales](https://github.com/IrishMorales).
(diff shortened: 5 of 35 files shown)


cs01:
I am not a pyenv user (maybe one of the other pipx maintainers is?), so I don't really know how to test this or show that it fixes anything. 

Could you run some commands that show this fixes something, and copy+paste them here so I can see what that looks like?

Assuming this change is useful and we want to merge, the help text will also need to document this environment variable (see `main.py`).

wren:
Generally (for any user): the new env var should let you override the default python used by pipx to install packages. So, if you `export PIPX_DEFAULT_PYTHON=python3.6`, then pipx will use `python3.6` to install by default (but is still override-able by `--python`). 

Specifically (for Pyenv users): this works for with `export PIPX_DEFAULT_PYTHON="$PYENV_ROOT/shims/python"` since Pyenv uses that shim to always point to the correct Python (note: I think `PYENV_ROOT` defaults to `~/.pyenv`).

So, this is testable without Pyenv simply by changing the env var and verifying that Pipx is using the correct Python to install packages.

## Pyenv Tests
I tested this by running the latest pipx release side-by-side with this the source from this PR.

https://asciinema.org/a/347120

```sh
$ pyenv global 3.8.3

$ python --version
Python 3.8.3

$ env | grep -i pipx_default_python
PIPX_DEFAULT_PYTHON=/Users/jonathan/.cache/pyenv/shims/python

$ python src/pipx install jrnl
  installed package jrnl 2.4.3, Python 3.8.3
    - jrnl
done! ✨ 🌟 ✨

$ python src/pipx uninstall jrnl
uninstalled jrnl! ✨ 🌟 ✨

$ pipx install jrnl
  installed package jrnl 2.4.3, Python 3.8.3
    - jrnl
done! ✨ 🌟 ✨

$ pipx uninstall jrnl
uninstalled jrnl! ✨ 🌟 ✨

$ pyenv global 3.6.10

$ python --version
Python 3.6.10

$ python src/pipx install jrnl
  installed package jrnl 2.4.3, Python 3.6.10
    - jrnl
done! ✨ 🌟 ✨

$ python src/pipx uninstall jrnl
uninstalled jrnl! ✨ 🌟 ✨

$ pipx install jrnl
  installed package jrnl 2.4.3, Python 3.8.3
    - jrnl
done! ✨ 🌟 ✨

$ pipx uninstall jrnl
uninstalled jrnl! ✨ 🌟 ✨

```

uranusjr:
I needed to do something similar in https://github.com/pipxproject/pipx/issues/244#issuecomment-580382747. This would also be a good workaround for #395 as well. In general, `sys.executable` is not the best value in many repackaging situations, and this would be a welcoming change. And if it happens to also help pyenv users, more wonderful 😄

cs01:
Looks good to me. I think we should add a unit test before landing. Since the tests on CI generally only have one python version installed, we can set `PIPX_DEFAULT_PYTHON` to be a path that doesn't exist, then test that pipx raises an exception saying the Python interpreter doesn't exist.

wren:
I'm not super familiar with this project, so I'm not sure if I did what you asked, but I think this is ready now.

uranusjr:
It just occured to me, would it be worthwhile to incorporate `shutil.which()` into the resolution so a user can specify (say) `python3` instead of `/usr/bin/python3`? It’s probably a good idea to add some error reporting around this code as well, since the exception may be quite cryptic if the user sets `DEFAULT_PYTHON` to an invalid value.

tests/test_constants.py
```diff
@@ -40,3 +42,9 @@ def test_use_emjois(monkeypatch, windows, USE_EMOJI, encoding, expected):
         if USE_EMOJI is not None:
             monkeypatch.setenv("USE_EMOJI", USE_EMOJI)
         assert use_emjois() is expected
+
+
+def test_default_python(monkeypatch):
+    monkeypatch.setattr(constants, "DEFAULT_PYTHON", "bad_python")
```
cs01, line 48: Should this be `PIPX_DEFAULT_PYTHON`?

cs01:
> It just occured to me, would it be worthwhile to incorporate `shutil.which()` into the resolution so a user can specify (say) `python3` instead of `/usr/bin/python3`? It’s probably a good idea to add some error reporting around this code as well, since the exception may be quite cryptic if the user sets `DEFAULT_PYTHON` to an invalid value.

Agreed. We should raise a `PipxError()` with a clear error message and suggested action for the user if the python interpreter can't be found.

tests/test_constants.py
wren: I'm not sure, but I don't think so. The way constants are set up, environment variables are read before the tests are run, so patching the environment variable itself (`PIPX_DEFAULT_PYTHON`) wouldn't do anything here. So, we need to patch the python variable that holds the environment variable instead.

And I'm fairly new to your code base, so I might be wrong, but it seems to me this is the same way the other constants are handled in the test suite (like patching `LOCAL_BIN_DIR` instead of `PIPX_BIN_DIR` [here](https://github.com/pipxproject/pipx/blob/master/tests/conftest.py#L21)).

wren:
Sounds good. I can add some better error output.

For the use of `shutil.which()`: I'm happy to add if that's what y'all want, but I personally don't think it's desirable behavior. It can lead to some unpredictable results as people move in and out of virtual envs, subshells, etc. I think a full path gives users more control. And if a user really wants this functionality, they can add already do something like `export PIPX_DEFAULT_PYTHON="$(which python)"` to get the same result, so I'm not sure it would add anything other than a larger maintenance burden.

itsayellow:
> For the use of `shutil.which()`: I'm happy to add if that's what y'all want, but I personally don't think it's desirable behavior. It can lead to some unpredictable results as people move in and out of virtual envs, subshells, etc. I think a full path gives users more control. And if a user really wants this functionality, they can add already do something like `export PIPX_DEFAULT_PYTHON="$(which python)"` to get the same result, so I'm not sure it would add anything other than a larger maintenance burden.

I agree with this.  I feel like being explicit in environment config is good.  For our commands in an interactive shell it is nice that `--python` doesn't need a full path, but I don't feel the same way about env variables, and there is possibility of unpredictable results as @wren says.

uranusjr:
Contrary to instincts, a `shutil.which()` may actually make things more explicit. Commands like `python3` would still work without `shutil.which()` since we eventually pass the value to `subprocess`, which does `PATH` resolution on the command anyway. But that resolution logic and behaviour can be different from what user expects, such as [bpo-38905](https://bugs.python.org/issue38905). `shutil.which()`, on the other hand, always behaves the same, and matches the expectation in the majority of cases.

itsayellow:
If we do use `shutil.which` I would also suggest a `logging.info()` statement showing the full resolved path that `shutil.which` returns, so in pipx verbose mode the user (and/or pipx debugging team) knows which python is getting used.

wren:
Hi, just checking back on this.

I just need a final decision on whether or not to include `shutil`, and I can wrap this PR up.

itsayellow:
I'm ok with it, and would like the logging.info() statement to show that we found `PIPX_DEFAULT_PYTHON`, are using it, and what the full resolved path is.

tests/test_constants.py
itsayellow: I think we could do it with `importlib.reload(pipx.constants)` but I'm not sure if there would be side-effects.

uranusjr:
The `shutil` discussion and improvement can be made separatedly. But we’ll probably need to drop the `resolve()` part in that case, since that would covert `python3` to `$PWD/python3` which would be very confusing to the user.

cs01:
@itsayellow @uranusjr do you guys want to request changes or are you good merging like this?

uranusjr:
I suggest improving treatment to relative paths, either by using `resolve()` to turn them absolute, or rejecting outright based on `is_absolute()`.

itsayellow:
I suggest adding a `logging.info()` statement inside of `main.py` `setup()` after statement `logging.info(f"pipx version is {__version__}")`.
Something like:
```
logging.info(f"Default python interpreter: {constants.DEFAULT_PYTHON}")
```

itsayellow:
In general that will make debugging much easier in the future.

wren:
Sounds good. I'll do the requested updates this week. Thanks for the feedback!

gaborbernat:
@wren any progress on this, you still plan to do it?

wren:
Ah, yes. Time got away from me in the last week, sorry! I'll make some time today for it.

itsayellow:
FYI `black` just made a new release hours ago, which is likely why your Lint test is failing.

I'm going to see if I can make a merge that will fix things soon.

itsayellow:
OK, if you merge pipx master now the tests *should* work again.

wren:
I just came back to check on this and realized I didn't hit submit on my comment. Sorry!

--

So, I've made the requested changes. I also added a call to `shutil.which` because it was the simplest way to deal with the relative path handling requirement.

The only test failing is related to typing (since the variable can now be `None` in addition to a string). I can fix this either by updating the type hints in the relevant functions, or by forcing a string. Either way is fine with me. Do y'all have a preference?

uranusjr:
I think it’s reasonable to default to sys.executable when which returns None, i.e. `which(...) or sys.executable`.

wren:
@uranusjr I'm okay with that approach. Just to be clear, though, that would make it impossible to give an error if the env var is invalid since it's immediately with the sys python. Is that okay?

I could also move the `which` call into main.py, and then overwrite `DEFAULT_PYTHON` with the absolute path after it's been validated by `which` (and the relevant error has been thrown). Would you prefer that?

uranusjr:
Ah, I did not think about the error reporting issue if the environment variable is invalid, thanks for the reminder. Maybe we should generate a warning in that case instead. Following the approach of #465 (which moves `DEFAULT_PYTHON` to its own module and would conflict with this PR), we should probably refactor this logic into a function, and raise a `PipxError` is the environment variable is invalid instead.

wren:
Hi, this is updated now. I believe this meets all of the previously established criteria:

- `DEFAULT_PYTHON` is always a string (so no updating of type hints is necessary)
- Supports the new Windows embeddable python code (#465)
- Any env var setting is validated with `shutil.which`
- Sys executable (including the new windows one) is not run through which (since it's redundant)
- A `PipxError` is raised when `which` fails
- `DEFAULT_PYTHON` is updated with an absolute path if validation succeeds
- logging statement in place for help debugging
- The env var can still be overridden by passing `--python` on cli

Anything I've left out?

src/pipx/interpreter.py
```diff
@@ -56,7 +57,23 @@ def _find_default_windows_python() -> str:
...
+    if WINDOWS:
+        return _find_default_windows_python()
+    else:
+        return sys.executable
+
+
+def _get_absolute_python_interpreter(env_python: str) -> str:
+    which_python = shutil.which(env_python)
+    if not which_python:
+        raise PipxError(f"Default python interpreter '{env_python}' is invalid")
```
uranusjr, line 70: `{env_python!r}` or `{repr(env_python)}` would handle the quoting edge cases better.

uranusjr (commented):
One minor nitpick, otherwise this looks good to me!

src/pipx/interpreter.py
wren: Hi, would you mind giving an example of a quoting edge case that `repr` would fix? I thought that f-strings already handled escaping, and since `env_python` is always a string, `repr` seems redundant to me. Happy to add either way, but I just want to understand if I'm missing something.

uranusjr: The main difference is when the path contains a quote character. `repr()` would automatically choose the best quoting scheme, and perform better than the current naive method of simply adding quote characters around the path.

Alternatively you can just not quote the path, like you did for the “python interpreter is” message below. That would cause some confusion when there’s space in the path, but there’s no perfect solution here, unfortunately.

wren: Ah, okay. Thank you for explaining!

uranusjr (approved):

src/pipx/interpreter.py
```diff
@@ -56,7 +57,23 @@ def _find_default_windows_python() -> str:
...
+
+
+def _get_absolute_python_interpreter(env_python: str) -> str:
+    which_python = shutil.which(env_python)
+    if not which_python:
+        raise PipxError(f"Default python interpreter {repr(env_python)} is invalid")
+    return which_python
+
+
+env_default_python = os.environ.get("PIPX_DEFAULT_PYTHON")
```
itsayellow, line 74: Since `env_default_python` is not used anywhere else but this `if` statement, could you combine lines 74,76 into one line for code simplicity and readability:
```
if not os.environ.get("PIPX_DEFAULT_PYTHON"):
```

itsayellow (changes requested):
Looks mostly good to me, just one little nit.

src/pipx/interpreter.py
wren: It's used in two places below, which is why I assigned it to a variable instead of calling `os.environ` twice. Do you still want me to get rid of it?

itsayellow: Oh sorry I see inside the else.  Ok you can forget this.

itsayellow:
Hey @wren , just in case it got lost, could you please list PIPX_DEFAULT_PYTHON and a help blurb for it in main.py around line 50, inside PIPX_DESCRIPTION, under "Optional Environment Variables"?  Then I'll merge this.

wren:
@itsayellow Yup, sorry. I'm in the middle of moving right now, so it'll be a few days before I can get to this.

wren:
@itsayellow All ready for merging. Thanks for your patience!

itsayellow:
@wren, thanks for your persistence as well!

itsayellow (approved):

gaborbernat (approved):

itsayellow:
Just one more thing and I'll do it...updating changelog entry.

PR #450, opened 2020-07-24
Update CI python version from 3.7 to 3.8, README fix

itsayellow:

I didn't add a changelog entry since this was an internal testing issue and not affecting pipx itself.  (But I can if it's needed.)

## Summary of changes

I bumped the python version to 3.8 of all tests that only use one python version (doc, lint, macos tests, windows tests), since that's the current stable python version.

I also fixed the badges on `README.md` so their height matches each other, and gave an `alt` tag to the Test badge.

## Test plan

Tested by running
```
```

### DIFF
--- a/.github/workflows/on-push.yml
+++ b/.github/workflows/on-push.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


itsayellow:
Thanks for the quick approval, @gaborbernat 🙂

PR #451, opened 2020-07-26
List injected versions

itsayellow:

## Summary of changes
When listing injected packages using `pipx list --include-injected`, now the version of the injected package is listed after its name.

## Test plan

Tested by running
```
pipx install pycowsay
pipx inject pycowsay cowsay
pipx list --include-injected
```
In the current pipx, the name only `cowsay` is displayed, in this PR, `cowsay 2.0.3` is displayed.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,4 +1,5 @@
 dev
+- Add version after each injected package name for `pipx list --include-injected`
 - Change metadata recorded from version-specified install to allow upgrades in future.  Adds pipx dependency on `packaging` package.
 - [bugfix] Prevent python error in case where package has no pipx metadata and advise user how to fix.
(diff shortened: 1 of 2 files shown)


itsayellow:
> Should we do this by default rather than hide it behind a flag?

Personally I think it should be the default.  It matches the way we present the main package.  And it seems too small to me to add a whole option to the command.

But if others disagree with me I can definitely change it.

cs01:
I'm fine with this being the default

gaborbernat:
Can you remove the flag then 😀

itsayellow:
Oh wait, maybe I misunderstood.  Did you mean `--include-injected`?  I thought you were talking about another flag.

itsayellow:
@gaborbernat, @cs01, or anyone, if we want to remove `--include-injected`, I can submit a PR for it, just open an Issue.

Issue #444, opened 2020-07-07
"ValueError: Format specifier missing precision" on pipx install <package>


halloleo:
**Describe the bug**

When running `pipx install cowsay` I get the following error:

```
pipx install cowsay
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/lib/python3.7/site-packages/pipx/animate.py", line 78, in print_animation
    cur_line = f"{s} {message:.{max_message_len}}"
ValueError: Format specifier missing precision

Exception in thread Thread-2:
Traceback (most recent call last):
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/lib/python3.7/site-packages/pipx/animate.py", line 78, in print_animation
    cur_line = f"{s} {message:.{max_message_len}}"
ValueError: Format specifier missing precision

done! ✨ 🌟 ✨
  installed package cowsay 2.0.3, Python 3.7.7
  These apps are now globally available
    - cowsay
```

So `cowsay` installs correctly, but something goes wrong before that.

**How to reproduce**

My environment is:

&nbsp; | &nbsp;
-- | -------
OS | macOS Mojave
Python | Version 3.7.7<BR>installed via `brew`<BR>(and pinned)
pipx | Version 0.15.4<BR>installed via `pip`<BR>straight into the main  Python 3.7 environment

itsayellow:
Very interesting.  The only way I can think of this happening is if somehow `shutil.get_terminal_size()` is returning a very small number for the number of columns, which would make `max_message_len` become a negative number.

Can you try to open your python3.7 interactive prompt and execute the following, reporting back what it says, please?
```
import shutil
shutil.get_terminal_size()
```

halloleo:
[...]

PR #452, opened 2020-07-26
Fix bug term size

itsayellow:

## Summary of changes
Fixes bug #444 .

I was unable to recreate that bug specifically, but other
## Test plan

Tested by running
```
env COLUMNS=2 pipx install pycowsay
```
For current pipx this crashes with a `ValueError`.  For pipx with this PR patch it executes successfully.


itsayellow:
I think I branched from the wrong starting branch!  Let me shut down this PR and start again.

PR #453, opened 2020-07-26
Disable animation if terminal width is reported as 0 or small

itsayellow:

## Summary of changes
Fixes bug #444 .

Some terminal / interactive programs and mixtures cause the terminal columns to be reported by `shutil.get_terminal_size` as 0.  I was able to find some examples online talking about the interactive shell inside of emacs doing this sometimes.  

In general, if the terminal columns are smaller than our continuation characters, the animate thread will crash with a python stack trace, and this is unacceptable.

This PR prevents the thread crash from happening.

## Test plan

Tested by running
```
env COLUMNS=2 pipx install pycowsay
```
For current pipx the animate thread crashes with a `ValueError` and a stacktrace.  For pipx with this PR patch it executes successfully.

### DIFF
--- a/src/pipx/animate.py
+++ b/src/pipx/animate.py
@@ -69,17 +69,22 @@ def print_animation(
     period: float,
     animate_at_beginning_of_line: bool,
 ):
-    (term_cols, _) = shutil.get_terminal_size(fallback=(9999, 24))
+    default_term_cols = 9999
+    (term_cols, _) = shutil.get_terminal_size(fallback=(default_term_cols, 24))
     event.wait(delay)
     while not event.wait(0):
         for s in symbols:
             if animate_at_beginning_of_line:
                 max_message_len = term_cols - len(f"{s} ... ")
+                if max_message_len <= 0:
+                    max_message_len = default_term_cols - len(f"{s} ... ")
                 cur_line = f"{s} {message:.{max_message_len}}"
                 if len(message) > max_message_len:
                     cur_line += "..."
             else:
                 max_message_len = term_cols - len("... ")
+                if max_message_len <= 0:
+                    max_message_len = default_term_cols - len("... ")
                 cur_line = f"{message:.{max_message_len}}{s}"
 
             clear_line()
--- a/src/pipx/animate.py
+++ b/src/pipx/animate.py
@@ -16,11 +16,12 @@ EMOJI_ANIMATION_FRAMES = ["⣷", "⣯", "⣟", "⡿", "⢿", "⣻", "⣽", "⣾"
 NONEMOJI_ANIMATION_FRAMES = ["", ".", "..", "..."]
 EMOJI_FRAME_PERIOD = 0.1
 NONEMOJI_FRAME_PERIOD = 1
+MINIMUM_COLS_ALLOW_ANIMATION = 16
 
 
 def _env_supports_animation():
     (term_cols, _) = shutil.get_terminal_size(fallback=(0, 0))
-    return stderr_is_tty and term_cols > 16
+    return stderr_is_tty and term_cols > MINIMUM_COLS_ALLOW_ANIMATION
 
 
 @contextmanager
--- a/src/pipx/animate.py
+++ b/src/pipx/animate.py
@@ -29,7 +29,8 @@ def animate(
 ) -> Generator[None, None, None]:
 
     if not do_animation or not _env_supports_animation():
-        # no op
+        # No animation, just a single print of message
+        print(f"{message}...")
         yield
         return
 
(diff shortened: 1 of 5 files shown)


itsayellow:
Is this worth a separate test in `test_animate.py`?

src/pipx/animate.py
```diff
@@ -69,17 +69,22 @@ def print_animation(
...
                 max_message_len = term_cols - len(f"{s} ... ")
+                if max_message_len <= 0:
+                    max_message_len = default_term_cols - len(f"{s} ... ")
                 cur_line = f"{s} {message:.{max_message_len}}"
                 if len(message) > max_message_len:
                     cur_line += "..."
             else:
                 max_message_len = term_cols - len("... ")
+                if max_message_len <= 0:
+                    max_message_len = default_term_cols - len("... ")
```
uranusjr, line 87: Similar to above.

src/pipx/animate.py
```diff
@@ -69,17 +69,22 @@ def print_animation(
...
-    (term_cols, _) = shutil.get_terminal_size(fallback=(9999, 24))
+    default_term_cols = 9999
+    (term_cols, _) = shutil.get_terminal_size(fallback=(default_term_cols, 24))
     event.wait(delay)
     while not event.wait(0):
         for s in symbols:
             if animate_at_beginning_of_line:
                 max_message_len = term_cols - len(f"{s} ... ")
+                if max_message_len <= 0:
+                    max_message_len = default_term_cols - len(f"{s} ... ")
```
uranusjr, line 80: I would just do something like

```python
...
if max_message_len > 0:
    cur_line = f"{s} {message:.{max_message_len}}"
    if len(message) > max_message_len:
        cur_line += "..."
else:
    cur_line = f"{s} {message}"
```

since the fallback value is an arbitrary value anyway.

itsayellow: In my mind I was wavering on whether the fallback `default_term_cols` should be effectively infinity (the way it is now) or a reasonable guess like 80 or 40.  So I wrote the code so that it would still work if we decided that `default_term_cols` should be some smaller value.

In general I couldn't think get this part of the code to look non-awkward to me, so if we want to enshrine the infinitely-long default line, I can change it to your suggestion.

itsayellow: If we cannot determine a reasonable terminal width, do we want to guess a number like 80?  Or should we not truncate the message at all (current behavior)?

uranusjr: Personally I would choose to not truncate, and let the terminal software (i.e. emulator these days) do what it wants. This generally results in either truncated or wrapped text, both are fine IMO.

itsayellow: The wrapped text is the problem with animated lines, before we truncated, if the terminal was too narrow, the message would spill over to the next physical line.  Then the "delete line" terminal character would only go to the beginning of the last physical line wrapped extra bit, and not to the beginning of the entire multi-line.  This would make the animate thread spew animate frame lines down the terminal instead of rewriting the same line over and over as it was meant to.

itsayellow: This conversation brings me back to an earlier thought.  I think we should just disable animation if we cannot determine the terminal width, or if it's uselessly-small (like < 10).  In such cases I would guess we are in some primitive situation anyway, or piping to something, etc.

@cs01, @uranusjr what do you think of this?

cs01: > This conversation brings me back to an earlier thought. I think we should just disable animation if we cannot determine the terminal width, or if it's uselessly-small (like < 10). In such cases I would guess we are in some primitive situation anyway, or piping to something, etc.
> 
> @cs01, @uranusjr what do you think of this?

Yes I like this. I would favor keeping the code more simple+maintainable if the cost is occasionally not animating. Animating is really a "nice to have" kind of thing anyway. I don't think anyone will get too upset if apps are installed successfully but the animation didn't display.

itsayellow:
I added a print statement so that even if we do not do animation, we still print `message` with an ellipsis tacked on once to stdout to indicate something is happening.  (e.g. "creating virtual environment...", "installing pycowsay...")

Let me know if people like that idea.

itsayellow:
I also modified `test_shared_libs.py` because one of the macos tests took 22 minutes and broke the test.  The problem is that `now` is computed at import, and if the test for "not time to update shared libraries" is set only 20 minutes before the deadline, by the end of the test it unfortunately becomes "time to update shared libraries".

To fix the wide variances of time between import and test, I moved the setting of `now` inside the test function.

itsayellow:
Additionally I revamped the animation tests to use parametrization instead of for-loops.  This is easier to read and gives better test isolation.  It eliminates past problems with stderr from one sub-test affecting the next sub-test.

I also added a test to verify that `stderr_is_tty=False` or any term_cols less than 17 will disable animation.

Issue #338, opened 2020-01-17
create snap (linux) and scoop (windows) packages


AlJohri:
**How would this feature be useful?**
We have homebrew (which is slowly getting better) on the macOS side of things. I think adding a snap package would cover almost all the major [linux distributions](https://snapcraft.io/docs/installing-snapd) these days. https://snapcraft.io/

Additionally a `scoop` package should make it much easier to install pipx on Windows: https://scoop.sh/

**Describe the solution you'd like**
Easy system installations of pipx:
```
brew install pipx
snap install pipx
scoop install pipx
```

**Describe alternatives you've considered**
As you obviously know from creating `pipx`, `pip` is not an ideal installation method for system wide utilities. I think creating system packages for pipx on all three major platforms will help this tool grow.

ohk2kt3t4:
Regarding scoop, have you tried submitting a PR to add pipx to scoop main or the extras bucket?

uranusjr:
I mentioned in #244 I created a Scoop manifest, [available here](https://github.com/uranusjr/pipx-standalone), but there are still disagreements how to do this best (see discussion in the linked issue). It’s probably a better idea to delay submitting this to the main bucket until there’s an agreement, and the Scoop manifest can be incorporated into the main pipx repository. Feel free to try that out in the mean time though. I’ve been using it for a while now and have not noticed any glitches.

aaronsteers:
I'm not familiar with scoop but we use chocolatey [a lot](https://docs.dataops.tk/setup/windows) for windows installs. Are there advantages to scoop over chocolatey or are the pretty similar?

I often tell data analysts they need to install python just to get the pip package manager. It would be nice to either have a pipx standalone (bundled with python or only installs python if it is missing). Even if it still requires python installed, we could simply tell people:

```cmd
choco install python3 # skip if you have it already
choco install pipx
```

cs01:
I have no problem with pipx being on chocolatey and I doubt any other maintainers do. That said, pipx is maintained by unpaid volunteers in our spare time, so the limiting factor in getting pipx on all these package managers is the amount of time and energy the maintainers have to dedicate to that process. @AlJohri stepped up and got the ball rolling with pipx on brew. If you are interested in owning pipx on chocolatey, that would be welcome. If not, maybe one of the other maintainers would like to. I am personally not able to.

uranusjr:
[...]

Issue #244, opened 2019-10-18
pipx as a frozen standalone binary


clbarnes:
pipx solves the problem of system tools depending on python installations, of which you may have one (a problem because messing with the system python is dicey), or many (a problem because you have to keep version-gardening).

Even if you're disciplined and keep as much as possible in virtualenvs, you still end up having to install one system tool outside of a virtual environment... pipx (the call is coming from inside the house!). If it would be welcome, I'm considering trying to freeze pipx into a static binary using PyOxidizer. It would then be pretty easy to deploy on different package managers etc. and generally be a "real" system tool, isolated from the system and other un-virtualenv'd pythons. Once pipx is installed, other python-based system tools are a breeze to install.

However, I suspect there might be some interaction between PyOxidizer and the bits of pipx which complain about being used in a virtualenv. I'll see what happens.

clbarnes:
The way that the `venv_metadata_inspector` is currently read falls afoul of the resource loader issue [documented here](https://pyoxidizer.readthedocs.io/en/latest/packaging_pitfalls.html). In python 3.7+, this can be resolved with the ResourceLoader API.

cs01:
> the call is coming from inside the house!

:laughing: 

pipsi solved this by bootstrapping itself, and pipx used to install itself this way. I didn't really see a big benefit from doing this though so I stopped doing it with pipx because it added a bunch of extra confusion to people trying to install it for the first time. At that time pipx had I think zero dependencies and now it has a couple plus several transient dependencies. @mattsb42  made a clever package on PyPI called pipipxx which does this (from its README):

>  builds a temporary virtual environment, installs pipx there, and then uses that pipx to install pipx in your user local space, just like any other pipx-installed tool. What you end up with is a pipx installation that is itself managed by pipx.

Anyway, back to PyOxidizer:

> The way that the venv_metadata_inspector is currently read falls afoul of the resource loader issue documented here. In python 3.7+, this can be resolved with the ResourceLoader API.

This isn't a problem is it? PyOxidizer bundles the entire Python runtime, so if the PyOxidizer version of pipx is built with 3.7, the resource loader api will be available and work just fine.

cs01:
Oh btw this was discussed a little in another issue. https://github.com/pipxproject/pipx/issues/146#issuecomment-532560587

cs01:
[...]

PR #465, opened 2020-08-13
Support running with the Windows embeddable Python distribution

pfmoore:

## Summary of changes

Adds support for using pipx with the Windows embeddable distribution, which doesn't include a `venv` module itself. The change makes pipx use the embeddable distribution for running its own code, but uses the system default Python for creating virtual environments.

The code is not specific to the embeddable distribution, but is triggered if the `venv` module is not importable, which may be useful for other limited Python distributions.

## Test plan

There is a version of pipx installable using the [scoop](https://scoop.sh/) package manager which uses this patch. I have been using this distribution for some time now.

### DIFF
--- a/src/pipx/constants.py
+++ b/src/pipx/constants.py
@@ -3,7 +3,6 @@ import sys
 from pathlib import Path
 from textwrap import dedent
 
-DEFAULT_PYTHON = sys.executable
 DEFAULT_PIPX_HOME = Path.home() / ".local/pipx"
 DEFAULT_PIPX_BIN_DIR = Path.home() / ".local/bin"
 PIPX_HOME = Path(os.environ.get("PIPX_HOME", DEFAULT_PIPX_HOME)).resolve()
--- a/src/pipx/interpreter.py
+++ b/src/pipx/interpreter.py
@@ -15,14 +15,14 @@ def has_venv() -> bool:
         return False
 
 
+# The following code was copied from https://github.com/uranusjr/pipx-standalone
+# which uses the same technique to build a completely standalone pipx
+# distribution.
+#
 # If we are running under the Windows embeddable distribution,
 # venv isn't available (and we probably don't want to use the
 # embeddable distribution as our applications' base Python anyway)
 # so we try to locate the system Python and use that instead.
-
-# This code was copied from https://github.com/uranusjr/pipx-standalone
-# which uses this technique to build a completely standalone pipx
-# distribution
 
 
 def _find_default_windows_python() -> str:
@@ -37,8 +37,10 @@ def _find_default_windows_python() -> str:
     if python is None:
         raise PipxError("no suitable Python found")
 
+    # If the path contains "WindowsApps", it's the store python
     if "WindowsApps" not in python:
         return python
+
     # Special treatment to detect Windows Store stub.
     # https://twitter.com/zooba/status/1212454929379581952
 
--- a/src/pipx/interpreter.py
+++ b/src/pipx/interpreter.py
@@ -1,0 +1,60 @@
+import shutil
+import subprocess
+import sys
+
+from pipx.constants import WINDOWS
+from pipx.util import PipxError
+
+
+def has_venv() -> bool:
+    try:
+        import venv  # noqa
+
+        return True
+    except ImportError:
+        return False
+
+
+# If we are running under the Windows embeddable distribution,
+# venv isn't available (and we probably don't want to use the
+# embeddable distribution as our applications' base Python anyway)
+# so we try to locate the system Python and use that instead.
+
+# This code was copied from https://github.com/uranusjr/pipx-standalone
+# which uses this technique to build a completely standalone pipx
+# distribution
+
+
+def _find_default_windows_python() -> str:
+
+    if has_venv():
+        return sys.executable
+
+    py = shutil.which("py")
+    if py:
+        return py
+    python = shutil.which("python")
+    if python is None:
+        raise PipxError("no suitable Python found")
+
+    if "WindowsApps" not in python:
+        return python
+    # Special treatment to detect Windows Store stub.
+    # https://twitter.com/zooba/status/1212454929379581952
+
+    proc = subprocess.run(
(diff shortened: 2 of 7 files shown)


src/pipx/constants.py
```diff
@@ -30,6 +30,48 @@ def is_windows() -> bool:
...
+# venv isn't available (and we probably don't want to use the
+# embeddable distribution as our applications' base Python anyway)
+# so we try to locate the system Python and use that instead.
+
+# This code was copied from https://github.com/uranusjr/pipx-standalone
+# which uses this technique to build a completely standalone pipx
+# distribution
+
+
+def _find_default_python():
```
cs01, line 43: If this is only used on windows, suggest changing the name to contain `windows`, something like `_find_default_python_windows()`.

src/pipx/constants.py
```diff
@@ -30,6 +30,48 @@ def is_windows() -> bool:
...
+    # Special treatment to detect Windows Store stub.
+    # https://twitter.com/zooba/status/1212454929379581952
+    import subprocess
+
+    proc = subprocess.run(
+        [python, "-V"], stdout=subprocess.PIPE, stderr=subprocess.DEVNULL,
+    )
+    if proc.returncode != 0:
+        # Cover the 9009 return code pre-emptively.
+        raise EnvironmentError("no available Python found")
```
cs01, line 61: Though `EnvironmentError` is technically accurate, pipx only catches `PipxError`s (at the bottom of `main.py`) and prints a single line, user-facing error. This will result in a full stack trace, which we want to avoid. Could these be changed to `PipxError`s?

src/pipx/constants.py
pfmoore: No problem, done.

src/pipx/constants.py
pfmoore: Also done

pfmoore: Hmm, this failed because of a circular import - constants imports util which in turn imports constants. I'll need to look in more detail at how to address this. I'll come back with a revised patch in a while...

cs01:
> There is a version of pipx installable using the scoop package manager which uses this patch. I have been using this distribution for some time now.

Perhaps documentation for this should be added to pipx?

pfmoore:
> Perhaps documentation for this should be added to pipx?

Perhaps. It's not my project, though, so I'd be reluctant to give it more publicity without the owner (@uranusjr) confirming he was OK with that.

src/pipx/constants.py
pfmoore: Actually, after looking for a while for ways to avoid the circular import it occurred to me that there's no point. If we can't find a suitable value for `DEFAULT_PYTHON`, that will happen when constants.py is *imported*, which is outside the scope of the try..except that catches PipxError.

I see two possible fixes here:

1. (Simple, but a bit of a hack) Replace the exception with a simple `raise SystemExit("Python interpreter does not support venv")`. This will avoid an unfriendly traceback, but will remain outside of the general `PipxError` handling.
2. (More complex, maybe better?) Replace DEFAULT_PYTHON with a function, called when needed. I'd propose splitting this out into a separate file, `pipx.interpreter`, to break the circular import.

With option (2) it would also be possible to add some simple unit tests (monkeypatch `import venv` to fail, and check the default python is not `sys.executable`). But we'd be adding some non-trivial complexity for a fairly niche case.

What would you prefer?

uranusjr:
I’ve mentioned the project in both #244 and #338. Neither message got any feedback, and I assumed the interest has been minimal.

It’s perfectly fine if it’s added to the documentation, but I’d caution that the maintenance has been low-effort (I regularly skip pipx updates I don’t care to follow), and anyone wanting to use it should be ready to contribute 🙂

pfmoore:
(@uranusjr Offtopic here, but I've got some scripts that do auto-updates of a scoop bucket - I'm pretty sure I could set up an auto-update process that tracked new pipx releases. It's a bit fiddly, though, so I can't promise how soon I'd be able to get to it)

src/pipx/constants.py
pfmoore: @cs01 Gentle ping - which approach do you thing I should take with this?

cs01: Hi sorry for the slow response. 

I lean toward option 2. This will probably not be the last time we need to add additional logic to determining DEFAULT_PYTHON.

pfmoore: No worries, I'll see if I can implement that approach. I just didn't want to over-engineer for what is a relatively rare case, without knowing whether you supported it in principle.

pfmoore:
Sorry for the delay. New version uploaded. I had to rebase and force-push as I got in a mess merging with master (black changed its behaviour and I did a whole lot of fixups before I merged and found master handled it differently...)

Anyway, the resulting change seems cleaner as a single commit, so hopefully no harm done.

src/pipx/interpreter.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+# If we are running under the Windows embeddable distribution,
+# venv isn't available (and we probably don't want to use the
+# embeddable distribution as our applications' base Python anyway)
+# so we try to locate the system Python and use that instead.
+
+# This code was copied from https://github.com/uranusjr/pipx-standalone
+# which uses this technique to build a completely standalone pipx
+# distribution
```
uranusjr, line 25: Do these comment blocks belong inside the function below instead?

pfmoore: It was intentional to put the comments before the function rather than inside it (I find it more readable that way). But I can reword it a bit, I guess. Maybe swap the paragraphs, and say "The following code was copied from ... which uses the same technique..."?

uranusjr: Yeah, the paragraphs seem more natural to me if swapped.

pfmoore: Done.

cs01 (approved):

cs01:
Thank you, @pfmoore

pfmoore:
Awesome, thanks!

PR #467, opened 2020-08-19
Add pre-commit support

itsayellow:

## Summary of changes
I've been using this pre-commit with the config file in this PR for a while now and really like it.  I'd like to make it part of the repo.

If you don't install pre-commit hooks into your git repo, this file will have no effect on you.

If you do install pre-commit and its hooks, this config will do our lint tests (and more) before each commit:
* Run isort
* Run black
* Run flake8
* Run mypy

If it finds problems it will abort the commit.  In the case of black and isort it will additionally fix the problems by running those utilities to modify the code!

I really like being notified, and/or fixing the problems **before** I commit, instead of being notified after the fact by our Lint test.

The only downside I can think of is for this PR is that it adds another dot-file to the repo.

## Test plan

Tested by running
```
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,28 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/#installation for installation instructions
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v3.2.0
+  hooks:
+  - id: end-of-file-fixer
+  - id: check-added-large-files
+  #- id: trailing-whitespace
+  #- id: check-yaml
+- repo: https://github.com/timothycrosley/isort
+  rev: 5.4.2
+  hooks:
+  - id: isort
+    args: ['--profile','black']
+- repo: https://github.com/psf/black
+  rev: 19.10b0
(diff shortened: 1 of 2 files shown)


gaborbernat (approved):

cs01:
I have never taken the time to try pre-commit. I will give it a go. We should probably mention it in the docs for contributing.

cs01:
We should add a nox session to run the equivalent of what pre-commit does, otherwise users will get errors for things like isort but have no way of fixing it with pipx's tooling.

itsayellow:
> We should add a nox session to run the equivalent of what pre-commit does, otherwise users will get errors for things like isort but have no way of fixing it with pipx's tooling.

Yes, this is currently a slight incongruity between the two.

However, the pre-commit use of isort actually just does the "isorting" itself, so it's not really a roadblock, more of an automatic fixer like black.

Currently pre-commit users will be now submitting "better" code than our online linters check for.

itsayellow:
To clarify:  the pre-commit hooks like `black` and `isort` actually are run not as checkers but as modifiers.  

The first time you try and commit, if there are things those tools need to do, it prevents the commit but modifies the code and adds those changes.  Then if you commit again it will successfully commit both your code edits and the changes made by the pre-cmmit hooks.

Issue #257, opened 2019-10-27
add isort to lint tests


cs01:
**How would this feature be useful?**

Ensure imports have been autosorted will make code consistent/readable and save time for developers.

**Describe the solution you'd like**

Use `isort` and https://github.com/asottile/seed-isort-config in the nox `lint` session. Also add a new nox session `format` to automatically run black, isort, and anything else we add.

Something like
```
    session.run("seed-isort-config", "--application-directories=pipx")
    session.run("isort", "--project=pipx", "--recursive", "--apply", *source_files)
```

**Describe alternatives you've considered**

Manually run isort, which is what I've been currently doing.

gaborbernat:
Can we just call pre-commit instead; with black and all that, see e.g. https://github.com/tox-dev/tox/blob/master/.pre-commit-config.yaml

itsayellow:
I started running pre-commit on one computer a while ago and am liking it.  Among other things it's nice for easily fixable things like black.  It's also nice that we can have a standard `.pre-commit-config.yaml` for the pipx project if we wish.

itsayellow:
I have a mockup of two approaches for adding isort to lint tests.  I prefer the one that shows the actual diff of the necessary changes to fix a file when that file has badly-sorted imports.  But broadly speaking there are two options:

`isort --check --diff` actually shows diffs for failing files.  It's verbose but even if you aren't running isort yourself you could fix the file manually from the provided diff information.
Its output looks like this:
```
nox > isort --check --diff --profile black src/pipx tests noxfile.py get-pipx.py setup.py generate_docs.py
ERROR: /home/runner/work/pipx/pipx/src/pipx/animate.py Imports are incorrectly sorted and/or formatted.
ERROR: /home/runner/work/pipx/pipx/src/pipx/shared_libs.py Imports are incorrectly sorted and/or formatted.
--- /home/runner/work/pipx/pipx/src/pipx/animate.py:before	2020-09-20 04:11:42.338241
+++ /home/runner/work/pipx/pipx/src/pipx/animate.py:after	2020-09-20 04:11:52.216611
@@ -1,8 +1,8 @@
+import shutil
 import sys
 from contextlib import contextmanager
 from threading import Event, Thread
 from typing import Generator, List
-import shutil
 
 from pipx.constants import emoji_support
 
--- /home/runner/work/pipx/pipx/src/pipx/shared_libs.py:before	2020-09-20 04:11:42.338241
[...]

PR #473, opened 2020-08-23
Add configuration for isort

itsayellow:

## Summary of changes
Adds config for isort
* uses `profile=black` for black-compatible formatting
* adds `helpers` (e.g. in `tests/test_install.py`) to known_first_party imports so isort places it in the right block of imports

This is one step to making #257 work.

## Test plan

Tested by running
```
```

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,0 +1,4 @@
+# can be merged into pyproject.toml at a later date
+[settings]
+profile=black
+known_first_party=helpers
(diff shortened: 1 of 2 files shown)


gaborbernat (approved):

cs01:
I tried getting isort to work once and gave up on it. It would re-sort packages in different order depending on a bunch of different things, and I eventually decided it wasn't worth spinning my wheels on it. I'm glad you got it working though.

PR #476, opened 2020-08-24
Assert on python version to determine if venv is available

jonringer:

## Test plan

Tested by running
```
```


jonringer:
actually, now that I think about it, you require python 3.6+ for your package, so this should always be true

jonringer:
closing in favor of  #477

PR #477, opened 2020-08-24
Remove venv check

jonringer:
## Summary of changes
venv is available through the standard library since 3.3, since you require 3.6+ for your package, it will always be true, so just remove assertion.

https://www.python.org/dev/peps/pep-0405/

## Test plan

Tested by running
```
```

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -9,7 +9,6 @@ from pipx import main
 
 def assert_not_in_virtualenv():
     assert not hasattr(sys, "real_prefix"), "Tests cannot run under virtualenv"
-    assert getattr(sys, "base_prefix", sys.prefix) != sys.prefix, "Tests require venv"
 
 
 def run_pipx_cli(pipx_args: List[str]):


FRidh:
If the point is to detect whether we are not in a `virtualenv` environment *and* require we are in a `venv`, then the helper seems correct to me. If the helper does not need to run inside a `venv`, then it can be removed. No idea why it would require being tested inside a `venv`.

The second line tests whether `base_prefix` is unequal to `prefix`. This is the case when you are inside a `venv` environment. To check whether `virtualenv` is used instead of `venv` the first line is used.

In case `base_prefix` is used it falls back to `prefix`, which of course always fails. Thus, the test only functions in Python 3.3 and above (because that's when `base_prefix` was introduced).

https://virtualenv.pypa.io/en/16.7.9/reference.html#compatibility-with-the-stdlib-venv-module

uranusjr:
Since modern virtualenv implements PEP 405 and is as capable as the built-in venv, maybe the two checks should be reversed to allow tests to be run inside it as well?

gaborbernat (changes requested):
Yeah, I agree with @uranusjr here.

jonringer:
In the linked PR, you can see that we don't use virutalenv and we remove the assertion that this PR is all about. We still had to disable tests which touched network. But on the whole, the test suite doesn't need them.

gaborbernat:
The thing is the test suite can run under virtualenv 20+ though, not?

jonringer:
Probably? But that's not the point of this PR, which is to remove the inconclusive assertion on venv.

gaborbernat (approved):

PR #478, opened 2020-08-25
Fix regression with install of local path with extras

### LINKED ISSUE #475: pipx fails to install editable packages with extras in spec
**Describe the bug**

pipx 0.15.5.0 refuses to install editable packages with [extras](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) in spec

**How to reproduce**

```bash
pipx install --editable .[dev]
# Unable to parse package spec: .[dev]
```

**Expected behavior**

`pipx` installs local package and extras in editable mode.

itsayellow:

## Summary of changes
Fix regression with 0.15.5.0 which erroneously disallowed installation from local paths with package extras.

Fixes #475 

## Test plan

Tested by running
From pipx git clone main directory:
```
pipx '.[pdf]'
```
There is no `pdf` extra, but this PR code merely gives a warning that there is no `pdf` extra and continues to install pipx.

Existing pipx code gives instead the following (wrong) error:
```
Unable to parse package spec: .[pdf]
```

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,7 +1,7 @@
 graft src
 include *.md pipx_demo.gif logo.png LICENSE
 
-exclude .isort.cfg .pre-commit-config.yaml get-pipx.py makefile generate_docs.py .flake8 mkdocs.yml noxfile.py .coveragerc
+exclude testdata .isort.cfg .pre-commit-config.yaml get-pipx.py makefile generate_docs.py .flake8 mkdocs.yml noxfile.py .coveragerc
 prune templates
 prune tests
 prune docs
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,7 +1,8 @@
 graft src
 include *.md pipx_demo.gif logo.png LICENSE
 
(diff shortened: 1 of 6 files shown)


itsayellow:
I need to come up with some tests before I'm ready to merge this.

I'm thinking I'd like to make a very simple local package to run tests installing local paths.  Does anyone have suggestions where to put such a local test package?  Something like a subdirectory of `tests/testdata/`?  Or is there a better place/way?

itsayellow:
Right now I'm working with directory `testdata` in the top level, so as to not interfere with pytest looking for test code in `tests`.  But I'm open to suggestions on how to structure the directories.

gaborbernat (approved):
Can you fix the merge conflict otherwise looks good to me 👍

itsayellow:
I don't understand what's happening with `black` inside the Lint check.  It fails on github but not for me locally.  It's claiming that files need to be reformatted since the merge of PR #477 .

itsayellow:
Ah `black` just released a new version 4 hours ago.  It must be changing the formatting vs. the old version which is what I have stored locally.

I wonder if we need pin the version of `black` in the Lint test.

PR #480, opened 2020-08-26
Publish docs automatic workflow

### LINKED ISSUE #333: Deploy docs: GitHub Action triggered by release

It would be cool if new documentation could automatically run whenever we make a release.

Implementing this action might help us gain experience and confidence to implement and Action to automatically publish to PyPI in the future.

I've started investigating GitHub Actions, and this Issue is to track figuring out how to make an Action that would build the site using mkdocs and deploy, triggered by a Release.

The current way of generating and publishing the docs is fine, but it's another thing that needs to be remembered and executed properly each time.  Triggering automatically on release would remove one more thing to do by hand, hopefully ensuring it's done consistently and accurately.

itsayellow:

## Summary of changes
Closes #333 

This rebuilds and publishes our docs whenever there is a published release.  It can also manually be initiated whenever somebody activates this action from the "Actions" tab, Select "Publish docs via GitHub Pages" and use the pulldown "Run workflow".

**NOTE** the action used here does a rebase in the `gh-pages` branch, so the result is that there is only one commit on the `gh-pages` branch.  All history in the `gh-pages` branch is gone after it executes.

I've tested this in a mock pipx repo and it seems to work well, both with the release trigger and with the manual trigger.

## Test plan

Tested by running
```
```

### DIFF
--- a/.github/workflows/publish_docs.yml
+++ b/.github/workflows/publish_docs.yml
@@ -1,6 +1,6 @@
 name: Publish docs via GitHub Pages
 on:
-  workflow_dispatch: ''
+  workflow_dispatch:
   release:
     types: [published]
 
--- a/.github/workflows/publish_docs.yml
+++ b/.github/workflows/publish_docs.yml
@@ -1,0 +1,24 @@
+# https://docs.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#manually-running-a-workflow
+# https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/
+
+name: Publish Docs
+
+on:
+  workflow_dispatch:
+    inputs:
+      name:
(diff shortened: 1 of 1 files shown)


gaborbernat (approved):

cs01:
This is super cool. I have been wanting to look into this for pipx as well as my other projects. I will probably be copying and pasting this in the future. Thanks for automating more things!

`nox -s publish` already publishes docs. I suppose the next logical step is to make a publish github workflow, and not run the `nox -s publish` session locally?

itsayellow:
> `nox -s publish` already publishes docs. I suppose the next logical step is to make a publish github workflow, and not run the `nox -s publish` session locally?

Yes!  I was working my way up to that incrementally.  I wanted to try some low-stakes experiment(s) first before doing the entire release using Github Actions.

PR #481, opened 2020-08-26
Use persistent github cache for pip cache

itsayellow:

## Summary of changes
This uses the persistent github workflow cache to preserve the pip cache between test workflows.  It speeds up operation by not needing to rebuild or download wheels once they are already in the pip cache.

It's hard to quantify how much time this saves because the normal variation in test times for any given workflow test is already large.  It looks like from my testing (sample size N=6) that the biggest help is with the macOS target.  The macOS "build regex wheel" step can take anywhere from 25 seconds to 2 minutes normally, and with this PR takes 4 seconds reliably.  

Second in impact is the Windows target, which looks like it's saving something like 1 minute in total.

In addition, the normal pipx tests that require many installs of big python packages seem to save something like 30 seconds over the whole test for the remaining OS tests.

One remaining question: all of the internet examples for generating a cache key for github workflow caches involve hashing `requirements.txt`.  Since we don't really have that, and the cache is supporting many different python packages, I dispensed with hashing any file altogether and just made a persistent cache key only based on the OS and python version.  (e.g. a cache key would be `Linux-pip3.8`).  I assume this is ok, as pip will take care of expiring its own cache items.  But I'm happy to be educated otherwise if I'm wrong. 

Another minor change I did was to use `runner.os` instead of `matrix.os` to check for macOS so that we could use an OS-version-independent string.

Refs:
https://docs.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows
https://github.com/actions/cache/blob/master/examples.md#python---pip

## Test plan

Tested by running
```
```

### DIFF
--- a/.github/workflows/on-push.yml
+++ b/.github/workflows/on-push.yml
@@ -66,8 +66,6 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python-version }}
-    - name: Upgrade pip
-      run: python -m pip install --upgrade pip
     - name: Get pip cache dir
       id: pip-cache
       run: |
@@ -77,6 +75,8 @@ jobs:
       with:
         path: ${{ steps.pip-cache.outputs.dir }}
         key: ${{ runner.os }}-pip${{ matrix.python-version }}
+    - name: Upgrade pip
+      run: python -m pip install --upgrade pip
     # macOS: Install wheel to allow wheel-building for nox and regex
     # Build regex wheel into pip cache, so future pipx installs don't compile.
     # This avoids future compile errors in nox tests because of missing
--- a/.github/workflows/on-push.yml
+++ b/.github/workflows/on-push.yml
@@ -68,6 +68,15 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Upgrade pip
       run: python -m pip install --upgrade pip
+    - name: Get pip cache dir
+      id: pip-cache
+      run: |
+        echo "::set-output name=dir::$(pip cache dir)"
+    - name: Persistent pip cache
+      uses: actions/cache@v2
+      with:
+        path: ${{ steps.pip-cache.outputs.dir }}
+        key: ${{ runner.os }}-pip
     # macos: Install wheel to allow wheel-building for nox and regex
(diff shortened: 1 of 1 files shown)


gaborbernat (approved):

cs01:
> The macOS "build regex wheel" step can take anywhere from 25 seconds to 2 minutes normally, and with this PR takes 4 seconds reliably.

Nice!!

> , I dispensed with hashing any file altogether and just made a persistent cache key only based on the OS and python version. (e.g. a cache key would be Linux-pip3.8). I assume this is ok, as pip will take care of expiring its own cache items. But I'm happy to be educated otherwise if I'm wrong.

I have never heard of this, so I'm learning by reading your PR. Do you know what cases this would result in pipx/pip using the wrong package from the cache?

itsayellow:
> > , I dispensed with hashing any file altogether and just made a persistent cache key only based on the OS and python version. (e.g. a cache key would be Linux-pip3.8). I assume this is ok, as pip will take care of expiring its own cache items. But I'm happy to be educated otherwise if I'm wrong.
> 
> I have never heard of this, so I'm learning by reading your PR. Do you know what cases this would result in pipx/pip using the wrong package from the cache?

No, nothing as bad as using the wrong item from the cache.

What I was referring to was how long cached items remain after they haven't been used in a while.  pip has an expiration procedure for removing unused items from its cache.  (Hard to tell what that is, but it is mentioned.)

I think Github's examples for python pip caches tend to put the hash of `requirements.txt` into the cache key.  If you do this, you will abandon your previous pip cache when you change your project's requirements, and thus I think the idea is that Github will easily detect a cache that hasn't been touched in a while and know to delete it.  

That didn't seem amenable to our case (no `requirements.txt` and lots of different installed projects from the tests) so I didn't hash anything to form the cache key, and as a result we will keep using the same Github cache key over and over and just rely on pip to delete unused items from its own cache.

I suppose the risk is that if pip doesn't delete some old cached packages for some reason, we could end up with a cache leak and it could start to grow.

itsayellow:
What just occurred to me now, is that if we wanted to be extra cautious, we could append `<year><month>` to the cache key and we would start fresh with a new cache every month regardless.

Issue #79, opened 2019-01-31
Update injected packages.


schinckel:
As mentioned in #72, it's currently not possible to update dependencies/injected packages in pipx.

Should this be supported in some form? What form should this take?

`pipx update <foo>`

Should this update all injected packages and dependencies, or not?

pmav99:
Blindly updating dependencies can be problematic. You need to know that the main package is compatible with the newer version. If strict semver could be assumed, then maybe updating of minor versions and patches could be justified, but it cannot.

In the past `pip` was using `--upgrade-strategy=eager` which was especially problematic. Since version 10 (released last Summer?), I think that the default changed to `--upgrade-strategy=only-if-needed` which is a much safer default. 

That being said, i have not really tested it. Hopefully, in a not so far away future, every python package will be using pipenv/poetry to define their dependencies and pipx will be able to install the specific set of dependencies that the main package's developers chose/approved.

Until then, if `--upgrade-strategy=only-if-needed` is working OKish, then I guess it is not a problem to implement this, but if there are problems, I think it is just simpler for pipx to do nothing and let users manually solve any problems when they appear  (it is just a virtualenv after all) and/or `pipx uninstall foo` followed by `pipx install foo`.

taketwo:
> Hopefully, in a not so far away future, every python package will be using pipenv/poetry to define their dependencies and pipx will be able to install the specific set of dependencies that the main package's developers chose/approved.

I'd like to point out that the injected packages are not necessarily dependencies of the "main" package. (Consider the example from README, we inject `requests` into `ptpython` REPL.) Thus in general case we don't (and never will) have a set of "officially approved" versions to match.

In my opinion, if we `install`ed the main package and then `inject`ed more packages into it _without_ explicitly specifying versions, then we are in the "use latest and greatest" mode. It's therefore logical to eagerly upgrade everything. Things might break, yes, but then again they might have been broken from the start if unversioned packages are injected.

cs01:
Agreed with @taketwo that the whole point of injecting dependencies is to add a new package that is NOT depended on by the package that was installed by pipx (the requests/ptpython example is apt here).

This introduces a challenge to pipx to basically inspect `pip freeze` and determine which packages were dependencies from the main package and which were injected. I don't know of a canonical way to do this. It's possible to hack something together to do this (for example could create a different temp virtual env, install there, inspect output of `pip freeze` to get set of dependent packages, but this is fraught with corner cases). 

[...]

PR #491, opened 2020-08-31
Improve support for suffixed packages

abn:

## Summary of changes
This change allows for safe listing and upgrade of packages installed with a suffix. As part of this change, a new metadata property "suffix" has been introduced.

## Test plan

Tested by running
```sh
pipx install --suffix=_test pycowsay
pipx upgrade pycowsay_test
pipx inject pycowsay_test pip
pipx list
pipx uninstall pycowsay_test
```

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,6 +1,7 @@
 dev
 
 - [docs] Update license
+- [bugfix] Fixed regression in list, inject and upgrade commands when suffixed packages are used.
 
 
 0.15.5.1
--- a/src/pipx/commands/common.py
+++ b/src/pipx/commands/common.py
@@ -86,7 +86,8 @@ def _symlink_package_apps(
 ) -> None:
     for app_path in app_paths:
         app_name = app_path.name
-        symlink_path = Path(local_bin_dir / add_suffix(app_name, suffix))
+        app_name_suffixed = add_suffix(app_name, suffix)
+        symlink_path = Path(local_bin_dir / app_name_suffixed)
         if not symlink_path.parent.is_dir():
             mkdir(symlink_path.parent)
 
@@ -117,12 +118,12 @@ def _symlink_package_apps(
             )
             symlink_path.unlink()
 
-        existing_executable_on_path = which(app_name)
+        existing_executable_on_path = which(app_name_suffixed)
         symlink_path.symlink_to(app_path)
 
         if existing_executable_on_path:
             logging.warning(
-                f"{hazard}  Note: {app_name} was already on your PATH at "
+                f"{hazard}  Note: {app_name_suffixed} was already on your PATH at "
                 f"{existing_executable_on_path}"
             )
 
@@ -133,12 +134,13 @@ def get_package_summary(
     package: str = None,
     new_install: bool = False,
     include_injected: bool = False,
-    suffix: str = "",
 ) -> str:
     venv = Venv(path)
     python_path = venv.python_path.resolve()
-    if package is None:
-        package = path.name
+
+    package_metadata = venv.get_package_metadata(package)
+    package = package_metadata.package or venv.root.name
+
     if not python_path.is_file():
         return f"   package {red(bold(package))} has invalid interpreter {str(python_path)}"
     if not venv.package_metadata:
@@ -147,8 +149,6 @@ def get_package_summary(
             f"       It was likely installed using a pipx version before 0.15.0.0.\n"
             f"       Please uninstall and install this package, or reinstall-all to fix."
         )
-
-    package_metadata = venv.package_metadata[package]
 
     if package_metadata.package_version is None:
         not_installed = red("is not installed")
@@ -160,7 +160,7 @@ def get_package_summary(
     )
     exposed_binary_names = sorted(p.name for p in exposed_app_paths)
     unavailable_binary_names = sorted(
-        {add_suffix(name, suffix) for name in package_metadata.apps}
+        {add_suffix(name, package_metadata.suffix) for name in package_metadata.apps}
         - set(exposed_binary_names)
     )
     # The following is to satisfy mypy that python_version is str and not
@@ -179,6 +179,7 @@ def get_package_summary(
         exposed_binary_names,
         unavailable_binary_names,
         venv.pipx_metadata.injected_packages if include_injected else None,
+        suffix=package_metadata.suffix,
     )
 
 
@@ -215,10 +216,13 @@ def _get_list_output(
     exposed_binary_names: List[str],
     unavailable_binary_names: List[str],
     injected_packages: Optional[Dict[str, PackageInfo]] = None,
+    suffix: str = "",
 ) -> str:
     output = []
+    suffix = f" ({bold(shlex.quote(package + suffix))})" if suffix else ""
     output.append(
-        f"  {'installed' if new_install else ''} package {bold(shlex.quote(package))} {bold(package_version)}, {python_version}"
+        f"  {'installed' if new_install else ''} package {bold(shlex.quote(package))}"
+        f" {bold(package_version)}{suffix}, {python_version}"
     )
 
     if not python_path.exists():
@@ -276,9 +280,10 @@ def run_post_install_actions(
     include_dependencies: bool,
     *,
     force: bool,
-    suffix: Optional[str] = None,
 ):
-    package_metadata = venv.package_metadata[package]
+    package_metadata = venv.get_package_metadata(package)
+    package = package_metadata.package or package
+    display_name = f"{package}{package_metadata.suffix}"
 
     if not package_metadata.app_paths and not include_dependencies:
         # No apps associated with this package and we aren't including dependencies.
@@ -295,7 +300,7 @@ def run_post_install_actions(
 
         if package_metadata.app_paths_of_dependencies:
             raise PipxError(
-                f"No apps associated with package {package}. "
+                f"No apps associated with package {display_name}. "
                 "Try again with '--include-deps' to include apps of dependent packages, "
                 "which are listed above. "
                 "If you are attempting to install a library, pipx should not be used. "
@@ -303,7 +308,7 @@ def run_post_install_actions(
             )
         else:
             raise PipxError(
-                f"No apps associated with package {package}. "
+                f"No apps associated with package {display_name}. "
                 "If you are attempting to install a library, pipx should not be used. "
                 "Consider using pip or a similar tool instead."
             )
@@ -318,26 +323,30 @@ def run_post_install_actions(
         if venv.safe_to_remove():
             venv.remove_venv()
         raise PipxError(
-            f"No apps associated with package {package} or its dependencies."
+            f"No apps associated with package {display_name} or its dependencies."
             "If you are attempting to install a library, pipx should not be used. "
             "Consider using pip or a similar tool instead."
         )
 
-    if suffix is None:
-        suffix = ""
     expose_apps_globally(
-        local_bin_dir, package_metadata.app_paths, package, force=force, suffix=suffix
+        local_bin_dir,
+        package_metadata.app_paths,
+        package,
+        force=force,
+        suffix=package_metadata.suffix,
     )
 
     if include_dependencies:
         for _, app_paths in package_metadata.app_paths_of_dependencies.items():
             expose_apps_globally(
-                local_bin_dir, app_paths, package, force=force, suffix=suffix
+                local_bin_dir,
+                app_paths,
+                package,
+                force=force,
+                suffix=package_metadata.suffix,
             )
 
-    print(
-        get_package_summary(venv_dir, package=package, new_install=True, suffix=suffix)
-    )
+    print(get_package_summary(venv_dir, package=package, new_install=True,))
     warn_if_not_on_path(local_bin_dir)
     print(f"done! {stars}", file=sys.stderr)
 
--- a/src/pipx/pipx_metadata_file.py
+++ b/src/pipx/pipx_metadata_file.py
@@ -35,6 +35,7 @@ class PackageInfo(NamedTuple):
     apps_of_dependencies: List[str]
     app_paths_of_dependencies: Dict[str, List[Path]]
     package_version: str
+    suffix: str = ""
 
 
 class PipxMetadata:
@@ -100,11 +101,18 @@ class PipxMetadata:
         }
 
     def from_dict(self, input_dict: Dict[str, Any]) -> None:
-        self.main_package = PackageInfo(**input_dict["main_package"])
+        main_package_data = input_dict["main_package"]
+        if main_package_data["package"] != self.venv_dir.name:
+            # handle older suffixed packages gracefully
+            main_package_data["suffix"] = self.venv_dir.name.replace(
+                main_package_data["package"], ""
+            )
+
+        self.main_package = PackageInfo(**main_package_data)
         self.python_version = input_dict["python_version"]
         self.venv_args = input_dict["venv_args"]
         self.injected_packages = {
-            name: PackageInfo(**data)
+            f"{name}{data.get('suffix', '')}": PackageInfo(**data)
             for (name, data) in input_dict["injected_packages"].items()
         }
 
--- a/src/pipx/pipx_metadata_file.py
+++ b/src/pipx/pipx_metadata_file.py
@@ -39,6 +39,9 @@ class PackageInfo(NamedTuple):
 
 
 class PipxMetadata:
+    # Only change this if file format changes
+    __METADATA_VERSION__: str = "0.2"
+
     def __init__(self, venv_dir: Path, read: bool = True):
         self.venv_dir = venv_dir
         # We init this instance with reasonable fallback defaults for all
@@ -61,9 +64,6 @@ class PipxMetadata:
         self.python_version: Optional[str] = None
         self.venv_args: List[str] = []
         self.injected_packages: Dict[str, PackageInfo] = {}
-
-        # Only change this if file format changes
-        self._pipx_metadata_version: str = "0.1"
 
         if read:
             self.read()
@@ -97,7 +97,7 @@ class PipxMetadata:
             "injected_packages": {
                 name: data._asdict() for (name, data) in self.injected_packages.items()
             },
-            "pipx_metadata_version": self._pipx_metadata_version,
+            "pipx_metadata_version": self.__METADATA_VERSION__,
         }
 
     def from_dict(self, input_dict: Dict[str, Any]) -> None:
--- a/tests/test_pipx_metadata_file.py
+++ b/tests/test_pipx_metadata_file.py
@@ -63,14 +63,17 @@ BLACK_PACKAGE_REF = PackageInfo(
 
 
 def test_pipx_metadata_file_create(tmp_path):
-    pipx_metadata = PipxMetadata(tmp_path)
+    venv_dir = tmp_path / TEST_PACKAGE1.package
+    venv_dir.mkdir()
+
+    pipx_metadata = PipxMetadata(venv_dir)
     pipx_metadata.main_package = TEST_PACKAGE1
     pipx_metadata.python_version = "3.4.5"
     pipx_metadata.venv_args = ["--system-site-packages"]
     pipx_metadata.injected_packages = {"injected": TEST_PACKAGE2}
     pipx_metadata.write()
 
-    pipx_metadata2 = PipxMetadata(tmp_path)
+    pipx_metadata2 = PipxMetadata(venv_dir)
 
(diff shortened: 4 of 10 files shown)


src/pipx/pipx_metadata_file.py
```diff
@@ -35,6 +35,7 @@ class PackageInfo(NamedTuple):
     apps_of_dependencies: List[str]
     app_paths_of_dependencies: Dict[str, List[Path]]
     package_version: str
+    suffix: str = ""
 
 
 class PipxMetadata:
```
itsayellow, line 41: For any change to the pipx metadata format, line 66 `self._pipx_metadata_version: str = "0.1"` needs to be updated with an incremented version string.

I think a good approach would be a new module global `PIPX_METADATA_VERSION = "0.2"` at the top, and change line 66 to `self._pipx_metadata_version: str = PIPX_METADATA_VERSION`

abn: @itsayellow I have made the changes but using a slightly different approach. I reckon since this version is tied to the class, I have made it a class variable. Should ahve the same desired effect. Happy to change it to your suggestion if you think that is better.

itsayellow: Can you use something more like `__METADATA_VERSION__` instead of `__VERSION__`?  I want to make sure nobody confuses this with the version of pipx, it is specifically for the format of the metadata we write to file.

abn: @itsayellow consider it done! :)

abn:
@itsayellow is there any further changes required for this one?

itsayellow:
> @itsayellow are there any further changes required for this one?

Still working on it.  With unpaid volunteer labor you need to be patient. 🙂

abn:
> Still working on it. With unpaid volunteer labor you need to be patient.

Well aware of those pains, was not intending to push for resolution. Just making sure to dot the i's and cross the t's. :)

itsayellow:
Oh my gosh, so currently it looks like parts of pipx are actually broken when using packages with a suffix?  I had no idea.  This is less of an "improvement" and more of a necessary bugfix!

cs01:
@abn Thank you for the PR. I looked it over and it looks good to me -- I have no comments. It is rare to see no lint errors, unit tests passing AND new unit tests all in a single commit. Impressive! 

I'll wait for @itsayellow's review before merging since he wrote the metadata code and knows it better than me.

cs01:
> Oh my gosh, so currently it looks like parts of pipx are actually broken when using packages with a suffix? I had no idea. This is less of an "improvement" and more of a necessary bugfix!

Yeah I was kind of surprised that these changes were needed, though I'm glad they were; they seem like basic features of the suffix functionality.

abn:
@itsayellow ah yes, it is broken. I figured you were aware of this ... limitation of the current implementation. I recently started using pipx quite regularly and is finding the suffix feature rather useful for triage/testing the cli tool I help out with. That is what initiated the fix. :)

@cs01 I think I had a lint failire on my inital push ... so not all clean :wink:

src/pipx/commands/common.py
```diff
@@ -148,7 +149,8 @@ def get_package_summary(
             f"       Please uninstall and install this package, or reinstall-all to fix."
         )
 
-    package_metadata = venv.package_metadata[package]
+    package_metadata = venv.get_package_metadata(package)
```
itsayellow, line 152: I think instead of these two line changes, we need to go a little farther back and change starting line 141:
```
if package is None:
    package = path.name
```
to instead be
```
if package is None:
  package = venv.package_metadata.main_package["package"]
```
Since we have now broken old pipx's one-to-one relationship between path <-> main_package name.

itsayellow: And then I think you don't need `venv.get_package_metadata()` and you don't need to search for the package name with its for-loop.

src/pipx/commands/common.py
```diff
@@ -174,11 +176,12 @@ def get_package_summary(
         python_version,
         python_path,
         package_metadata.package_version,
-        package,
+        package_metadata.package or package,
```
itsayellow, line 179: I think this should not be necessary if you get `package` correct in the first place (see above comment.)

src/pipx/commands/common.py
```diff
@@ -148,7 +149,8 @@ def get_package_summary(
             f"       Please uninstall and install this package, or reinstall-all to fix."
```
itsayellow, line 149: In the function definition for get_package_summary you no longer have any use for argument `suffix` and it should be removed.

itsayellow (changes requested):
In general I'd like to look over this again after the following change that I've alluded to in the comments:

I think you can remove the uses of `package_metadata = venv.get_package_metadata()` and revert back to `package_metadata = venv.package_metadata[package]` if you first make sure you get the correct `package`, which is now not necessarily the name of the directory.  The main_package name is already in the metadata and you should use that (as I've marked up in the source) instead of starting out assuming that the directory path is the package name.

Let me know if there's something I'm missing and this isn't possible in all cases, but from my read I think what I said can be done.

src/pipx/commands/common.py
abn: I am not sure if this will cover all the cases unfortunately. I had something similar in my first attempt. The issue was upgrade of injected packages that failed in this scenario. Additioanlly, the `get_package_metadata` is also useful if the suffix for installed app packages are enabled int he future. While it is probably not the most efficient method, it does the trick reliably. I am happy to make the suggested changes, and handle the key error explicitly in the upgrade code. Unless there is a cleaner way of doing this that I am missing.

abn:
> Let me know if there's something I'm missing and this isn't possible in all cases, but from my read I think what I said can be done.
@itsayellow  see my response for why I ended up doing that. TLDR; upgrade of injected packages was not working as expected.

src/pipx/commands/common.py
itsayellow: Can you explain what's going wrong with upgrade?  I wasn't aware we could upgrade injected packages.

Also, could you explain what you mean by "useful if the suffix for installed app packages are enabled int he future"?

In general I'd like to fix the implicit assumption in the code that pathname==main_package, because ever since we have suffixes that is incorrect.

abn: I am purely going by the possible code-paths here. It might be that I am misunderstanding what `pipx` wants to support. In theory any package in a venv ca be updated based on this signature.

https://github.com/pipxproject/pipx/blob/20238b1153eb5ade0eb851511dc2cf51c7eac4eb/src/pipx/commands/upgrade.py#L15-L23

Replacing the following with `package_metadata = venv.pipx_metadata.main_package` seems incorrect.

https://github.com/pipxproject/pipx/blob/20238b1153eb5ade0eb851511dc2cf51c7eac4eb/src/pipx/commands/upgrade.py#L42

And following that there is also this call, which I expect is intended to expose apps (my understanding of the pipx app concept might be incomplete) from a pipx environment. 

https://github.com/pipxproject/pipx/blob/20238b1153eb5ade0eb851511dc2cf51c7eac4eb/src/pipx/commands/upgrade.py#L70-L72

Since each of the packages now have a `suffix` property, in theory each exposed app can use a suffix. So this implies that if required, one can expose each app with its own suffix or an inherited suffix. This might not be what the project wants, and I can appreciate that.

I feel that in order to break that assumption safely old installations in the wild will need to be updated I feel. Otherwise a `pipx` upgrade can break existing environments. Again, a fair few assumptions here.

itsayellow: Currently we don't support upgrading anything but the main package (see the line `# TODO 20191026: upgrade injected packages also (Issue #79)`)

> Replacing the following with `package_metadata = venv.pipx_metadata.main_package` seems incorrect.

That would be incorrect.  What I was proposing was to obtain metadata on the main package was:
```
package = venv.pipx_metadata.main_package
package_metadata = venv.package_metadata[package]
```

> Since each of the packages now have a suffix property, in theory each exposed app can use a suffix. So this implies that if required, one can expose each app with its own suffix or an inherited suffix. This might not be what the project wants, and I can appreciate that.
>
> I feel that in order to break that assumption safely old installations in the wild will need to be updated I feel. Otherwise a `pipx` upgrade can break existing environments. Again, a fair few assumptions here.

I think right now we don't want to support any suffixes besides the main package, (@cs01 can correct me).  Doing otherwise sounds too complicated and confusing.  Currently the only pipx command with a `--suffix` option is `install`, so only the main package can currently have a suffix, and we don't have to worry about injected packages or dependencies in the wild having a suffix.

Definitely you are correct, `upgrade`'s use of `expose_apps()` needs a `suffix` argument, but I would argue that we currently only suffix the apps from the main package (as is the current behavior.)

In general, so far in pipx, injected packages are definitely second-class citizens.  We are not implementing everything for them that we do for the main package, like suffixes on their apps.

Grafting the `--suffix` option onto pipx clearly had more side-effects than originally envisaged.  I'm just trying to figure out the least-bad way to leave the code after still fixing these bugs.  Viewing all the code paths at this point in time as possible may not make sense because we probably need some refactoring and rethinking to completely integrate suffixes.

itsayellow: Did you actually see a failure with `upgrade` and `package = venv.pipx_metadata.main_package`, or are you basing this based just on the code?

abn: > That would be incorrect. What I was proposing was to obtain metadata on the main package was:

That was a typo on my side. My point still stands that this overrides a mandatory argument in the `upgrade` function. Even when a package is provided, we reset this to the `main_package`. If this is indeed the intended behaviour we should probably drop the `package` parameter and simply rely on the metadata of the `Venv`.

> I would argue that we currently only suffix the apps from the main package (as is the current behavior.)

I am happy to do that, but I just find that a step backwards, especially since there already is a TODO to support this.

> Viewing all the code paths at this point in time as possible may not make sense because we probably need some refactoring and rethinking to completely integrate suffixes.

I agree. However, this is not necessarily handling all code paths, but rather just the ones that seemed obvious due to the interface. And as I said, happy to reduce the scope of the PR.

abn: > package = venv.pipx_metadata.main_package
> package_metadata = venv.package_metadata[package]

Isin't `venv.pipx_metadata.main_package` a `PackageInfo` instance?

itsayellow: > > package = venv.pipx_metadata.main_package
> > package_metadata = venv.package_metadata[package]
> 
> Isin't `venv.pipx_metadata.main_package` a `PackageInfo` instance?

Ah yes you are right.  Sorry I was just typing from my mind.

itsayellow: As for `upgrade()`, right now the `venv_dir` and and `package` arguments I believe are more of an implementation detail that are supposed to track each other--`venv_dir` is just the full path and `package` is just the name of the package.

itsayellow: Another possibility that could be considered is that 
`@property def package_metadata(self)` in class `Venv` returns a dict with package + package_suffix as the keys to the dict instead of just package.  Do you have any idea if that would help things?  Or would it mess up something else that depends on unsuffixed package names?

abn: I recall I had tried to use that as a fix initially; but there was something that felt harder than it should be and existing tests were failing. I can try that again see how far I can get with it. This simply avoids the loop in the cases where the main package is suffixed and is the one for which the metadata is being retrived. Although, overall I think it is probably not ideal as most lookups are package names, not name + suffix.

Additionally, while trying to implement your suggestions, I realiased that the `run_post_install_actions` will also have similar issues. SInce this is also called after inject with a non-main package. So it should handle both cases to support today's use cases. 

I could do something like this in places where a package is passed in. But then again this logic might be better placed in a method under `Venv`.

```py
    suffixed_package_name = f"{venv.pipx_metadata.main_package.package}{venv.pipx_metadata.main_package.suffix}"
    if package == suffixed_package_name:
        package = venv.pipx_metadata.main_package.package or package

    if package not in venv.package_metadata:
        raise PipxError(f"Package {red(bold(package))} not found in {venv.root}.")
```

itsayellow: What if you modified `def package_metadata(self)` to only apply the suffix for the main_package?  The main package is already applied in a separate if-clause anyway.

Is there a way this breaks?

itsayellow: I was playing around with the code a bit, and following seems to work at least with `pipx list`:
```
    @property
    def package_metadata(self) -> Dict[str, PackageInfo]:
        return_dict = self.pipx_metadata.injected_packages.copy()
        if self.pipx_metadata.main_package.package is not None:
            return_dict[
                self.pipx_metadata.main_package.package + self.pipx_metadata.main_package.suffix
            ] = self.pipx_metadata.main_package
        return return_dict
```
I used most of your `pipx_metadata_file.py` code.

I don't think we should try to apply the suffix to everything, just the main package and its apps.

src/pipx/pipx_metadata_file.py
```diff
@@ -35,9 +35,13 @@ class PackageInfo(NamedTuple):
     apps_of_dependencies: List[str]
     app_paths_of_dependencies: Dict[str, List[Path]]
     package_version: str
+    suffix: str = ""
```
itsayellow, line 38: I think `suffix` should be a member of `PipxMetadata`.  I don't think we want more than one suffix for a given venv.

itsayellow:
@cs01, there's some policy questions here that I'm curious for your thoughts about.
My thoughts:
* Only one suffix per venv
* Only the main package and its apps get the suffix (not injected packages + their apps)
* Should `--include-deps` apps also get the suffix?

src/pipx/commands/common.py
abn: I need to double check how existing installs would be affected. Will do that when I get some time to look at this again. 

I suppose the policy question, even if not permanent, should be addressed before reducing the change scope.

abn:
Assuming that the included apps are exposed to path, I'd suggest we should allow suffix configuration there too. 

I do currently have a use case where a certain pip version is a dependency that gets injected and would like to expose multiple versions from different environments. But this suffix might differ from the main package suffix. I appreciate this is not in scope for the PR, but wanted to mention this for the policy decision.

itsayellow:
> I do currently have a use case where a certain pip version is a dependency that gets injected and would like to expose multiple versions from different environments. But this suffix might differ from the main package suffix. I appreciate this is not in scope for the PR, but wanted to mention this for the policy decision.

In this case, why would the pip suffix need to differ from the main package suffix?

itsayellow:
My rationale for a simple policy regarding suffixes *right now*, is that it's already rippling through the code, and absorbing a simple pervasive change now and incrementally improving/expanding it later will be easier (and hopefully more successful) for implementation.  Also, a more complicated system of suffixes would be hard to explain/document from a user perspective.

I am sensitive to how useful a more powerful/expansive approach of suffixes could be, but there is a tradeoff between power and complexity.

cs01:
> @cs01, there's some policy questions here that I'm curious for your thoughts about.
> My thoughts:
> 
> * Only one suffix per venv
> * Only the main package and its apps get the suffix (not injected packages + their apps)
> * Should `--include-deps` apps also get the suffix?

> Also, a more complicated system of suffixes would be hard to explain/document from a user perspective.

We need to consider first whether the feature can be understood and supported by part time volunteers, and second if a significant portions of users would use the feature.

pipx has been slowly accreting new features, and each new feature adds significant complexity. Between inject, `--include-deps`, and metadata tracking, the codebase has become much more complicated. pipx is, IMO, very useful as it is, and this suffix feature, although pretty cool, is a bit of a corner case.

We should not sacrifice maintainability of the codebase for a feature that is not essential, especially since the project is maintained by part time volunteers. This is what happened to pipenv; its complexity became overwhelming and the project suffered.  

At some point, users who want such specific features need to just make a venv and bash script instead of shoehorning it all into pipx. IMO the only way the suffix feature can work and pipx can remain viable is if the feature and implementation is kept extremely simple. I am fine not supporting suffixes at all, or only supporting them with functionality that is reduced but maintainable and easily documented. Maybe that's just me be being selfish because I'm perfectly happy with pipx as it is and don't want increased maintenance burden.

> * Only one suffix per venv
> * Only the main package and its apps get the suffix (not injected packages + their apps)

I'm fine with these limitations. 

> * Should `--include-deps` apps also get the suffix?

If `--include-deps` is used, and the same package was already installed with a different suffix, there will be name collisions if a suffix is not added to the apps of the dependent packages. So I think from a user perspective there has to be. I just wonder how hard this will be from a maintainer's perspective.

Maybe a simpler way to accomplish this and remove limitations is to symlink an entire folder with the suffix name, like `black_suffix` in the bin dir. Then users can run `black_suffix/black` from the command line.

itsayellow:
What if we described suffixes as "experimental and subject to possible change" in both the help and docs?

cs01:
> What if we described suffixes as "experimental and subject to possible change" in both the help and docs?

That sounds like a good idea since things are in flux.

abn:
@cs01 just to add to the discussion, the suffix feature does not need to be complex. From my perspective, simply ensuring the main package is the only one that can be suffixed is good enough to reduce unnecessary complexity. 

The reason why I turned to `pipx` was this particular feature for what it is worth. This makes using multiple versions of python tools in the same environment a breeze. And I do sincerely hope that this feature remains, albeit with possible caveats.

itsayellow:
> @cs01 just to add to the discussion, the suffix feature does not need to be complex. From my perspective, simply ensuring the main package is the only one that can be suffixed is good enough to reduce unnecessary complexity.
> 
> The reason why I turned to `pipx` was this particular feature for what it is worth. This makes using multiple versions of python tools in the same environment a breeze. And I do sincerely hope that this feature remains, albeit with possible caveats.

I think that's a good approach.

To summarize what @cs01 and I (and I believe @abn 🙂)  agree about:
* One suffix only per venv
* Suffix is applied to main package only
* Suffix is applied to apps of main package
* Suffix is applied to main package's dependencies' apps if `--include-deps` is used to install main package

Hopefully with this approach there is an implementation that keeps code complexity down.

itsayellow:
As a minor side note:

> Maybe a simpler way to accomplish this and remove limitations is to symlink an entire folder with the suffix name, like `black_suffix` in the bin dir to the venv's bin dir. Then users can run `black_suffix/black` from the command line, and pipx doesn't have to do anything complicated other than create a venv and install to it.

Does this work?  I couldn't get `<subdirectory>/<app>` in a `PATH` directory to work for me in zsh.

cs01:
> Does this work? I couldn't get `<subdirectory>/<app>` in a `PATH` directory to work for me in zsh.

I just tried it myself and it doesn't work. I assumed you could access folders on your PATH, but that doesn't seem to be the case.

src/pipx/commands/common.py
abn: @itsayellow I have updated things a bit. The change defaults to venv's main package instead of venv root name. I have also removed the fallback loop to determine suffixed packaged.

Decided to retain the helper to abstract away some common functionality, and will also make maintenance easier in the future.

Let me know what you think about this version.

docs/changelog.md
```diff
@@ -1,6 +1,7 @@
 dev
 
 - [docs] Update license
+- [bugfix] Regression in list, inject and upgrade commands when suffixed packages are used.
```
itsayellow, line 4: Can you change "Regression" to "Fixed regression".

tests/test_pipx_metadata_file.py
```diff
@@ -63,14 +63,17 @@
 
 
 def test_pipx_metadata_file_create(tmp_path):
-    pipx_metadata = PipxMetadata(tmp_path)
+    venv_dir = tmp_path / TEST_PACKAGE1.package
+    venv_dir.mkdir()
```
itsayellow, line 67: Why is this necessary?

itsayellow (changes requested):
Thanks, I think this is mostly ok, and a big improvement over our current broken implementation of suffixes.

I mainly have the one question about why you changed the name of the temp directory in `test_metadata_file_create()`.

tests/test_pipx_metadata_file.py
abn: This is required because of https://github.com/pipxproject/pipx/pull/491/files#diff-349c0b5cf5479f95786c8426e2656d2bR105-R109

itsayellow (approved):

joshbode:
Oops - I just noticed this PR - I made the initial suffix PR (#445) but I clearly didn't consider the impact enough (e.g. `pipx list`, etc).

Thank you for fixing this @abn :)

abn:
@cs01 @itsayellow can we tag a new release with this fix in please? Seems like there isin't a new one yet.

PR #496, opened 2020-09-17
upgrade: do not enforce package spec for pipx package

abn:

## Test plan
Tested by running
```
pipx install --suffix @master 'pipx @ git+https://github.com/pipxproject/pipx.git'
pipx upgrade pipx@master
```

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -2,6 +2,7 @@ dev
 
 - [docs] Update license
 - [bugfix] Fixed regression in list, inject and upgrade commands when suffixed packages are used.
+- [bugfix] Do not reset package url during upgrade when main package is `pipx`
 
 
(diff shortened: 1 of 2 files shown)


itsayellow:
I think this is really old code, and I was always curious why this check on `pipx` was present.

@cs01, do you know the intent of this code that @abn wants to remove?

cs01:
Yes this is old! This was back before pipx had the name pipx on PyPI. It was called `pipx-app` but installed to a venv called `pipx`. It used to rename `pipx-app` to `pipx` (or the other way around, I don't remember). Then we got the name `pipx` on PyPI, and logic got updated instead of removed. It should be fine to delete. It's just an old, outdated hack.

itsayellow (approved):

PR #512, opened 2020-10-02
Use native nox for tests (not @cs01 fork of nox)

### LINKED ISSUE #508: Can we use regular nox (and not the special native-venv nox fork?)

If we switched to regular nox (and not @cs01's) fork we could normalize our flow and take advantage of nox development and bug fixes.

The comment in `noxfile.py` says to use @cs01's fork of nox...
```
# until this is fixed in nox. See
# https://github.com/theacodes/nox/issues/199
```

Currently nox theacodes/nox#199 is closed, and a patch (theacodes/nox#231) merged.  nox has option `@nox.session(venv_backend='venv')` that can allow the use of a `venv` backend.

Can we switch to this now?  Is there any reason to keep using the special fork of nox?

Alternative would be to keep doing what we're doing.  However I'm starting to wonder how stale the special fork of nox is getting.

itsayellow:

## Summary of changes
Due presumably to updates in virtualenv, and possibly nox, we can now use regular nox for our tests and don't need to use the special fork of nox. 

Closes #508 

This seems to work fine.  However, in the future if we ever notice something that requires `venv` to be used in nox tests, we can add `venv_backend="venv"` as an argument to `@nox.session()` before `def tests()` in `noxfile.py`.

## Test plan

Tested by running

```
```

### DIFF
--- a/.github/workflows/on-push.yml
+++ b/.github/workflows/on-push.yml
@@ -86,8 +86,8 @@ jobs:
       run: |
         python -m pip install wheel
         python -m pip install regex
-    - name: Install native-venv version of nox
-      run: pip install git+https://github.com/cs01/nox.git@5ea70723e9e6#egg=nox
+    - name: Install nox
+      run: pip install nox
     - name: Execute Tests
       run: |
         nox --non-interactive --session tests-${{ matrix.python-version }}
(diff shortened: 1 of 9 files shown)


cs01 (approved):
Amazing! Thank you. I'm so glad I can delete my branch now.

PR #514, opened 2020-10-04
Reinstall command

### LINKED ISSUE #256: Add `reinstall` command

per @sahensley 's comment in https://github.com/pipxproject/pipx/pull/246

> Also, a bit of a side trek but would it be worth introducing reinstall for an individual package? I could possible see a need if only certain venvs were made with a Python interpreter that is no longer present but it may be a bit of a niche case. The instance I've run into this when I've had multiple Python versions compiled via Pyenv.

Add top-level `reinstall` command, as an alternative to the `reinstall-all` command.

Uninstall then reinstall the package, but that is not super very friendly.

itsayellow:

## Summary of changes
Adds a `reinstall` command to only reinstall one venv.

Closes #256 

## Test plan

Tested by running
```
pipx install pycowsay
pipx reinstall pycowsay
```

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -5,7 +5,7 @@ dev
 - [bugfix] Do not reset package url during upgrade when main package is `pipx`
 - Update help text to show description for `ensurepath` and `completions` help
 - Add support for user-defined default python interpreter via new PIPX_DEFAULT_PYTHON.  Helpful for use with pyenv among other uses.
-
+- Add reinstall command for reinstalling a single venv.
 
 0.15.5.1
 
--- a/src/pipx/commands/__init__.py
+++ b/src/pipx/commands/__init__.py
@@ -2,12 +2,11 @@ from .ensure_path import ensure_pipx_paths
 from .inject import inject
 from .install import install
 from .list_packages import list_packages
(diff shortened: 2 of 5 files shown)


itsayellow:
It would be simple to make this take multiple argument for multiple venvs.  I kept it at one venv because that matched the other pipx sub-commands.

gaborbernat:
> Adds a `reinstall` command to only reinstall one venv.

Any reason why just not make this the behaviour of ``pipx install -f`` or ``pipx install --overwrite``?

itsayellow:
> Any reason why just not make this the behaviour of `pipx install -f` or `pipx install --overwrite`?

One difference would be that this completely reinstalls the venv, including re-injecting all of the injected packages, while `install` is currently only charged with affecting the main package of a venv.  

This command also reuses all of the original options for the venv and installed packages like `pip_args`, `venv_args`, `install_dependencies` without forcing you to re-supply them on the command line, like `install` might imply you need to do.

itsayellow:
I also just thought it would make sense to the user as an analog to `upgrade` / `upgrade-all`.

gaborbernat (approved):

PR #516, opened 2020-10-06
Fix pep508

### LINKED ISSUE #515: Installation not pulling in optional extras-requires packages.
(Full disclosure, this issue is occurring after an OS upgrade, though I've uninstalled and re-installed pipx) -- great tool by the way!!

I have a project on Bitbucket (private). I have historically installed locally with pipx; 
`pipx install "my-project[cli] @ git+ssh://git@bitbucket.org/my-company/myproject.git"`

The setup.cfg for the project has an install-requires of
`[options.extras_require]`
```
cli =
    begins
    docker
```

Installing goes smoothly, but the `begins` and `docker` packages do not install. I verify by looking within `/home/steve/.local/pipx/venvs/my-project/lib/python/site-packages`

[pipx-verbose.txt](https://github.com/pipxproject/pipx/files/5330613/pipx-verbose.txt)

**Expected behavior**
Historically, this would have correctly detected the [cli] designation and installed the appropriate packages. I'm not sure if this has to do with a recent upgrade or something is broken on my end. 

If it is an issue on my part, can you recommend troubleshooting tips? 

Thank you, Sincerely, Steve Jackson

itsayellow:

## Summary of changes
Fixes bug in handling of PEP508 package specifications including URLs and extras

Closes #515 

I was too aggressive in the original code for "cleaning" PEP508 package specifications for pipx.  I originally removed the package name if there was a URL present, on the idea that the user-supplied name might not match the URL and we should just use the URL only to avoid confusion.  Unfortunately, this also removed the only way to supply package extras with a URL.

The new code is more surgical.  The basic idea: convert package specification string to Requirement object, null out anything we don't want as pipx, convert Requirement object back to string.  We then use that string for `package_or_url`.

To guard against the user supplying a mismatched package name to the URL in their PEP508 package specification, I modified `valid_pypi_name` to return `None` if a URL is present, signifying to the calling code that it needs to do a trial install of the URL to determine package name, even if it's present in the PEP508 package specifier.   I also, introduced a new function `fix_package_name` which fixes the package_name part of a spec if necessary.

## Test plan

`pipx install <package>[<extra>] @ URL`

Does anybody have a good suggestion for a test package that uses a URL and has at least one possible extra?  I couldn't easily find one.

### DIFF
--- a/tests/test_package_specifier.py
+++ b/tests/test_package_specifier.py
@@ -50,6 +50,16 @@ TEST_DATA_PATH = "./testdata/test_package_specifier"
             "black@ https://github.com/ambv/black/archive/18.9b0.zip",
             True,
         ),
+        (
+            "black[extra] @ https://github.com/ambv/black/archive/18.9b0.zip",
+            "black[extra]@ https://github.com/ambv/black/archive/18.9b0.zip",
+            True,
+        ),
+        (
+            'my-project[cli] @ git+ssh://git@bitbucket.org/my-company/myproject.git ; python_version<"3.7"',
+            "my-project[cli]@ git+ssh://git@bitbucket.org/my-company/myproject.git",
+            True,
+        ),
         ("path/doesnt/exist", "non-existent-path", False,),
         (
             "https:/github.com/ambv/black/archive/18.9b0.zip",
--- a/tests/test_package_specifier.py
+++ b/tests/test_package_specifier.py
@@ -3,12 +3,50 @@ from pathlib import Path
 import pytest  # type: ignore
 
 from pipx.package_specifier import (
+    fix_package_name,
     parse_specifier_for_install,
     parse_specifier_for_metadata,
+    valid_pypi_name,
 )
 from pipx.util import PipxError
 
 TEST_DATA_PATH = "./testdata/test_package_specifier"
+
+
+@pytest.mark.parametrize(
+    "package_spec_in,package_name_out",
+    [
+        ("Black", "black"),
+        ("https://github.com/ambv/black/archive/18.9b0.zip", None),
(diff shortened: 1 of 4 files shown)


tests/test_package_specifier.py
```diff
@@ -3,14 +3,52 @@
...
+        ),
+        (
+            "nox@https://github.com/ambv/black/archive/18.9b0.zip",
+            "black",
+            "black@ https://github.com/ambv/black/archive/18.9b0.zip",
+        ),
+        (
+            "nox[extra]@https://github.com/ambv/black/archive/18.9b0.zip",
+            "black",
+            "black[extra]@ https://github.com/ambv/black/archive/18.9b0.zip",
```
gaborbernat, line 44: These moved under the psf org now, no longer under ambv.

gaborbernat (approved):
> Does anybody have a good suggestion for a test package that uses a URL and has at least one possible extra? I couldn't easily find one.

What's wrong with the black git URIs you use in the tests?

tests/test_package_specifier.py
itsayellow: For these tests they are just made up urls, so it doesn't really matter if they exist, only that they are the proper syntax.

I don't think black has an extra called `extra`, I just made that up.

I wanted to find a good package with an extra I could actually try and install.

itsayellow:
Just found that pipenv has both a `dev` extra and a `tests` extra.  Installing via PEP508 with both extra and URL with this PR's code seems to work properly.

itsayellow:
nox has a `tox_to_nox` extra and it is MUCH smaller to download than pipenv.

itsayellow:
Thanks for the reviews, @gaborbernat , I appreciate the feedback and everybody else seems busy. 🙂

PR #519, opened 2020-10-08
Propose better way to register fish completions

Tadaboody:
- The standard way to add fish completions
- Doesn't require require reloading the shell

## Test plan

Tested by running
```fish
❯ pipx completions

Add the appropriate command to your shell's config file
so that it is run on startup. You will likely have to restart
or re-login for the autocompletion to start working.
...

fish:
    # Not required to be in the config file, only run once
    register-python-argcomplete --shell fish pipx >~/.config/fish/completions/pipx.fish

❯ register-python-argcomplete --shell fish pipx >~/.config/fish/completions/pipx.fish
```

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,6 +1,7 @@
 dev
 
 - [docs] Update license
+- [docs] Display a more idomatic command for registering completions on fish.
 - [bugfix] Fixed regression in list, inject, upgrade, reinstall-all commands when suffixed packages are used.
 - [bugfix] Do not reset package url during upgrade when main package is `pipx`
 - Updated help text to show description for `ensurepath` and `completions` help
--- a/src/pipx/constants.py
+++ b/src/pipx/constants.py
@@ -70,7 +70,8 @@ tcsh:
     eval `register-python-argcomplete --shell tcsh pipx`
 
 fish:
-    register-python-argcomplete --shell fish pipx | source
+    # Not required to be in the config file, only run once
+    register-python-argcomplete --shell fish pipx >~/.config/fish/completions/pipx.fish
 
 """
 )


itsayellow:
Do any @pipxproject/maintainers know fish?  It would be nice to have a fish user review this.

cs01:
> Do any @pipxproject/maintainers know fish? It would be nice to have a fish user review this.

I don’t. I think I just copied instructions from here https://github.com/kislyuk/argcomplete. I’m on my phone now so can’t dive in but if they are incomplete or inaccurate it would be good to update that link at the source, then pipx can copy that.

Tadaboody:
FWIW argcomplete suggests the same ^_^

```fish
❯ register-python-argcomplete --help
usage: register-python-argcomplete [-h] [--no-defaults] [--complete-arguments ...]
                                   [-s {bash,tcsh,fish}] [-e EXTERNAL_ARGCOMPLETE_SCRIPT]
                                   executable [executable ...]

Register a Python executable for use with the argcomplete module.

To perform the registration, source the output of this script in your bash shell
(quote the output to avoid interpolation).

Example:

    $ eval "$(register-python-argcomplete my-favorite-script.py)"

For Tcsh

    $ eval `register-python-argcomplete --shell tcsh my-favorite-script.py`

For Fish

    $ register-python-argcomplete --shell fish my-favourite-script.py > ~/.config/fish/my-favourite-script.py.fish
```

cecep2:
I'm using Fish and I can confirm that this works! But two comments:

1. When the directory `~/.config/fish/completions/` doesn't exist (which was the case for me), `register-python-argcomplete --shell fish pipx >~/.config/fish/completions/pipx.fish` will give you an error message. To avoid this, I would suggest to simply add the creation of this directory to the instructions and suggest users to do:

```fish
mkdir -p ~/.config/fish/completions ; register-python-argcomplete --shell fish pipx > ~/.config/fish/completions/pipx.fish
```

2. While you only have to run this command once, I like my dotfile setup to work out of the box on a new machine. For people like me, we could add a suggestion to put something like in to their `config.fish` file:

```fish
if begin type -q pipx; and not test -e ~/.config/fish/completions/pipx.fish; end
    mkdir -p ~/.config/fish/completions ; register-python-argcomplete --shell fish pipx > ~/.config/fish/completions/pipx.fish
end
```

This only executes the `register-python-argcomplete` command if `pipx` is installed and the file `~/.config/fish/completions/pipx.fish` doesn't exist.

gaborbernat (approved):
@cecep2 thanks for your comments. I think people will know if they get the folder is missing error to create it and run it again. As for the second use case feels advanced enough, so that those users will again know.

cecep2:
> @cecep2 thanks for your comments. I think people will know if they get the folder is missing error to create it and run it again.

I admit I didn't get it immediately. When the directory doesn't exist, the error message looks like this:

```
> register-python-argcomplete --shell fish pipx > ~/.config/fish/completions/pipx.fish
<W> fish: An error occurred while redirecting file '/Users/stefan/.config/fish/completions/pipx.fish'
open: No such file or directory
```

My first thought was: "Well, I want to create the file, so obviously it does not yet exist!". Needed a few seconds to understand that the missing directory is the issue, not the missing of the `pipx.fish` file I want to create. But sure, a minor issue.

Tadaboody:
@cecep2 I feel the missing dir is more of a fish problem so I created https://github.com/fish-shell/fish-shell/issues/7402

PR #535, opened 2020-10-25
Command select

ctbur:

## Summary of changes
I added both the `pipx select` and `pipx deselect` commands as promised in https://github.com/pipxproject/pipx/issues/505. I tried to capture your concerns in tests, see `test_select.py`. Please let me know if I should clarify something or add another test case.

## Test plan
Tested using the test cases in `test_select.py` and `test_deselect.py`.

### DIFF
--- a/src/pipx/commands/common.py
+++ b/src/pipx/commands/common.py
@@ -180,8 +180,8 @@ def find_selected_venvs_for_package(venv_container: VenvContainer, package: str)
             continue
 
         package_info = package_metadata[venv.main_package_name]
-        print(f'package: {package_info.package} == {package} and {package_info.selected}')
-        if package_info.package == package and package_info.selected:
+        print(f'package: {package_info.package} == {package} and {package_info.selected_as_default}')
+        if package_info.package == package and package_info.selected_as_default:
             print(f'added')
             selected_venvs.append(venv)
 
--- a/src/pipx/commands/common.py
+++ b/src/pipx/commands/common.py
@@ -11,7 +11,7 @@ from typing import Dict, List, Optional
 
 import userpath  # type: ignore
 
-from pipx import constants
+from pipx import constants, emojies
 from pipx.colors import bold, red
 from pipx.emojies import hazard, stars
 from pipx.package_specifier import parse_specifier_for_install, valid_pypi_name
@@ -169,14 +169,16 @@ def find_selected_venvs_for_package(venv_container: VenvContainer, package: str)
     for venv_dir in venv_container.iter_venv_dirs():
         # optimization: don't check packages that are not of the format f'{package}{suffix}'
         if not venv_dir.name.startswith(package):
-            print(f'skipping {venv_dir.name}')
             continue
 
         venv = Venv(venv_dir)
         package_metadata = venv.package_metadata
         if not package_metadata:
-            print(f'skipping2 {venv_dir.name}')
-            # TODO: print warning message
+            print(
+                f"{emojies.hazard} Package '{venv_dir.name}' has missing internal pipx metadata.\n"
+                f"    It was likely installed using a pipx version before 0.15.0.0.\n"
+                f"    Please uninstall and install this package, or reinstall-all to fix."
+            )
             continue
 
         package_info = package_metadata[venv.main_package_name]
--- a/src/pipx/commands/common.py
+++ b/src/pipx/commands/common.py
@@ -17,16 +17,59 @@ from pipx.emojies import hazard, stars
 from pipx.package_specifier import parse_specifier_for_install, valid_pypi_name
 from pipx.pipx_metadata_file import PackageInfo
 from pipx.util import WINDOWS, PipxError, mkdir, rmdir
-from pipx.venv import Venv
+from pipx.venv import Venv, VenvContainer
 
 
-def expose_apps_globally(
+def expose_package_globally(
+        local_bin_dir: Path, package_metadata: PackageInfo, *, force: bool, suffix: str = ""
+) -> None:
+    _expose_apps_globally(
+        local_bin_dir,
+        package_metadata.app_paths,
+        force=force,
+        suffix=suffix,
+    )
+
+    if package_metadata.include_dependencies:
+        for _, app_paths in package_metadata.app_paths_of_dependencies.items():
+            _expose_apps_globally(
+                local_bin_dir,
+                app_paths,
+                force=force,
+                suffix=suffix,
+            )
+
+
+def _expose_apps_globally(
     local_bin_dir: Path, app_paths: List[Path], *, force: bool, suffix: str = "",
 ) -> None:
     if not _can_symlink(local_bin_dir):
         _copy_package_apps(local_bin_dir, app_paths, suffix=suffix)
     else:
         _symlink_package_apps(local_bin_dir, app_paths, force=force, suffix=suffix)
+
+
+def unexpose_package_globally(local_bin_dir: Path, package_metadata: PackageInfo):
+    app_paths_str = package_metadata.apps.copy()
+    if package_metadata.include_apps:
+        app_paths_str.extend(package_metadata.apps_of_dependencies)
+
+    app_paths = [Path(app_path_str) for app_path_str in app_paths_str]
+    unexpose_apps_globally(local_bin_dir, app_paths)
+
+
+def unexpose_apps_globally(local_bin_dir: Path, app_paths: List[Path]):
+    for file in local_bin_dir.iterdir():
+        if WINDOWS:
+            for b in app_paths:
+                if file.name == b.name:
+                    file.unlink()
+        else:
+            symlink = file
+            for b in app_paths:
+                if symlink.exists() and b.exists() and symlink.samefile(b):
+                    logging.info(f"removing symlink {str(symlink)}")
+                    symlink.unlink()
 
 
 _can_symlink_cache: Dict[Path, bool] = {}
@@ -114,6 +157,35 @@ def _symlink_package_apps(
                 f"{hazard}  Note: {app_name_suffixed} was already on your PATH at "
                 f"{existing_executable_on_path}"
             )
+
+
+def find_selected_venvs_for_package(venv_container: VenvContainer, package: str) -> List[Venv]:
+    """
+    Returns all venvs that are selected as default for a package. If all venv metadata is valid, then only one venv is
+    selected. However, this function still searches for multiple ones to be able to recover from an invalid state.
+    """
+    selected_venvs = []
+
+    for venv_dir in venv_container.iter_venv_dirs():
+        # optimization: don't check packages that are not of the format f'{package}{suffix}'
+        if not venv_dir.name.startswith(package):
+            print(f'skipping {venv_dir.name}')
+            continue
+
+        venv = Venv(venv_dir)
+        package_metadata = venv.package_metadata
+        if not package_metadata:
+            print(f'skipping2 {venv_dir.name}')
+            # TODO: print warning message
+            continue
+
+        package_info = package_metadata[venv.main_package_name]
+        print(f'package: {package_info.package} == {package} and {package_info.selected}')
+        if package_info.package == package and package_info.selected:
+            print(f'added')
+            selected_venvs.append(venv)
+
+    return selected_venvs
 
 
 def get_package_summary(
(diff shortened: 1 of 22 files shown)


src/pipx/commands/common.py
```diff
@@ -11,16 +11,36 @@
...
 from pipx.emojies import hazard, stars
 from pipx.package_specifier import parse_specifier_for_install, valid_pypi_name
 from pipx.pipx_metadata_file import PackageInfo
 from pipx.util import WINDOWS, PipxError, mkdir, rmdir
-from pipx.venv import Venv
+from pipx.venv import Venv, VenvContainer
 
 
-def expose_apps_globally(
+def expose_package_globally(
```
ctbur, line 23: This function was extracted from `install` and `upgrade`.

src/pipx/commands/common.py
```diff
@@ -29,6 +49,29 @@ def expose_apps_globally(
...
+def unexpose_package_globally(local_bin_dir: Path, package_metadata: PackageInfo):
+    app_paths_str = package_metadata.apps.copy()
+    if package_metadata.include_apps:
+        app_paths_str.extend(package_metadata.apps_of_dependencies)
+
+    app_paths = [local_bin_dir / app_path_str for app_path_str in app_paths_str]
+    unexpose_apps_globally(local_bin_dir, app_paths)
+
+
+def unexpose_apps_globally(local_bin_dir: Path, app_paths: List[Path]):
```
ctbur, line 61: Same here, but extracted from `uninstall`.

src/pipx/commands/common.py
```diff
@@ -116,6 +159,35 @@ def _symlink_package_apps(
             )
 
 
+def find_selected_venvs_for_package(venv_container: VenvContainer, package: str, *, verbose=False) -> List[Venv]:
```
ctbur, line 162: This is used to guard against selecting multiple packages with suffixes at the same time, as well as trying to install a package with no suffix when the suffixed version is selected.

src/pipx/commands/deselect.py
```diff
@@ -0,0 +1,64 @@
+from pipx import constants
+from pipx.colors import red, bold
+from pipx.commands.common import find_selected_venvs_for_package, expose_package_globally, unexpose_package_globally
+from pipx.util import PipxError
+from pipx.venv import VenvContainer, Venv
+
+
+def deselect(
+    package_name: str,
```
ctbur, line 9: As per description in `main.py`: `package_name` can be a bare package name or have a suffix.

src/pipx/pipx_metadata_file.py
```diff
@@ -35,11 +35,13 @@ class PackageInfo(NamedTuple):
...
     package_version: str
     suffix: str = ""
+    # whether the package was selected with `pipx select`
+    selected_as_default: bool = False
 
 
 class PipxMetadata:
     # Only change this if file format changes
-    __METADATA_VERSION__: str = "0.2"
+    __METADATA_VERSION__: str = "0.2"  # TODO: does version need to be increased if field with default is added?
```
ctbur, line 44: Was not sure here. Please select:
```suggestion
    # This is also necessary when adding a field that has a default value
    __METADATA_VERSION__: str = "0.3"
```

ctbur: Or:
```suggestion
    # This is not necessary when adding a field that has a default value
    __METADATA_VERSION__: str = "0.2"
```

src/pipx/commands/upgrade.py
```diff
@@ -67,21 +68,20 @@ def upgrade(
     display_name = f"{package_metadata.package}{package_metadata.suffix}"
     new_version = package_metadata.package_version
 
-    expose_apps_globally(
+    # remove existing links, in case a binary is not provided in new version
+    unexpose_package_globally(constants.LOCAL_BIN_DIR, package_metadata)
```
ctbur, line 72: Technically unrelated to this PR. I realized that a package might not provide a binary anymore after an upgrade. If that happens, a dead symlink would remain. The fix is to remove everything and expose then binaries again.

src/pipx/commands/common.py
```diff
@@ -116,6 +153,37 @@ def _symlink_package_apps(
...
+    """
+    Returns all venvs that are selected as default for a package. If all venv metadata is valid, then only one venv is
+    selected. However, this function still searches for multiple ones to be able to recover from an invalid state.
+    """
+    selected_venvs = []
+
+    for venv_dir in venv_container.iter_venv_dirs():
+        # optimization: don't check packages that are not of the format f'{package}{suffix}'
+        if not venv_dir.name.startswith(package):
+            continue
```
ctbur, line 168: Could be removed, though generally optimizations are effective if they reduce the number of filesystem interactions.

src/pipx/commands/select.py
```diff
@@ -0,0 +1,97 @@
...
+            pip_args=selected_package_metadata.pip_args,
+            include_dependencies=selected_package_metadata.include_dependencies,
+            include_apps=selected_package_metadata.include_apps,
+            is_main_package=True,
+            suffix=selected_package_metadata.suffix,
+            selected_as_default=False,
+        )
+
+        # remove links of previously selected package
+        unexpose_package_globally(constants.LOCAL_BIN_DIR, selected_package_metadata)
```
ctbur, line 73: This is a bug. It will remove all links of the previously selected package, without recreating those with a suffix. Will fix and update the tests to cover this case.

src/pipx/main.py
```diff
@@ -648,14 +713,16 @@ def cli() -> int:
         if not parsed_pipx_args.command:
             parser.print_help()
             return 1
-        return run_pipx_command(parsed_pipx_args)
+        # return exit code of command or 0 if none provided
+        return run_pipx_command(parsed_pipx_args) or 0
```
ctbur, line 717: Added this because I ran into issues with `assert <some install command>` as it would sometimes return `None`.

src/pipx/pipx_metadata_file.py
```diff
@@ -35,11 +35,14 @@ class PackageInfo(NamedTuple):
...
     suffix: str = ""
+    # whether the package was selected with `pipx select`
+    selected_as_default: bool = False
 
 
 class PipxMetadata:
     # Only change this if file format changes
-    __METADATA_VERSION__: str = "0.2"
+    # This is also necessary when adding a field that has a default value
+    __METADATA_VERSION__: str = "0.3"
```
itsayellow, line 45: When you modify the metadata you need to update the code to gracefully recover from old metadata.  The immediate task is to modify `_convert_legacy_metadata()` to convert from all previous versions to the newest version.  Likely this will just mean adding the default value to the read json for `selected_as_default=False`.

src/pipx/commands/common.py
itsayellow: You probably also need to look for injected packages in the same venv that have been injected with `--include-apps` and/or `--include-deps`.

itsayellow:
OK, so let me think out loud here in the comments...

So you don't handle any injected packages with `select` and `unselect`, because injected packages in a suffixed venv do not have suffixes.

Is that correct?

itsayellow:
Another question:

One way I could imagine keeping track of the "selected" package, is to make a directory named the un-suffixed package name, and inside of it you have one file called `select`.  This file is a text file with one-line that is the name of the suffixed package that is selected by it.  

Then you wouldn't need to modify the metadata, and you wouldn't have to scan every suffixed package and load its venv and metadata to find the selected one.

e.g.
```
venvs/
    pylint_2.5.1/
    pylint_2.5.2/
    pylint/
        select (contains text "pylint_2.5.2")
```

Is there a drawback to this method?

src/pipx/pipx_metadata_file.py
ctbur: Should this not be automatically populated because I set a default value [here](https://github.com/pipxproject/pipx/pull/535/files#diff-902c1a727ef195654d9382fd98ccae279cf6eaad86f72621c3e8a2fe28d15d30R39)?

ctbur:
> So you don't handle any injected packages with `select` and `unselect`, because injected packages in a suffixed venv do not have suffixes.

I didn't consider injected packages at all. Does it make sense for injected packages in suffixed venvs not to have a suffix? I remove all links to a package's apps in several places. Are they affected and do I need to reinstate them in a certain way?

> [...] make a directory named the un-suffixed package name, and inside of it you have one file called select.  [...] Is there a drawback to this method?

Yes I thought about this. I decided against it because of several reasons:
- I didn't know how other commands work and they commands may mistake it for a full venv.
- I could imagine that sometimes a user deletes a venv directory directly. By storing the information in the same directory, this is less likely to consist inconsistency.
- The setting `is_selected_as_default` is related to `suffix`, which is also in the metadata file.

The only drawback that I was able to come up for the current solution was that it requires `O(#packages)` metadata lookups.

src/pipx/pipx_metadata_file.py
itsayellow: I think that should be correct.  But you still need to modify `_convert_legacy_metadata()` so that it doesn't raise a PipxError if it receives `pipx_metadata_version==0.2`.  And double-check that the other versions are still doing the proper thing (which I think they will.)

itsayellow:
> > So you don't handle any injected packages with `select` and `unselect`, because injected packages in a suffixed venv do not have suffixes.
> 
> I didn't consider injected packages at all. Does it make sense for injected packages in suffixed venvs not to have a suffix? I remove all links to a package's apps in several places. Are they affected and do I need to reinstate them in a certain way?

In general suffixes and injected packages used together is probably too complicated a scenario to be easily understandable by a user.  Right now we just don't suffix injected packages, even if they have apps.  This might be worth more thought to make sure that if a user does combine suffixes and injected packages, the behavior can at least be described and maybe even makes sense.

I think as long as you are only looking at the metadata for the main package, if you are only unexposing those apps in that metadata, it should not affect the injected packages.  You might want to double check with an experiment.

itsayellow:
> > [...] make a directory named the un-suffixed package name, and inside of it you have one file called select.  [...] Is there a drawback to this method?
> 
> Yes I thought about this. I decided against it because of several reasons:
> 
>     * I didn't know how other commands work and they commands may mistake it for a full venv.
> 
>     * I could imagine that sometimes a user deletes a venv directory directly. By storing the information in the same directory, this is less likely to consist inconsistency.
> 
>     * The setting `is_selected_as_default` is related to `suffix`, which is also in the metadata file.
> 
> 
> The only drawback that I was able to come up for the current solution was that it requires `O(#packages)` metadata lookups.

In general you shouldn't worry about a user deleting a venv manually.  If they do that, in general things will break (for a regular venv there will be orphaned apps in the `bin` directory.)

You are right that making a simple "select directory" would need to be addressed in all the other commands that access the venvs, and this would be a complication and take more code.

I was just asking because I wasn't sure how long it would take to load every suffix venv and parse its metadata.  Normally everything we do in pipx doesn't require scanning through multiple venvs.  

Also, I hesitate to release another metadata version if it can be avoided.  But in this case as you said before, it's very easy to update the metadata because the `is_selected_as_default` has a clear and unambiguous default (`False`) if it doesn't exist.

I think on balance this is approach is good, but I'm just thinking about it.

itsayellow:
The more I think about this, the more I prefer using a directory with a file in it more than the metadata mechanism for recording a selected venv.

Here is my rationale:
* In general it is much more direct and obvious
  * Debugging is easier, because you can just read a directory and a file in `venvs/` to see the select state rather than searching through multiple metadata files.
  * I believe it would make it easier for novice pipx contributors to understand both the mechanism and the code (this is a concern since as an open-source project, we get a lot of first-time contributors.)
* It won't be bad to modify the existing commands to handle a "select" directory, and we'd probably modify them anyway even for the metadata approach
  * It should be quick and simple to check that a `venvs/` subdirectory is a "select directory" (just check for the select file.)
  * To help the user we probably want to modify every command anyway to give specific guidance if they are incorrectly trying to execute a command on a "select" name.  e.g. "You are trying to uninstall a select venv, you should deselect instead." 
* The alternative metadata approach means every time we `install` (even if the user doesn't use the `select` feature) we would have to scan for adjacent suffixed directories.  I'd hesitate to slow down a basic operation even if the user isn't using an advanced feature.

I would be happy to walk you through what would need to be modified for the various commands to implement a "select venv subdirectory" approach.

gaborbernat:
Closing due to lack of activity. I Will reopen if you pick it up again.

PR #538, opened 2020-10-31
Wrap environment variable documentation to match rest of help output

spazm:
* changed case to match the output of argparse doc section headers.
* This is the same shutil command that argparse uses under the hood
  for determining the width for wrapping.
* indent long strings that are already processed by dedent.
* shutil is already installed as a requirement for argparse.
* The env lines end up long (longer than 88) to work correctly for
  wrapping and indenting.

Addresses wrapping issue mentioned in #523:

| Also, another minor improvement would be to get this help text to
| properly wrap when being viewed from the command (It currently
| doesn't wrap on the console properly.)

* [x] I have added an entry to `docs/changelog.md`

## Test plan

Tested by running
```
# command(s) to exercise these changes
# Run help with various screen widths.  Compare with output of command sent into pipe which does not detect width:

# wide terminal, mine is 137:
% echo $COLUMNS
137
% pipx --version
[...snip...]
optional environment variables:
  PIPX_HOME:           Overrides default pipx location. Virtual Environments will be installed to $PIPX_HOME/venvs.
  PIPX_BIN_DIR:        Overrides location of app installations. Apps are symlinked or copied here.
  USE_EMOJI:           Override emoji behavior. Default value varies based on platform.
  PIPX_DEFAULT_PYTHON: Overrides default python used for commands.
[...snip...]

# capped to a default 78 chars:
% pipx --version | cat

[...snip...]
optional environment variables:
  PIPX_HOME:           Overrides default pipx location. Virtual Environments
                       will be installed to $PIPX_HOME/venvs.
  PIPX_BIN_DIR:        Overrides location of app installations. Apps are
                       symlinked or copied here.
  USE_EMOJI:           Override emoji behavior. Default value varies based on
                       platform.
  PIPX_DEFAULT_PYTHON: Overrides default python used for commands.
[...snip...]

# tiny screen
% echo $COLUMNS
55
% pipx --help
[...snip...]
optional environment variables:
  PIPX_HOME:           Overrides default pipx
                       location. Virtual Environments
                       will be installed to
                       $PIPX_HOME/venvs.
  PIPX_BIN_DIR:        Overrides location of app
                       installations. Apps are
                       symlinked or copied here.
  USE_EMOJI:           Override emoji behavior.
                       Default value varies based on
                       platform.
  PIPX_DEFAULT_PYTHON: Overrides default python used
                       for commands.

optional arguments:
  -h, --help            show this help message and
                        exit
  --version             Print version and exit
[...snip...]
```

PR #539, opened 2020-10-31
Fix/default python during nox test

spazm:

## Test plan

Tested by running
```

% nox -s docs
# view changes in docs/docs.md
% grep overridden docs/docs.md | tail -1
/usr/bin/python and can be overridden by setting the environment

# compare with output when `PIPX__DOX_DEFAULT_PYTHON` is not set.
% pipx install --help |grep overridden |tail -1
/homd/spazm/src/github/spazm/pipx/venv/bin/python and can be overridden by setting the environment
```

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -77,6 +77,7 @@ def lint(session):
 @nox.session(python="3.8")
 def docs(session):
     session.install(*doc_dependencies)
+    session.env["PIPX__DOC_DEFAULT_PYTHON"] = "/usr/bin/python"
     session.run("python", "generate_docs.py")
     session.run("mkdocs", "build")
 
--- a/noxfile.py
+++ b/noxfile.py
@@ -77,7 +77,7 @@ def lint(session):
 @nox.session(python="3.8")
 def docs(session):
     session.install(*doc_dependencies)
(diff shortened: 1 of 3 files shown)


noxfile.py
```diff
@@ -77,6 +77,7 @@ def lint(session):
 @nox.session(python="3.8")
 def docs(session):
     session.install(*doc_dependencies)
+    session.env["PIPX__DOC_DEFAULT_PYTHON"] = "/usr/bin/python"
```
itsayellow, line 80: How about instead of 
`/usr/bin/python`
we use the text
`typically the python used to execute pipx`
?

spazm: This is a great idea, it looks much better when rendered.

I've made the change and also adjusted the line break for better fill.

itsayellow (changes requested):
Looks good, just need to pass Lint (re-run black on `noxfile.py`)

spazm:
Hiya pipx team.  

Thanks for a lovely project.  If you enjoy and/or approve this PR and #538 , I'd love if you tagged them both with `hacktoberfest-accepted` for hacktoberfest2020.

Happy Code-o-ween!
@spazm

itsayellow (approved):

PR #542, opened 2020-11-09
venv_dir, skip arguments: canonicalize before matching existing venv dirs

itsayellow:

## Summary of changes
Pipx commands accepting `venv_dir` as an argument currently won't match if a python-library-name equivalent is used for an existing venv.  One needs to specify exactly pipx's version (the canonical pypi name).  This means all lower-case, and only dashes (-).

~~This PR changes `VenvContainer.get_venv_dir()` to try the literal `venv_dir` pipx command argument first, and then the canonicalized version of the `venv_dir` argument to see if it matches an existing venv.  If neither match, it returns the literal venv_dir supplied argument as before.~~

~~I would've just canonicalized the entire `venv_dir` argument, but unfortunately this is not compatible with suffixed venvs, as it is currently possible for them to have non-canonical characters.  One possible change that could fix this would be to also canonicalize the suffix recorded as venv directory name.  I'd be interested to know if there is interest in that approach.  The suffix could still remain the same, but the suffixed venv directory would be canonicalized.~~

EDIT: I changed this to canonicalize both `venv_dir` and `--skip` arguments before trying to match existing venv directory names.  Also new venv directory names are now strictly canonicalized, even suffixes.  Internally, suffixes are still stored literally, so app names have the un-canonicalized suffix.  And whenever metadata is available, if there is a suffix we use the literal un-canonicalized suffix to display in messages.

## Test plan

This is best seen on linux which has a case-sensitive file system.  With this PR on linux:
```
> pipx install kaggle                                                 
  installed package kaggle 1.5.9, Python 3.9.0                                  
    - kaggle                                                                    
done! ✨ 🌟 ✨                                                                  
> pipx uninstall KaGgLe                                        
uninstalled kaggle! ✨ 🌟 ✨                                                    
```

Existing behavior on linux below:
```
> pipx install kaggle                                                 
  installed package kaggle 1.5.9, Python 3.9.0                                  
    - kaggle                                                                    
done! ✨ 🌟 ✨                                                                  
> pipx uninstall KaGgLe                                               
Nothing to uninstall for KaGgLe 😴                                              
```

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -6,13 +6,13 @@ dev
 - [bugfix] Override display of DEFAULT_PYTHON value when generating web documentation for `pipx install` #523
 - [bugfix] Wrap help documentation for environment variables.
 - [bugfix] Fixed uninstall crash that could happen on Windows for certain packages
-- [feature] Venv package name arguments now do not have to match exactly as pipx has them stored, but can be specified in any python-package-name-equivalent way. (i.e. case does not matter, and any number of `.`, `-`, or `_` characters in a row are equivalent.)
+- [feature] Venv package name arguments now do not have to match exactly as pipx has them stored, but can be specified in any python-package-name-equivalent way. (i.e. case does not matter, and `.`, `-`, `_` characters are interchangeable.)
 - [change] Venvs with a suffix: A suffix can contain any characters, but for purposes of uniqueness, python package name rules apply--upper- and lower-case letters are equivalent, and any number of `.`, `-`, or `_` characters in a row are equivalent.  (e.g. if you have a suffixed venv `pylint_1.0A` you could not add another suffixed venv called `pylint--1-0a`, as it would not be a unique name.)
 
 0.15.6.0
 
 - [docs] Update license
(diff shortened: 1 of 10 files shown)


cs01:
> I would've just canonicalized the entire venv_dir argument, but unfortunately this is not compatible with suffixed venvs, as it is currently possible for them to have non-canonical characters. 

Nice, great attention to detail

> One possible change that could fix this would be to also canonicalize the suffix recorded as venv directory name. I'd be interested to know if there is interest in that approach. The suffix could still remain the same, but the suffixed venv directory would be canonicalized.

This seems like the right approach to me.

cs01 (approved):

itsayellow:
> > One possible change that could fix this would be to also canonicalize the suffix recorded as venv directory name. I'd be interested to know if there is interest in that approach. The suffix could still remain the same, but the suffixed venv directory would be canonicalized.
> 
> This seems like the right approach to me.

Maybe I'll see if I can fold that into this PR.  It would make things simpler and cleaner.

PR #543, opened 2020-11-10
Add isort to lint nox session, fix import order inside pipx files

### LINKED ISSUE #257: add isort to lint tests

itsayellow:

## Summary of changes
This adds isort checking to our lint nox session.  If imports are not sorted properly, the lint session will return a diff instructing the user how they need to rearrange their import statements.

I corrected the import order in all python files via isort .  I also changed the couple of files with relative imports to use absolute imports.

I added the `scripts/` subdirectory to lint checking.

Question for @cs01: I noticed that `setup.py` imports both `re` and `ast` and it seems like they are never used.  Am I missing something?  Are they actually used somehow?  If not I'll remove them.

Fixes #257 

## Test plan

Tested by running
```
```


cs01:
I tried to use isort before but I remember it was “unstable”. Each time I ran it, it sorted things differently so I gave up. Maybe it’s improved since then or you have a better configuration for it.

cs01 (approved):
Lgtm besides the unused imports in setup.py

itsayellow:
> I tried to use isort before but I remember it was “unstable”. Each time I ran it, it sorted things differently so I gave up. Maybe it’s improved since then or you have a better configuration for it.

I've been using it for months now with pre-commit, and its been sorting all of my commits.  So far so good for me, it seems stable.  I haven't noticed any strange sorting behavior.

itsayellow:
> If they aren’t directly used in the file then they aren’t used anymore. Must have been something in the past that got removed that used them.

I finally tracked them down.  They used to be used in `get_version()`

PR #545, opened 2020-11-11
Make shared libraries create, upgrade not accept pip args

### LINKED ISSUE #544: Should installing pipx shared libs use `pip_args` from the latest package?
**Describe the bug**

Currently when installing or upgrading the pipx shared libs, the functions `_SharedLibs.upgrade()` and `_SharedLibs.create()` accept `pip_args` arguments. 
https://github.com/pipxproject/pipx/blob/08bd15610bc406dabe8bf0db4adaba4a3a8b89ea/src/pipx/shared_libs.py#L33
https://github.com/pipxproject/pipx/blob/08bd15610bc406dabe8bf0db4adaba4a3a8b89ea/src/pipx/shared_libs.py#L61

They are used if a normal `pipx install` needs a shared library create/upgrade, and the `pipx install` pip args are used to also create the shared libs.

They are NOT used when creating/upgrading the shared libs during a `pipx list` or any access of an existing venv.  (!!)

Recently this failed for me in the following situation:  In a Python 3.6 test I was trying to install a package for the first time using the pip arguments `--use-feature=2020-resolver`.  It caused pipx to fail with an error because before installing the package, it went to install the shared libs using pip argument `--use-feature=2020-resolver`.  Since the original Python 3.6 pip doesn't have that option, it couldn't even upgrade pip in the first place and crashed.

I'm wondering if we should just remove `pip_args` from `_SharedLibs.upgrade()` and `_SharedLibs.create()`.   It seems kind of arbitrary that only when upgrading the shared libs during a package install do they get the `pip_args` that happen to be used for that package, but not otherwise.

Is there a reason a user would need custom pip args to be used with the installation of the shared libs?  If so, maybe we need a different more reliable mechanism.

**How to reproduce**

**Expected behavior**

itsayellow:

## Summary of changes
Closes #544 

Changed the shared libs `create` and `upgrade` functions to no longer accept `pip_args`, which used to be applied inconsistently, sometimes taken from a recent `pipx install`, sometimes not at all.  Now they are never applied.  There are a lot of pip arguments that can even break the installation and/or upgrade of shared libraries (see example in #544).

It is assumed that `PIP_` environment variables can be used if pip behavior in pipx needs to be controlled consistently.

Also changed in `create()`:
```
pip install --upgrade --ignore-installed pip setuptools wheel
```
to
```
pip install --upgrade --force-reinstall pip setuptools wheel
```
As it seems safer to force a reinstall than just to ignore the old libraries.  This code is meant to force a fetch of a fresh pip from pypi to ignore any odd system custom pip version.

Incidental change: on the verbose display of shared libraries age in seconds, I removed the display of fractional seconds! 😄 

## Test plan

Tested by running
```
```

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -3,12 +3,12 @@ dev
 - Added reinstall command for reinstalling a single venv.
 - Changed `pipx run` on non-Windows systems to actually replace pipx process with the app process instead of running it as a subprocess.  (Now using python's `os.exec*`)
 - [bugfix] Fixed bug with reinstall-all command when package have been installed using a specifier. Now the initial specifier is used.
-- [bugfix] Override display of DEFAULT_PYTHON value when generating web documentation for `pipx install` #523
+- [bugfix] Override display of `PIPX_DEFAULT_PYTHON` value when generating web documentation for `pipx install` #523
 - [bugfix] Wrap help documentation for environment variables.
 - [bugfix] Fixed uninstall crash that could happen on Windows for certain packages
 - [feature] Venv package name arguments now do not have to match exactly as pipx has them stored, but can be specified in any python-package-name-equivalent way. (i.e. case does not matter, and `.`, `-`, `_` characters are interchangeable.)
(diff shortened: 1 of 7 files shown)


itsayellow:
Also!  I modified one and disabled another test involving `kaggle` temporarily, because without `2020-resolver` some dependencies of `kaggle` can cause pip installation to fail due to version conflicts.

itsayellow:
Ugh, urllib3 1.26 was released and it seems to break a lot of test packages.  I might just wait until pip 20.3 arrives (it should be soon, right?) and the 2020-resolver is default before trying to merge this.  The tests should hopefully work then.

PR #549, opened 2020-11-12
GitHub Workflow: Publish to PyPI

### LINKED ISSUE #522: Automated PyPI release using Github Actions on Github release

Proposing a one-stop release process that is script-based using Github Actions.  This would ensure consistency, make releases easier, and prevent user release errors.

The trigger would be a Github release, which would then instigate a PyPI release.

Using this as base instructions, to be modified for our use case:
https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

@cs01, one thing we would need to automate PyPI release is a new Github Secret for repository pipx, which would be a PyPI API token.  Currently I don't have access to the Secrets settings for repo pipx, either you'd need to make a secret or make that repo setting available to pipx-maintainers.

Otherwise, manual like usual.

itsayellow:

## Summary of changes
Closes #522 

Add a PyPI Publish job that executes automatically when both a git push and git tag have been made.  This follows both the [PyPA example](https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/) and @gaborbernat's example code he linked to in #522 .

In addition I added a Test PyPI Publish job that only executes manually through github's web interface using `workflow_dispatch`.  It contains the same code as the regular PyPI Publish job, except publishing to Test PyPI.  I thought this should be useful for debugging before trying a real PyPI publish.

I'm hoping to make a release using this method once this PR is merged.

I wasn't thrilled about putting the PyPI Publish in the same workflow file as our normal "on push" tests, because now "Publish pipx to PyPI on release" shows up as a "skipped" job every time we push normally.  But I also wanted to make publishing to PyPI depend on our normal tests.  To accomplish this I could either do as I've done here, or duplicate the test code in a separate "on-publish.yml" file.  I didn't like the idea of duplicating code that might not inherit fixes and changes in the normal CI on-push code, so I put them all together in the same file like @gaborbernat's example does.

*EDIT:*
#### noxfile.py

Revamped `noxfile.py` to use python style for globals (ALL_CAPS) and reorganized sessions so they are all together in that file.

Added `PYTHON_DEFAULT_VERSION` global to ease changing the python version used for single-version sessions.

Renamed nox session `docs` to `build_docs`.

Added `pip install --upgrade pip` to critical nox sessions, because the lack of this was a problem for me at least once in the past for me when executing locally on my computer.

`publish_docs` now runs session `build_docs` instead of duplicating the doc build code.

#### Github Workflows

Removed `on: release` from `on-push.yml` because it generates 3 separate extra/redundant triggers (for release: `created`, `released`, `published`) in addition to the trigger for push that happens anyway on release.

Created `env: default-python` variable as a better way of defining a .yml "constant" instead of a single-value matrix like before.  Unfortunately using env variables is limited to inside of `name` or `with` tags, so it cannot be used inside the `tests:strategy:matrix:include` definitions.

Combined `Upgrade pip` and `Install nox` into one "step".

Renamed github workflow job `docs` to `verify-docs`.

#### docs

Updated `docs/contributing.md` with current nox sessions.

Bumped examples to show use of Python 3.9 instead of 3.7.

## Test plan

Tested by running
```
```


gaborbernat:
While trying to figure out how we build the package itself I discovered we don't have a ``pyproject.toml`` file where this is specified... Can we fix this, please?

gaborbernat:
> If you're asking just to use `pyproject.toml` instead of `setup.py` for pipx, this sounds like a good separate issue to me. I'm thinking that can be addressed later, independently of this PR.

Well, yes and no. I think the package to release should be built with https://github.com/pypa/build as that adheres to PEP-518, which we should follow. And that requires this.

itsayellow:
> > If you're asking just to use `pyproject.toml` instead of `setup.py` for pipx, this sounds like a good separate issue to me. I'm thinking that can be addressed later, independently of this PR.
> 
> Well, yes and no. I think the package to release should be built with https://github.com/pypa/build as that adheres to PEP-518, which we should follow. And that requires this.

That's what I meant as well.

I'm assuming it can be done in a separate issue?  Right now I'm trying to make a pipx release sooner rather than later.

PR #551, opened 2020-11-13
Convert pipx to use PEP 517

### LINKED ISSUE #550: Use `pyproject.toml` for pipx package definition primarily instead of setup.py
While trying to figure out how we build the package itself I discovered we don't have a ``pyproject.toml`` file where this is specified... Can we fix this, please?

_Originally posted by @gaborbernat in https://github.com/pipxproject/pipx/issues/549#issuecomment-726673524_

uranusjr:

Fix #550. I manually checked the generated metadata and everything seems identical, except some delibrate changes:

* I added RFC-822 format to `author-email`. This is supported, and is more friendly to email clients.
* The `Requires-Python` metadata is added since the cfg format can safely ignore unknown keys in older setuptools.
* I added the `Python :: 3.9` trove classifier because why not 🙂 

noxfile is updated to build with the new PyPA PEP 517 [build](https://github.com/pypa/build/) tool instead of calling `setup.py` directly.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -94,11 +94,10 @@ def develop(session):
 
 @nox.session(python="3.8")
 def build(session):
-    session.install("setuptools")
-    session.install("wheel")
+    session.install("build")
     session.install("twine")
     session.run("rm", "-rf", "dist", "build", external=True)
-    session.run("python", "setup.py", "--quiet", "sdist", "bdist_wheel")
+    session.run("python", "-m", "build")
 
 
 def has_changes():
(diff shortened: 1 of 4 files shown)


itsayellow:
I've been meaning to get around to this, thanks so much @uranusjr for the PR.

itsayellow:
Question: will the new `setup.py` still allow for editable installs?

uranusjr:
Yes, setuptools automatically backfills setup.cfg fields back to `setup()`.

itsayellow:
Is there a good reference for how to write a declarative `setup.cfg`, and how all of the directives?  All I could find was this:
https://setuptools.readthedocs.io/en/latest/userguide/declarative_config.html

uranusjr:
That’s the main source of documentation I know as well (plus reading configuration from other projects like setuptools and wheel).

itsayellow (approved):
So modern. 👏

PR #560, opened 2020-11-19
Add and fix pipx exit codes

itsayellow:

## Summary of changes

pipx had erroneous or uncoded exit codes in the past.  `main.py` expected exit codes from the commands, but almost none of them actually returned anything, and `upgrade` returned the wrong exit codes.  This PR attempts to fix all of that.

I made some "consts" names for the error codes in `constants.py`.  I gave unique names to the error conditions, and just `EXIT_CODE_OK = 0` for the non-error condition. 

Here are the error exit code == 1 conditions.  Assume all other outcomes give an exit code of `0` for no error.
| Command       | Exit Situation                   | Exit Code | Notes |
| ------------- | -------------------------------- | --------- | ----- |
| install       | venv already exists              | 1         |       |
| install       | installation error (`PipxError`) | 1         |       |
| inject        | installation error (`PipxError`) | 1         |       |
| upgrade       | upgrade error (`PipxError`)      | 1         |       |
| upgrade-all   | upgrade error in any package     | 1         | `PipxError`s are first caught for each package, then one is raised at end |
| uninstall     | venv doesn't exist               | 1         |       |
| uninstall-all | uninstall error in any package   | 1         |       |
| reinstall     | venv doesn't exist               | 1         |       |
| reinstall     | install error (`PipxError`)      | 1         |       |
| reinstall-all | reinstall error in any package   | 1         |       |
| list          | Any package has "problems"       | 1         | Problems include: missing python interpreter, missing metadata, no main package version (not installed) |
| run           |                                  | \<whatever exit code command returns\> | |
| runpip        |                                | \<whatever exit code `pip` command returns\> | |

`ensurepath` and `completions` always return exit code 0 for no error.

### Other changes
Modified `pipx runpip` to actually return the `pip` exit code.

Modified `pipx list` to make shorter per-package summaries of problems, and at the end of list make helpful messages on how you might fix the problems.  (Problems include: missing python interpreter, missing metadata, no main package version (not installed))  For example:
```
> pipx list
venvs are in /Users/mclapp/.local/pipx/venvs
apps are exposed on your $PATH at /Users/mclapp/.local/bin
   package finddup 0.1, Python 3.9.0
    - finddup
   package isort 5.6.4, Python 3.9.0
    - isort
   package kaggle 1.5.9, Python 3.9.0
    - kaggle
   package nox 2020.8.22, Python 3.9.0
    - nox
    - tox-to-nox
   package pytivo 0.2.0, Python 3.9.0
    - pytivo
   package vulture has missing internal pipx metadata.
   package zipls 0.2, Python 3.9.0
    - zipls

One or more packages have a missing internal pipx metadata.
   They were likely installed using a pipx version before 0.15.0.0.
   Please uninstall and install these package(s) manually to fix.
```

## Test plan

Tested by running
```
```


gaborbernat (approved):

itsayellow:
@gaborbernat thanks so much for quick review!  Sorry to bother with another commit, but I thought it would be useful.

The new commit just creates a `NewType` of `ExitCode`, so we can type-check that an exit code is being passed around instead of a generic int.  This actually happened before with upgrade, upgrade returned an int that was being used for other purposes, but it got interpreted as an exit code in `main.py`.

Let me know if you think this is not a good idea or implemented in the wrong way.

PR #566, opened 2020-11-26
Implement rotating file logging, log more info to file.

### LINKED ISSUE #274: Log Exceptions, Errors to file with traceback info
We catch any `Exception` in `install()` but are silent about it--there is no way to know what happened.

For debugging user bugs, this would be extremely helpful (see #273 for example.)

Save the caught exception using something like `except Exception as e` and then use either a output what it is using plain `repr(e)` or more sophisticated printout using the `traceback` library for example.  We could do this only if `pipx --verbose` is on.

None.

I think any more information than the current "no information" would be helpful

itsayellow:

## Summary of changes
Closes #274 

This adds a subdirectory `logs/` to `PIPX_HOME`.  Each pipx command run gets a new logfile written to this directory with a datestamp in the filename.  The 10 most recent log files are kept and others are deleted.

`logging.DEBUG` log level and above are logged to these files.  The console output logging is `logging.INFO` and above for `--verbose` mode, and `logging.WARNING` and above for non-verbose mode.

We used to have no `logging.debug()` logging statements, and I added a lot to capture the stack trace for various errors that we catch and may not print out fully to the console

Since these logs do **not** need to be turned on manually (i.e. with `--verbose`), it should allow a very verbose log to be available for one-off or non-replicable errors.

I implemented the file rotation myself, because trying to use `logging.handlers.RotatingFileHandler` and manually initiating a `doRollover()` for each new command caused file permission errors on Windows.

## Test plan

After running 3 pipx commands:
```shell
> ls ~/.local/pipx/logs
cmd_2020-11-25_19.26.39.log  cmd_2020-11-25_19.26.58.log
cmd_2020-11-25_19.26.53.log
> 
```

### DIFF
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -723,7 +723,7 @@ def cli() -> ExitCode:
         return run_pipx_command(parsed_pipx_args)
     except PipxError as e:
         print(str(e), file=sys.stderr)
-        logging.debug(f"PipxError: {e}")
+        logging.debug(f"PipxError: {e}", exc_info=True)
         return ExitCode(1)
     except KeyboardInterrupt:
         return ExitCode(1)
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -616,33 +616,29 @@ def get_command_parser() -> argparse.ArgumentParser:
     return parser
 
 
-def setup(args: argparse.Namespace) -> None:
-    if "version" in args and args.version:
-        print_version()
-        sys.exit(0)
-
-    root_logger = logging.getLogger()
-    root_logger.setLevel(logging.DEBUG)
-
+def setup_file_logging(logger: logging.Logger) -> None:
     # don't use utils.mkdir, to prevent emit of log message
     constants.PIPX_LOG_DIR.mkdir(parents=True, exist_ok=True)
-    file_handler = RotatingFileHandler(
-        constants.PIPX_LOG_DIR / "command.log", backupCount=5
-    )
+    log_file = constants.PIPX_LOG_DIR / "command.log"
+    log_file_exists = log_file.exists()
+    file_handler = RotatingFileHandler(log_file, backupCount=5)
     file_handler.setLevel(logging.DEBUG)
-    file_handler.doRollover()
+    if log_file_exists:
+        file_handler.doRollover()
     file_handler.setFormatter(
         logging.Formatter(
             "{relativeCreated: >8.1f}ms ({funcName}:{lineno}): {message}", style="{"
         )
     )
-    root_logger.addHandler(file_handler)
+    logger.addHandler(file_handler)
     # log these infos only to file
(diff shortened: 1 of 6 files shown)


src/pipx/main.py
```diff
@@ -614,19 +616,68 @@ def get_command_parser() -> argparse.ArgumentParser:
...
+            try:
+                existing_log.unlink()
+                print(f"Removed file {existing_log}", file=sys.stderr)
+            except FileNotFoundError:
+                pass
+
+    datetime_str = time.strftime("%Y-%m-%d_%H.%M.%S")
+    log_file = constants.PIPX_LOG_DIR / f"cmd_{datetime_str}.log"
+    counter = 1
+    while log_file.exists():
```
cs01, line 636: Suggest adding a `and counter < max_logs` to be defensive. While loops always scare me.

src/pipx/main.py
```diff
@@ -614,19 +616,68 @@ def get_command_parser() -> argparse.ArgumentParser:
     return parser
 
 
+def setup_file_handler() -> logging.Handler:
```
cs01, line 619: I have not found the built in python logging utilities to be too user friendly, but I wonder if there is something built in to it that automatically rotates files and takes care of the dirty work for you. Here are a couple things I found.

https://www.blog.pythonlibrary.org/2014/02/11/python-how-to-create-rotating-logs/
https://docs.python.org/3/library/logging.handlers.html#baserotatinghandler
https://docs.python.org/3/library/logging.handlers.html#logging.handlers.RotatingFileHandler

src/pipx/main.py
```diff
@@ -672,9 +723,13 @@ def cli() -> ExitCode:
         return run_pipx_command(parsed_pipx_args)
     except PipxError as e:
         print(str(e), file=sys.stderr)
+        logging.debug(f"PipxError: {e}", exc_info=True)
         return ExitCode(1)
     except KeyboardInterrupt:
         return ExitCode(1)
+    except Exception:
+        logging.debug("Uncaught Exception:", exc_info=True)
```
cs01, line 731: Nice! All these extra debug logs will be very useful.

src/pipx/main.py
itsayellow: I mentioned that in the blurb for this PR. I started using `RotatingFileHandler`.  It's not a perfect fit.  It wants to keep appending to the current file until it gets too big at which point it rolls all the files.  I wanted a separate log file for every command invocation.

I managed to used it anyway in the beginning, by manualling initiating a file roll with `doRollover()`, but this caused a crash on Windows, because by creating the handler it opened the log file, and then manually trying to rename that open file made Windows upset.

cs01: Whoops sorry about that. I see it now. Custom implementation sounds good to me!

gaborbernat (approved):

PR #567, opened 2020-11-26
update maintainer list

cs01:
Maintainer list was out of date

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ For more details, see the [installation instructions](https://pipxproject.github
 
 pipx is a tool to help you install and run end-user applications written in Python. It's roughly similar to macOS's `brew`, JavaScript's [npx](https://medium.com/@maybekatz/introducing-npx-an-npm-package-runner-55f7d4bd282b), and Linux's `apt`.
 
-It's closely related to pip. In fact, it uses pip, but is focused on installing and managing Python packages that can be run from the command line directly as applications. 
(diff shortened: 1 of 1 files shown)


itsayellow:
@cs01, I thought we were both lead maintainers?

gaborbernat (approved):

cs01:
> @cs01, I thought we were both lead maintainers?

We are.

@gaborbernat thanks for the reviews. Could you please not merge them after approving?

gaborbernat:
Ah, my bad. I'd expect in that case to be marked as draft PR.

cs01:
> Ah, my bad. I'd expect in that case to be marked as draft PR.

That's a good call. I am trying to figure out the right expectations around when things should be merged or not. I personally like to merge my own PRs, but it probably varies from person to person.

cs01:
We have a thing at my job where if you add the hashtag “#shipit” to the summary, and only then, does a reviewer have the ability to merge a PR. To me, marking as draft means “I want CI results, but don’t want anyone to review this”. I would like to adopt those conventions for pipx too. What do you think, does that work? Any thoughts or preferences @itsayellow?

gaborbernat:
Sure. 👍

itsayellow:
Yeah that sounds good to me too.

itsayellow:
Yeah when I make something a Draft, in my mind it's so people can see what I'm doing and comment, but with the understanding that I'm not done yet.  But yes even if it's not a draft, I like the convention you're proposing.  Let's do that.

PR #571, opened 2020-11-28
Docs: Add a section in  Contributing detailing how to use pre-commit with pipx

itsayellow:

## Summary of changes
Added a section to the docs in "Contributing" that recommends and describes how to use `pre-commit` with pipx.

## Test plan

Tested by running
```
```

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -7,6 +7,29 @@ To run the pipx executable from your source tree during development, run pipx fr
 python src/pipx --version
 ```
 
+## Pre-commit
+The use of [pre-commit](https://pre-commit.com/) is recommended.  It can show
+all and fix some lint errors before commit, saving you the trouble of finding
+out later that it failed CI Lint errors, and saving you from having to run `nox
+-s lint` separately.
+
+In the pipx git repository is a `.pre-commit-config.yaml` configuration file
+tailored just for pipx and its lint requirements.  To use pre-commit in your
+clone of the pipx repository, you need to do the following **one-time setup
+procedure**:
+
+1. Install pre-commit as described by https://pre-commit.com/#installation
+2. In the top level directory of your clone of the pipx repository, execute `pre-commit install`
+
+Afterwards whenever you commit in this repository, it will first run pipx's
+personalized lint checks.  If it makes a fix to a file (e.g. using `black` or
+`isort`), you will need to `git add` that file again before committing it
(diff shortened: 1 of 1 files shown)


docs/contributing.md
```diff
@@ -7,6 +7,29 @@ To run the pipx executable from your source tree during development, run pipx fr
...
+all and fix some lint errors before commit, saving you the trouble of finding
+out later that it failed CI Lint errors, and saving you from having to run `nox
+-s lint` separately.
+
+In the pipx git repository is a `.pre-commit-config.yaml` configuration file
+tailored just for pipx and its lint requirements.  To use pre-commit in your
+clone of the pipx repository, you need to do the following **one-time setup
+procedure**:
+
+1. Install pre-commit as described by https://pre-commit.com/#installation
```
cs01, line 21: Do you use pipx to install pre commit?

itsayellow: Hmmm, good question.  I use Homebrew because I'm on a mac.  I think it should work on pipx though!  I'll have to experiment.

itsayellow: Yeah that's nice, you can totally use pipx.  We should make a PR to `pre-commit` to suggest pipx as an install method!

I'll fix the docs to talk about pipx install, good call.

itsayellow: I thought about adding `brew install pre-commit` as an option, but in the end I figured I would keep it simple.  The only reason I might use Homebrew is that if you're messing with pipx and somehow screw it up, then you aren't possibly making pre-commit nonfunctional.  But this probably is a low probability event.

Advanced users looking for other ways to install can always go to the linked pre-commit webpage.

itsayellow: I just filed an issue to ask pre-commit if they'd want to list us as an installation option.  He shut it down as a duplicate of a PR that was also shut down.  😄  I won't link to it here, but it was a pre-commit PR from 6 days ago.

At least we have other users who are evangelizing pipx!

cs01: The reasoning he mentioned in the linked issue is

> it's not that I don't see value, it's that anything that works for pip also works for pipx so there's no point in putting that in the docs here

which I don't understand. If there is value in using pipx over pip, why wouldn't he want to share that? 

Anyway, sounds like pipx is a valid way to install it so it would be okay if we mention it in pipx's docs.

itsayellow: Agreed on all counts.  I've already added it to the PR here.

PR #569, opened 2020-11-28
New Troubleshooting webpage in docs

itsayellow:
* [x] I have added an entry to `docs/changelog.md`

## Summary of changes
Added a "Troubleshooting" page with common solutions to common problems.  Also added a link to the Troubleshooting page to the GitHub bug report template urging users to check it first.

I looked through the past year or so of Issues to try and find common user problems that were easily remedied (and not actual pipx bugs.)

Hopefully this will save time on the easiest bug reports and at least give a webpage to link to for answers to common problems.

## Test plan

Tested by running
```
# command(s) to exercise these changes
```

PR #573, opened 2020-11-29
Fix troubleshooting test_venv.

itsayellow:

## Summary of changes
Follow-up to PR #569 .  Fixing the "Try it with pip" section to not include activate / deactivate, and specify different Windows instructions, as @cs01 recommended.

## Test plan

Tested by running
```
```

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -87,10 +87,16 @@ This is a tip for advanced users.  An easy way to check if pipx is the problem
 or a package you're trying to install is the problem, is to try installing it
 using `pip`.  For example:
 
+### Unix or macOS
 ```
 python3 -m venv test_venv
-source test_venv/bin/activate
-python3 -m pip install <problem-package>
+test_venv/bin/python3 -m pip install <problem-package>
+```
+
+### Windows
+```
+python -m venv test_venv
+test_venv/Scripts/python -m pip install <problem-package>
 ```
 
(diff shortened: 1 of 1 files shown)


docs/troubleshooting.md
```diff
@@ -87,10 +87,16 @@ This is a tip for advanced users.  An easy way to check if pipx is the problem
...
 python3 -m venv test_venv
-source test_venv/bin/activate
-python3 -m pip install <problem-package>
+test_venv/bin/python3 -m pip install <problem-package>
+```
+
+### Windows
+```
+python -m venv test_venv
+test_venv/Scripts/python -m pip install <problem-package>
```
cs01, line 99: Looks great! I'm not sure, but I think one windows it needs to be `python.exe`?

itsayellow: I believe I actually tried on my VM Windows...but now you make me want to make sure and try it again 😆 

I think like regular entry points on windows, you don't need to type the `.exe`, but I will check again.

itsayellow: Yes it works.  😄  Also I remember now that I had to change `python3` to `python` on Windows and I found this out by trying these commands in Windows.

itsayellow:
@cs01, is this good to merge?

cs01:
> @cs01, is this good to merge?

Yup looks good to me!

gaborbernat (approved):

PR #576, opened 2020-12-01
Fix tests on Windows by using `python -m install --upgrade pip`

itsayellow:

## Summary of changes
Somehow just now (?) Windows tests broke because we are using `pip install --upgrade pip`.  Windows can't upgrade an in-use pip which is open.

Switched to the better form `python -m install --upgrade pip`.

## Test plan

Tested by running
```
```

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -80,7 +80,7 @@ def on_master_no_changes(session):
 
 @nox.session(python=PYTHON_ALL_VERSIONS)
 def tests(session):
-    session.install("--upgrade", "pip")
+    session.run("python", "-m", "pip", "install", "--upgrade", "pip")
     prebuild_wheels(session, PREBUILD_PACKAGES)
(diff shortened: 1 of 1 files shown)


noxfile.py
```diff
@@ -80,7 +80,7 @@ def on_master_no_changes(session):
 
 @nox.session(python=PYTHON_ALL_VERSIONS)
 def tests(session):
-    session.install("--upgrade", "pip")
+    session.run("python", "-m", "pip", "install", "--upgrade", "pip")
```
cs01, line 83: Should we file an issue on the nox project to switch to this?

itsayellow: Hmm maybe.  It only is a problem for windows and `pip`.

I can file an issue.

itsayellow: Looks like nox is about to merge a PR that does this very thing. :)

PR #583, opened 2020-12-09
Fix and improve logging

### LINKED ISSUE #581: Using the root logger badly interacts with pytest
**Describe the bug**

Because pipx uses the root logger, this causes problems with pytest.

Specifically, in pytest the root logger persists between tests, so every time we run another pipx test in the same pytest session, our `setup`-added handlers keep accumulating.  This causes the same message to be printed to STDERR many times over (multiple stream handlers).  It also causes attempts to print the message to old files that are now closed, which adds many logging errors (one error for each closed log file from a previous test.)

To fix I'm going to use the `logger = logging.getLogger(__name__)` syntax like we should've been doing all along.

**How to reproduce**

**Expected behavior**

itsayellow:

## Summary of changes
Closes #581 

The main problem that this PR solves is that previously during a pytest session, because `logging` was never getting unloaded, we kept adding logging Handlers to the root logger which were never removed.  As the tests went on, we ended up with many, many stream handlers creating redundant messages, and many file handlers trying (unsuccessfully) to write to closed log files.

This new code uses `logging.config.dictConfig` to setup loggers, handlers, and formatters.  This config API resets all previous logging state (this is what `incremental=False` does).  So at the end of every invocation of `setup_logging`, even in pytest, the result is only 2 handlers attached to the pipx logger.  I've debugged and verified that this is so.

Most of the changes here change our code from using the root logger to using a "pipx" logger.  i.e. using `logger = logging.getLogger(__name__)` at the top of the files and e.g. `logger.info()` instead of `logging.info()` to use the "pipx" logger everywhere and not the root logger.  This allows us to effectively manage our own logger separately from the root logger.  In pytest, the root logger has special pytest handlers added to it.

The other interesting change affects `util.py`, `venv.py`, and `shared_libs.py`.  We now fully log `run_subprocess`, adding `stdout`, `stderr`, and `returncode` to the file logs for each subprocess invocation for all `std*` are captured.  Additionally, to capture more info to the file logs, I removed `run_verify()` (which didn't capture stderr and stdout, but just let them go directly to the console) and replaced it with a call to `run_subprocess` that both captures stderr and stdout and echoes to the console on error.  I tried to replicate the old user-visible behavior, but now allowing capturing subprocess output to the logs in addition.  I split the "verify" part into a separate function `subprocess_post_check()` so that any output could be printed outside of the `animate` context to not screw up the text animation.

I think the main change in behavior from a user standpoint is that piping `pip` errors through pipx to the console now displays errors as plain text color instead of `pip`'s red text color.  I thought that was ok (even better?) because now we get to log the error messages to file.

## Test plan

Tested by running
```
```

### DIFF
--- a/src/pipx/commands/common.py
+++ b/src/pipx/commands/common.py
@@ -20,6 +20,8 @@ from pipx.pipx_metadata_file import PackageInfo
 from pipx.util import WINDOWS, PipxError, mkdir, rmdir
 from pipx.venv import Venv
 
+logger = logging.getLogger(__name__)
+
 
 class VenvProblems:
     def __init__(
@@ -89,7 +91,7 @@ def _copy_package_apps(
         if not dest.parent.is_dir():
             mkdir(dest.parent)
         if dest.exists():
-            logging.warning(f"{hazard}  Overwriting file {str(dest)} with {str(src)}")
+            logger.warning(f"{hazard}  Overwriting file {str(dest)} with {str(src)}")
             dest.unlink()
         if src.exists():
             shutil.copy(src, dest)
@@ -106,7 +108,7 @@ def _symlink_package_apps(
             mkdir(symlink_path.parent)
 
         if force:
-            logging.info(f"Force is true. Removing {str(symlink_path)}.")
+            logger.info(f"Force is true. Removing {str(symlink_path)}.")
             try:
                 symlink_path.unlink()
             except FileNotFoundError:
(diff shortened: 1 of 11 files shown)


itsayellow:
This is an enabler for my last PR for this revision cycle that I'm working on finishing, to switch `venv_metadata_inspector.py` from using `pkg_resources` to using `importlib.metadata` and `packaging`.  To really test the new metadata inspection I made some huge optional tests that install many many packages.  Those very large tests exposed the problems with our current logging methodology, when they actually exhausted system memory and were shut down by the OS.

gaborbernat (approved):

PR #588, opened 2020-12-18
Overhaul of `venv_metadata_inspector.py`, now using modern python libraries

### LINKED ISSUE #528: [Windows] Unable to inject jaraco.clipboard
**Describe the bug**
Today I'm upgrading my Python 3.8 envs with Python 3.9. `reinstall-all` failed (found executables missing), so I uninstalled manually and attempted to reinstall. After `pipx install 'xonsh[ptk]'`, I attempted to inject two deps:

```
PS C:\WINDOWS\system32> pipx inject xonsh jaraco.clipboard
Internal error with venv metadata inspection.
Unable to install jaraco-clipboard.
Check the name or spec for errors, and verify that it can be installed with pip.
←[?25h
PS C:\WINDOWS\system32> pipx inject xonsh keyring
  injected package keyring into venv xonsh
done!
←[?25h
```

With `--verbose`:

```
PS C:\WINDOWS\system32> pipx inject --verbose xonsh jaraco.clipboard
←[?25lpipx > (setup:567): pipx version is 0.15.6.0
pipx > (setup:568): Default python interpreter is 'c:\\program files\\python39\\python.exe'
pipx > (run_pipx_command:135): Virtual Environment location is C:\Users\jaraco\.local\pipx\venvs\xonsh
pipx > (needs_upgrade:54): Time since last upgrade of shared libs, in seconds: 238598.75953483582. Upgrade will be run by pipx if greater than 2592000.0.
pipx > (package_name_from_spec:247): Determined package name: jaraco-clipboard
pipx > (package_name_from_spec:248): Package name determined in 0.0s
pipx > (_parsed_package_to_package_or_url:127): cleaned package spec: jaraco-clipboard
installing jaraco-clipboard...
pipx > (run_subprocess:113): running C:\Users\jaraco\.local\pipx\venvs\xonsh\Scripts\python.exe -m pip install jaraco-clipboard
Requirement already satisfied: jaraco-clipboard in c:\users\jaraco\.local\pipx\venvs\xonsh\lib\site-packages (2.0.1)
Requirement already satisfied: jaraco.windows>=3.4; sys_platform == "win32" in c:\users\jaraco\.local\pipx\venvs\xonsh\lib\site-packages (from jaraco-clipboard) (5.0.0)
Requirement already satisfied: six in c:\users\jaraco\.local\pipx\venvs\xonsh\lib\site-packages (from jaraco.windows>=3.4; sys_platform == "win32"->jaraco-clipboard) (1.15.0)
Requirement already satisfied: jaraco.text in c:\users\jaraco\.local\pipx\venvs\xonsh\lib\site-packages (from jaraco.windows>=3.4; sys_platform == "win32"->jaraco-clipboard) (3.2.0)
Requirement already satisfied: jaraco.ui in c:\users\jaraco\.local\pipx\venvs\xonsh\lib\site-packages (from jaraco.windows>=3.4; sys_platform == "win32"->jaraco-clipboard) (2.0.1)
Requirement already satisfied: jaraco.collections in c:\users\jaraco\.local\pipx\venvs\xonsh\lib\site-packages (from jaraco.windows>=3.4; sys_platform == "win32"->jaraco-clipboard) (3.0.0)
Requirement already satisfied: path.py>=6.2 in c:\users\jaraco\.local\pipx\venvs\xonsh\lib\site-packages (from jaraco.windows>=3.4; sys_platform == "win32"->jaraco-clipboard) (12.5.0)
Requirement already satisfied: jaraco.structures>=1.1.1 in c:\users\jaraco\.local\pipx\venvs\xonsh\lib\site-packages (from jaraco.windows>=3.4; sys_platform == "win32"->jaraco-clipboard) (2.0)
Requirement already satisfied: more-itertools in c:\users\jaraco\.local\pipx\venvs\xonsh\lib\site-packages (from jaraco.windows>=3.4; sys_platform == "win32"->jaraco-clipboard) (8.5.0)
Requirement already satisfied: jaraco.functools in c:\users\jaraco\.local\pipx\venvs\xonsh\lib\site-packages (from jaraco.text->jaraco.windows>=3.4; sys_platform == "win32"->jaraco-clipboard) (3.0.1)
Requirement already satisfied: jaraco.classes in c:\users\jaraco\.local\pipx\venvs\xonsh\lib\site-packages (from jaraco.ui->jaraco.windows>=3.4; sys_platform == "win32"->jaraco-clipboard) (3.1.0)
Requirement already satisfied: path in c:\users\jaraco\.local\pipx\venvs\xonsh\lib\site-packages (from path.py>=6.2->jaraco.windows>=3.4; sys_platform == "win32"->jaraco-clipboard) (15.0.0)
pipx > (run_subprocess:113): running C:\Users\jaraco\.local\pipx\venvs\xonsh\Scripts\python.exe -c <contents of venv_metadata_inspector.py> jaraco-clipboard C:\Users\jaraco\.local\pipx\venvs\xonsh\Scripts
pipx > (get_venv_metadata_for_package:282): Internal error with venv metadata inspection.
pipx > (get_venv_metadata_for_package:283): venv_metadata_inspector.py Traceback:
  File "<string>", line 159, in <module>
  File "<string>", line 120, in main
  File "<string>", line 37, in get_apps
  File "C:\Users\jaraco\.local\pipx\shared\Lib\site-packages\pkg_resources\__init__.py", line 465, in get_distribution
    dist = get_provider(dist)
  File "C:\Users\jaraco\.local\pipx\shared\Lib\site-packages\pkg_resources\__init__.py", line 341, in get_provider
    return working_set.find(moduleOrReq) or require(str(moduleOrReq))[0]
  File "C:\Users\jaraco\.local\pipx\shared\Lib\site-packages\pkg_resources\__init__.py", line 884, in require
    needed = self.resolve(parse_requirements(requirements))
  File "C:\Users\jaraco\.local\pipx\shared\Lib\site-packages\pkg_resources\__init__.py", line 770, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'jaraco-clipboard' distribution was not found and is required by the application
pipx > (_parsed_package_to_package_or_url:127): cleaned package spec: jaraco-clipboard
Unable to install jaraco-clipboard.
Check the name or spec for errors, and verify that it can be installed with pip.
←[?25h
```

**Expected behavior**
The package should have installed correctly.

It looks like the problem is the ~~mangling~~normalization of the name. The package name is `jaraco.clipboard` not `jaraco-clipboard`:

```
PS C:\WINDOWS\system32> pip-run jaraco-clipboard
Collecting jaraco-clipboard
  Using cached jaraco.clipboard-2.0.1-py2.py3-none-any.whl (4.6 kB)
Collecting jaraco.windows>=3.4; sys_platform == "win32"
  Using cached jaraco.windows-5.0.0-py3-none-any.whl (54 kB)
Collecting jaraco.text
  Using cached jaraco.text-3.2.0-py2.py3-none-any.whl (8.1 kB)
Collecting more-itertools
  Using cached more_itertools-8.5.0-py3-none-any.whl (44 kB)
Collecting jaraco.collections
  Using cached jaraco.collections-3.0.0-py2.py3-none-any.whl (9.4 kB)
Collecting jaraco.ui
  Using cached jaraco.ui-2.0.1-py2.py3-none-any.whl (7.5 kB)
Collecting six
  Using cached six-1.15.0-py2.py3-none-any.whl (10 kB)
Collecting path.py>=6.2
  Using cached path.py-12.5.0-py3-none-any.whl (2.3 kB)
Collecting jaraco.structures>=1.1.1
  Using cached jaraco.structures-2.0-py2.py3-none-any.whl (4.3 kB)
Collecting jaraco.functools
  Using cached jaraco.functools-3.0.1-py3-none-any.whl (6.7 kB)
Collecting jaraco.classes
  Using cached jaraco.classes-3.1.0-py2.py3-none-any.whl (5.7 kB)
Collecting path
  Using cached path-15.0.0-py3-none-any.whl (21 kB)
Installing collected packages: six, more-itertools, jaraco.functools, jaraco.text, jaraco.classes, jaraco.collections, jaraco.ui, path, path.py, jaraco.structures, jaraco.windows, jaraco-clipboard
Successfully installed jaraco-clipboard jaraco.classes-3.1.0 jaraco.collections-3.0.0 jaraco.functools-3.0.1 jaraco.structures-2.0 jaraco.text-3.2.0 jaraco.ui-2.0.1 jaraco.windows-5.0.0 more-itertools-8.5.0 path-15.0.0 path.py-12.5.0 six-1.15.0
WARNING: You are using pip version 20.2.3; however, version 20.2.4 is available.
You should consider upgrading via the 'c:\program files\python39\python.exe -m pip install --upgrade pip' command.
Python 3.9.0 (tags/v3.9.0:9cf6752, Oct  5 2020, 15:34:40) [MSC v.1927 64 bit (AMD64)] on win32
>>> import pkg_resources
>>> pkg_resources.get_distribution('jaraco-clipboard')
  File "c:\program files\python39\lib\site-packages\pkg_resources\__init__.py", line 480, in get_distribution
    dist = get_provider(dist)
  File "c:\program files\python39\lib\site-packages\pkg_resources\__init__.py", line 356, in get_provider
    return working_set.find(moduleOrReq) or require(str(moduleOrReq))[0]
  File "c:\program files\python39\lib\site-packages\pkg_resources\__init__.py", line 899, in require
    needed = self.resolve(parse_requirements(requirements))
  File "c:\program files\python39\lib\site-packages\pkg_resources\__init__.py", line 785, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'jaraco-clipboard' distribution was not found and is required by the application
>>> pkg_resources.get_distribution('jaraco.clipboard')
jaraco.clipboard 2.0.1 (c:\users\jaraco\appdata\local\temp\pip-run-ir9ktr_4)
>>>
```

### LINKED ISSUE #534: pipx does not find dependencies from extras
**Describe the bug**

pipx accurately installs from a package spec with extras, but it does not keep track of the dependencies from extras in its metadata.

Consequently, using the argument `--include-deps` with `install` will not install the apps of the dependencies from extras.

The problem is that we only supply the name of the main package to `venv_metadata_inspector.py` and not its extras.

Is there a way to determine the extras used to install a package from the venv?  I was unable to find any.

**How to reproduce**

```
> pipx install 'nox[tox_to_nox]' --include-deps
⚠️  Note: register-python-argcomplete was already on your PATH at /usr/local/bin/register-python-argcomplete
  installed package nox 2020.8.22, Python 3.9.0
    - nox
    - python-argcomplete-tcsh
    - register-python-argcomplete
    - tox-to-nox
    - virtualenv
done! ✨ 🌟 ✨
```

**Expected behavior**

Since tox is installed to the venv with extra `tox_to_nox`, the app `tox` should have been exposed.

Also, viewing `pipx_metadata.json` in the nox venv shows no mention of the extra packages in `app_paths_of_dependencies` or `apps_of_dependencies`.

### LINKED ISSUE #587: Packages with . in their name are no longer install-able - Internal Error
This worked a few months ago, but now it fails. For example:

```
pipx install kreagroup.jsi18n                                                                                                                                                                    Internal error with venv metadata inspection.

Unable to install kreagroup-jsi18n.
Check the name or spec for errors, and verify that it can be installed with pip.
```

@itsayellow any ideas about why this could be?

itsayellow:

## Summary of changes
Closes #587.  Closes #534.  Closes #528 .

This revamps `venv_metadata_inspector.py` to use modern python libraries like `packaging` and `importlib.metadata` / `importlib_metadata` instead of `pkg_resources` for venv metadata inspection.

This fixes some things we didn't know we had problems with (see Issues above).  One thing I noticed about the old `venv_metadata_inspector.py` that isn't captured in any Issue, is that currently it completely ignores any environment markers, and just assumes every entry in an `install_requires` is equally valid.  The new code evaluates both markers and the presence or absence of package extras (which may be present in dependencies) correctly.

I started with #344, and made changes.

The biggest changes are to `venv_metadata_inspector.py` and `shared_libs.py`

I have two questions:
* Do we need to list `importlib_metadata` in pipx's `install_requires` if pipx doesn't actually use it in pipx code, but rather installs it in the shared venv? **RESOLVED--Removed it from `install_requires`**
* The list of libraries installed in the shared venv is:`pip`, `wheel`, `packaging`, `importlib-metadata`, and `setuptools`.  Are all of these necessary? **RESOLVED--Keeping all of them.**

After this PR merges I hope to release pipx 0.16.0.0!

## Test plan

The code was extensively tested with our new "slow" tests and 62 packages:
https://github.com/itsayellow/pipx/actions/runs/429239351

There were no errors about pipx's venv metadata inspection missing apps or missing apps of dependencies (This is specifically checked in the slow tests).  The only problems were packages that would not install at all on the CI systems due to missing wheels and/or missing header files on the system.


setup.cfg
```diff
@@ -34,6 +34,7 @@ install_requires =
     userpath>=1.4.1
     argcomplete>=1.9.4, <2.0
     packaging>=20.0
+    importlib_metadata; python_version < '3.8'
```
gaborbernat, line 37: No lower bound limit? Let's pick the latest major as min?

src/pipx/shared_libs.py
```diff
@@ -30,6 +31,15 @@ def __init__(self) -> None:
         # i.e. python_path is ~/.local/pipx/shared/python
         self._site_packages: Optional[Path] = None
         self.has_been_updated_this_run = False
+        # TODO: remove setuptools (wheel?)
+        self.required_packages = [
+            "pip",
+            "wheel",
+            "packaging",
+            "importlib-metadata",
+            "setuptools",
```
gaborbernat, line 40: Why do we need packaging and importlib metadata as shared lib?

setup.cfg
itsayellow: That's a good idea, but there is a bigger question:  Do we need importlib_metadata in setup.cfg at all?  

Technically pipx does not "need importlib_metadata installed to minimally run".  I included it because it is, in a way, a "dependency", but a strange one because pipx installs it itself only in the shared libs.  It is not used in pipx code other than `venv_metadata_inspector.py`, which doesn't run in pipx's environment but in the venv + shared_lib environments.

src/pipx/shared_libs.py
itsayellow: This is the major change to `venv_metadata_inspector.py`.  Instead of it using setuptools' `pkg_resources` to inspect each venv, it now uses `packaging` and `importlib-metadata`.

From `packaging.requirements` it uses `Requirement`, and from `packaging.utils` it uses `canonicalize_name()`.
From `importlib[.-]metadata` it uses `distribution()`, `distributions()`, and type `Distribution`. 

I couldn't figure out a way to get along without both of those dependencies if we transition away from `pkg_resources`.  If anyone can think of a way I'd be happy to pursue it!  Adding more libraries to the shared libs does increase the chance that we will conflict with requirements in the venvs.

uranusjr: I think we can avoid `importlib-metadata` in 3.8+ environments by adding a marker here (plus try-except imports in `venv_metadata_inspector`)?

setup.cfg
uranusjr: I would say it’s not needed here, for the exact reason you said. pipx’s runtime dependencies are entirely separate from its share libs, and having importlib-metadata in one place but not the other does not sound wrong to me.

src/pipx/shared_libs.py
itsayellow: I was guarding against the user possibly having a mix of 3.8+ and <3.8 venvs, in which case the shared libs would need `importlib-metadata` even if pipx and possibly other venvs were running in 3.8+.

itsayellow: Although now that I think about it, if the shared libs are installed 3.8+ but used with a venv that is <3.8, does that imply that the <3.8 venvs get to use `importlib.metadata`?

setup.cfg
itsayellow: Removed it.

src/pipx/shared_libs.py
itsayellow: And there is already a try-except sequence in `venv_metadata_inspector.py` to first attempt to use `importlib.metadata` and then to fall back on using `importlib-metadata`

gaborbernat (approved):

itsayellow:
@cs01, any thoughts?  I realize it's around Christmas and you probably have better things occupying yourself.  Just thought I'd try to at least give you an opportunity to chime in before I start releasing things. 🙂

Happy Holidays everyone, btw 🎄

src/pipx/venv_metadata_inspector.py
```diff
@@ -12,88 +40,114 @@
...
+    # "scripts" entry in setup.py is found here (test w/ awscli)
+    for path in dist.files or []:
+        dist_file_path = Path(dist.locate_file(path)).resolve()
+        try:
+            if dist_file_path.parent.samefile(bin_path):
+                apps.add(path.name)
+        except FileNotFoundError:
+            pass
+
+    # not sure what is found here
```
cs01, line 110: Is this still the case? I wonder if we can remove this section.

cs01:
Wow this is great, an early Christmas present :). This and will handle all the annoying corner cases, and be a nice foundation for the future of pipx.

> The list of libraries installed in the shared venv is: pip, wheel, packaging, importlib-metadata, and setuptools. Are all of these necessary?

The packages installed into the shared libs were originally introduced in https://github.com/pipxproject/pipx/pull/168/files.

As you know, the shared libs are essentially "merged" at runtime with pipx's package-specific venvs via the pth file. This is to save time by avoiding installing a bunch of the same things over and over for each venv. It forms a common environment for the app's venvs to be able to run pip and the metadata inspector.

pip is definitely required to install packages. pip often needs to build things from source. The source build system is often setuptools, so that is required in case the package is an sdist. To build into a wheel, it needs the `wheel` package. `packaging`, `importlib_metadata` are used directly by `venv_metadata_inspector.py` so they are required too. 

This LGTM!

src/pipx/shared_libs.py
cs01, line 40: I just realized there are no version constraints on these package specifiers. I have been getting bit on some semver changes on my other projects so it has been on my mind. However, thinking about these a little, it's likely we always want to be on the latest version of all of them since they all work in tandem (i.e. an update to the wheel format will be reflected across.

> Adding more libraries to the shared libs does increase the chance that we will conflict with requirements in the venvs.

I wonder what would happen here. The shared lib packages are added via the pth file, and aren't part of the requirements of the package being installed. I assume pip (and its new dependency resolver) only applies to the requirements/package being installed (and its dependencies). The shared lib pip invocation is separate from app package installation and manipulation, so maybe there won't be any issues here?

src/pipx/venv_metadata_inspector.py
itsayellow: I wondered too...but I was also really worried about corner cases because the universe of strange python packages is vast.

I wonder if this duplicates the above for loop `for path in dist.files`.  Maybe one of the other python packaging experts can give some guidance?

itsayellow: @uranusjr , @gaborbernat , do either of you know if `installed-files.txt` is only used for Python 2.*?  It's hard to find documentation on it, but what I find all seems to refer to Python 2.*.  If `installed-files.txt` is only used for Python 2.* then we can definitely remove it.

uranusjr: I don’t think it has anything to do with Python 2/3. AFAIK `installed-files.txt` is a setuptools/distutils artifact from the old days before `.dist-info` was specified. Newer setuptools versions should never write this.

itsayellow: I find references to ancient pip generating this file (there is a mention in changelog for pip 0.3 https://pip.pypa.io/en/stable/news/#id556)

itsayellow: Regardless of the library of origin, it seems very old.  If we are updating our shared libraries this code that views `installed-files.txt` seems like it can be removed.

Let me know if there are any objections, otherwise I will remove it.

uranusjr: I didn't read the code closely; can someone explain what it actually does? While new(-ish) packaging tools would never generate the file, it may still be present in systems based on Python packaging but never bothered to update.

itsayellow: It just goes through every line in file `installed-files.txt` and sees if any of them is in the venv `bin/` subdirectory.  Basically exactly the same thing as the for loop above it does by searching the iterable `Distribution.files`.

If this is part of ancient `pip`, I'm wondering if it was created each time on install on the client machine, and not made as part of the package distribution?

itsayellow: One of the few mentions I can find on the internet, from a setuptools Issue discussing a possible solution (June 2015):
https://github.com/pypa/setuptools/issues/371#issuecomment-202918638
> Just have setuptools add installed-files.txt to the egg-info directory. pip has been doing this for a long time before .dist-info was on the radar.

src/pipx/shared_libs.py
uranusjr: > I wonder what would happen here. The shared lib packages are added via the pth file, and aren't part of the requirements of the package being installed. I assume pip (and its new dependency resolver) only applies to the requirements/package being installed (and its dependencies). The shared lib pip invocation is separate from app package installation and manipulation, so maybe there won't be any issues here?

It depends on how and when the shared libs are injected. pip determines whether a package is installed by inspecting directories on `sys.path`, so if the shared libs are injected when `pip install` is run, pip will consider those packages as “installed”. If you’re going to add version constraints to those libraries, they could end up conflicting with applications users want to install.

I think maybe the best way to implement `venv_metadata_inspecor.py` is to run it as a [zipapp](https://docs.python.org/3/library/zipapp.html) instead. pipx can create a zipapp `venv_metadata_inspecor.pyz` that contains the script and all its dependencies in a cache directory, and run `python venv_metadata_inspecor.pyz` instead. The same technique can actually be used for `pip install` to avoid potential edge cases where an application depends on a certain pip and setuptools version not compatible with pipx (not likely to happen though).

itsayellow: The zipapp idea is very interesting.  Wouldn't pipx still need to be running inside of the target venv anyway for `importlib[.-]metadata.Distribution` to get information?  And wouldn't that end up with the same version conflicts?

I don't see any way to point `importlib[.-]metadata` to examine another venv than the current one, in which case I'm worried even a zipapp would encounter the same version conflicts.

Is there a way around these problems?

itsayellow: Oooh once I actually wrote out "any way to point importlib[.-]metadata to examine another venv than the current one" I realized that is the key, and I found this pointer, which may show the way to do that:
https://docs.python.org/3/library/importlib.metadata.html#extending-the-search-algorithm

I'm wondering if we could hook into that to facilitate running the code in one venv and inspecting another?

itsayellow: If we could do THAT, then we could make venv inspection just another part of pipx and not a separate runnable file.  That would really be cool.

itsayellow: Actually, maybe all we need to do is manipulate `sys.path`.  It might be as simple as that, we just need to put the directories of the venv to be inspected in `sys.path` before asking `importlib.metadata` about distribution information.  I've been experimenting and that seems to work.

itsayellow: Since I'm stuck at home alone on Christmas Eve 😆 , I've been trying the `sys.path` manipulation trick, and it works so well.  I may have a new PR to audition for you all soon.

uranusjr: pipx itself does not need to be run inside the target environment, only the `venv_metadata_inspecor` script.

> Actually, maybe all we need to do is manipulate `sys.path`.

This is precisely what a zipapp does; basically `sys.path.insert(0, path_to_zip)`. But by using a zipapp, you can separate pip + setuptools and other stuff, so those other stuff would not be present in the environment when `pip install` is run, thus not resulting vesion conflicts.

I’m not quite sure how `sys.path` manipulation would work without overhauling how shared libs are managed. Currently all shared libs are installed in one virtual environment, and injecting any one of them means injecting all of them. So when you inject pip to run `pip install`, pip will still see the shared packaging etc. libs and potentially emit version conflicts. To prevent that from happening, you’ll need to install packaging etc. into a separate environment, which is basically manually re-implement zipapp.

itsayellow: What I've been doing in another branch, is to have the shared libs be only what we currently have: `pip`, `setuptools`, and `wheel`.  Just to install new packages, not to inspect venvs.  Then I've made the equivalent of `venv_metadata_inspector.py` just another submodule of pipx.  The only thing I need to do that's "special" is: before cataloging `Distributions`, I swap out pipx's `sys.path` list of directories with the `sys.path` for the venv.  Then once I've cataloged all the `Distributions`, I swap pipx's `sys.path` back.

It seems to work great (I'm not importing anything during the temporary change in `sys.path`) and as a bonus, the venv inspection code is a first-class citizen of pipx, and we don't have to use subprocesses, pass around json, or use python 3.5 syntax for type-checking.

itsayellow: Basically I'm only changing `sys.path` in order to change what `importlib[.-]metadata` is looking at when we get info about the distributions.  Otherwise, the code is still part of the pipx package and not in a separate venv.

uranusjr: Ah OK, so you do all the inspection outside of the target virtual environment? That should work.

I believe you can also avoid `sys.path` manipulation by passing a path context, e.g. `importlib_metadata.distributions(path=path_to_venv_site_packages)`.

itsayellow: > I believe you can even avoid `sys.path` manipulation altogether by passing a path context, e.g. `importlib_metadata.distributions(path=path_to_venv_site_packages)` (not sure if this would work, @jaraco is the expert on this).

That would be even better.  Let me see what I can do.

jaraco: The `path` parameter is meant to be a sequence like `sys.path` so something like `importlib.metadata.distributions(path=[path_to_venv_site_packages])` may be what you're after.

itsayellow: Thanks, @jaraco.  I think I've got that working in another branch.

itsayellow:
Closing in favor of instead using #593

PR #593, opened 2020-12-27
Improved venv inspection

### LINKED ISSUE #528: [Windows] Unable to inject jaraco.clipboard

### LINKED ISSUE #534: pipx does not find dependencies from extras

### LINKED ISSUE #587: Packages with . in their name are no longer install-able - Internal Error

itsayellow:

## Summary of changes
This is the successor to (and based on) #588 .
Closes #587.  Closes #534.  Closes #528 .

This integrates the purpose of `venv_metadata_inspector.py` into code that is part of the pipx package and is no longer stand-alone.  The breakthrough is that `importlib.metadata` allows us to specify the path for it to search for package metadata.  This means that the venv inspection code no longer needs to run in the venv environment itself, but simply specify what `sys.path` would be if the venv environment were active.  

**EDIT**: We can also fetch the venv `environment` of environment markers and apply it when evaluating package requirements to determine if a dependency is a match for the environment (see `packaging.markers.Marker.evaluate`).

The new code does not need to be run in a subprocess nor exchange data via json, and benefits from logging and python3.6+ of being a native part of pipx.

There is only one tiny subprocess that runs inside the venv to be examined.  It gets the venv python version and what `sys.path` is when the venv is operational.  I used the entire `sys.path` of the activated venv because that is the existing behavior, and also in case the user includes options to use system libraries to fulfill dependencies.  **EDIT**: the subprocess now also fetches a dict specifying the venv environment for purposes of matching environment markers.

Additionally a huge speedup of the inspection for certain packages has been accomplished.  These two lines are responsible for up to a 2 second speedup in certain cases!
https://github.com/pipxproject/pipx/blob/48f6fe60169ef65120dc374152ed5d40a3a36bf2/src/pipx/venv_inspect.py#L92-L93
`get_apps()` was taking a huge amount of time, and the loop that checks each file was taking all the time.  This if-statement was a much faster way of knowing that the file is not an app than the `pathlib` checks in the loop.
Some selected packages and venv inspection time (start to finish) are shown below.  "Optimization" refers to the two lines in the code linked to above.

**venv inspection times (less is better)**
| package         |  this PR  |  pipx v0.15.6.0 | this PR no optimization  |
| ----------------|-----------:|------------:|------------:|
| ansible==2.9.13 |  259ms    |  1738ms    | 1689ms     |
| awscli          |  176ms    |  1140ms    |  998ms     |
| beancount       |  484ms    |  1110ms    |  820ms     |
| cloudtoken      |  239ms    |   697ms    |  462ms     |
| kibitzr         |  380ms    |   838ms    |  779ms     |
| kolibri         |  270ms    |  2456ms    | 2455ms     |
| magic-wormhole  |  536ms    |  1410ms    | 1028ms     |
| mkdocs          |  153ms    |   324ms    |  413ms     |
| sphinx          |  274ms    |   795ms    |  394ms     |

After this PR merges I hope to release pipx 0.16.0.0!

60+ packages (pipx slow tests) pass fine
https://github.com/itsayellow/pipx/actions/runs/489220534

## Test plan

Tested by running
```
```


src/pipx/package_specifier.py
```diff
@@ -188,6 +188,23 @@ def parse_specifier_for_upgrade(package_spec: str) -> str:
     return package_or_url
 
 
+def get_extras(package_spec: str) -> Set[Any]:
```
uranusjr, line 191: Aren’t extras supposed to only be strings? `packaging` declares `Requirement.extras` as `Set[str]`.

src/pipx/package_specifier.py
```diff
@@ -188,6 +188,23 @@ def parse_specifier_for_upgrade(package_spec: str) -> str:
...
+        (_, package_extras_str) = _split_path_extras(parsed_package.valid_local_path)
+        return Requirement("notapackage" + package_extras_str).extras
+
+    return set()
+
+
+def append_extras(package: str, package_or_url: str) -> str:
+    package_spec = Requirement(package)
+    package_spec.extras = get_extras(package_or_url)
+    return str(package_spec)
```
uranusjr, line 205: The interface around here is pretty weird. What are `package` and `package_or_url` supposed to contain? 

If `package` is only expected to contain the package name, there are better ways to append extras. If it can contain any PEP 508 requirements, this “append” operation does not look right.

src/pipx/venv.py
```diff
@@ -334,7 +290,9 @@ def _update_package_metadata(
         is_main_package: bool,
         suffix: str = "",
     ) -> None:
-        venv_package_metadata = self.get_venv_metadata_for_package(package)
+        venv_package_metadata = self.get_venv_metadata_for_package(
+            append_extras(package, package_or_url)
+        )
```
uranusjr, line 295: Again, what is `get_venv_metadata_for_package` supposed to accept? Why do we need to append extras to get venv metadata (but we didn’t previously)?

itsayellow: For example, if the user specified a specifier `mypackage[extra1]==2.0.1`, we used to only supply `mypackage` to `get_venv_metadata_for_package`.  Now we would supply `mypackage[extra1]`.  The needed information is just the package name and the extras, we don't need the rest of the specifier.

We didn't previously because our venv inspection was basically not accurate.

Previously, `venv_metadata_inspector.py` would return every single possible dependency of a package, even those that were not valid for the current install.  Examples:
1. It would match dependencies for extras that the installed package was not actually using
2. It would match dependencies for environment markers that were not valid for the current enviornment

src/pipx/package_specifier.py
itsayellow: I'm fine with changing the interface, but basically to do a proper venv inspection for this package we need both the package name, and the extras used to install that package.  We don't need other parts of the package specifier.

itsayellow: `package` is just the package name, and `package_or_url` is the full specifier.  This is naming scheme is somewhat legacy for the pipx codebase.

src/pipx/package_specifier.py
itsayellow: Yes good catch.  Changed.

src/pipx/venv_inspect.py
```diff
@@ -0,0 +1,227 @@
...
+    # Add an empty extra to enable evaluation of non-extra markers
+    if not extras:
+        extras.add("")
+    dependencies = []
+    for req in map(Requirement, dist.requires or []):
+        if not req.marker:
+            dependencies.append(req)
+        else:
+            for extra in extras:
+                if req.marker.evaluate({"extra": extra}):
```
itsayellow, line 62: Darn, @uranusjr's comments in another section just reminded me of this statement.  Unfortunately `req.marker.evaluate` is using the current environment.  When this code was running in the venv it was accurately evaluating environment markers.  But running in pipx it may not be (for example the python version may be wrong.)

itsayellow: Ah!  But `Marker.evaluate()` accepts an `environment` dict to supply an override-environment.  I need to implement that.

src/pipx/venv.py
itsayellow: Actually it is now very easy to change `get_venv_metadata_for_package` to take two arguments, package name and extras.  I had concatenated them together in one string before because `venv_metadata_inspector.py` was run like an executable, but now that `inspect_venv()` is part of pipx it is easy to supply real arguments.

If the package name and extras were supplied as separate arguments, would that work for a good interface?

src/pipx/package_specifier.py
itsayellow: > If `package` is only expected to contain the package name, there are better ways to append extras.

Just out of curiosity, what are the better ways you're thinking of?

src/pipx/venv_inspect.py
itsayellow: OK, I added implementation of code to get the environment inside the venv to apply when executing `Marker.evaluate()` to determine if a dependency is used.

src/pipx/venv_inspect.py
```diff
@@ -0,0 +1,268 @@
...
+    )
+    venv_info = json.loads(
+        run_subprocess(
+            [venv_python_path, "-c", command_str],
+            capture_stderr=False,
+            log_cmd_str="<inline fetch_info_in_venv commands>",
+        ).stdout
+    )
+    return (
+        venv_info["sys_path"],
```
itsayellow, line 213: Just realized that `sys.path` is returning the first entry as an empty string, which should probably be removed, since when we are fetching distributions we don't want to include the "current directory".

itsayellow:
The "small subprocess code" has grown to be a bit larger than it started.  I'm happy to break it out into a separate file if people think that's a good idea.  Let me know if people are interested in that.

src/pipx/package_specifier.py
uranusjr: Hmm. I haven’t thought it all through, but since this joined the package name and extras, I would probably make `get_venv_metadata_for_package` and `inspect_venv` take `package: str, extras: Set[str]` instead.

Another ways to do this is to perform marker evaluation. Instead of discarding dependency information with non-matching markers, all dependencies can be stored lossless PEP 508 strings in the metadata storage, and only parsed and filtered when they are actually used. This is cleaner IMO, but may require refactoring too significant to be done with the importlib-metadata migration.

itsayellow: I don't totally understand how this would work.  Part of it is that the recursive search includes dependencies of dependencies of ...  And if you include branches of that tree that don't really exist (because they are not needed for you OS, extras, etc.) you are making a lot of work for nothing.

itsayellow: As for the interface, I realized after looking again at the code that part of what makes it convenient to have an argument in the format of a requirement, is that that's what the recursive function is expecting.  So I'd probably be combining the name and the extras anyway to supply that function.

itsayellow: I changed `get_venv_metadata_for_package` and `inspect_venv take` separate package and extras arguments.

uranusjr:
Maybe call the script `venv_metadata_inspector.py` and restore the mechanism used to call it? :p

itsayellow:
> Maybe name the script `venv_metadata_inspector.py` and restore the mechanism used to call it? :p

Yes I guess so.  :)  That whole mechanism seemed to so tortured to me, but I can't argue with success.

itsayellow:
@uranusjr I can't tell how serious your request is with the :p emoji.  🙂 

Do you think it's ok to keep the command-string inline?  I kind of like it as it is because it's so simple and obvious.  And I remember having issues finding the old `venv_metadata_inspector.py` in the code when I was doing experiments freezing the code with PyInstaller.

Plus right now it's not all that complicated code.

uranusjr:
I think at its current length it's okay to do this either way. If you decide to make this a standalone file, all previous mechanisms should return (of course), but I'd want the file to be named something different; `venv_metadata_inspect.py` is pretty vague and arguably does not reflect what your current script actually does.

itsayellow:
Ah ok, then I'm totally in agreement.  I think since the code right now is about one function in length, I like leaving it where it is.

gaborbernat (approved):

PR #602, opened 2021-01-20
Run a command from a [pipx.run] entry point

uranusjr:

## Test plan

Tested by running
```
pipx run --spec="build @ https://github.com/uranusjr/build/archive/pipx-run.zip" build --version
```

### DIFF
--- a/src/pipx/commands/run.py
+++ b/src/pipx/commands/run.py
@@ -6,7 +6,12 @@ import urllib.parse
 import urllib.request
 from pathlib import Path
 from shutil import which
-from typing import List
+from typing import List, Optional, Union
+
+try:
+    from importlib.metadata import Distribution, EntryPoint
+except ImportError:
+    from importlib_metadata import Distribution, EntryPoint  # type: ignore
 
 from pipx import constants
 from pipx.commands.common import package_name_from_spec
@@ -17,6 +22,7 @@ from pipx.util import (
     PipxError,
     exec_app,
     get_pypackage_bin_path,
+    get_site_packages,
     pipx_wrap,
     rmdir,
     run_pypackage_bin,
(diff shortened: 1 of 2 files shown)


src/pipx/commands/run.py
```diff
@@ -145,21 +189,17 @@ def _download_and_run(
...
         # Let future _remove_all_expired_venvs know to remove this
         (venv_dir / VENV_EXPIRED_FILENAME).touch()
 
-    # This never returns
-    venv.run_app(app, app_args)
+    # These never return
+    if isinstance(command, EntryPoint):
+        venv.run_entry_point(command, app_args)
+    else:
+        venv.run_app(app, app_args)
```
henryiii, line 202:
```suggestion
        venv.run_app(command, app_args)
```

gets changed on windows. Wish single dispatch method was available pre 3.8.

uranusjr: It just occurs to me `f"{app}.exe"` is not actually needed. `subprocess` on Windows automatically resolves the extension (sine it is implemented with the `CreateProcess` Windows API).

henryiii: > This never returns

Why are these not annotated NoReturn instead of None, by the way? Would need fewer comments and the type checker would check the validity of the no return status. Unrelated, just thought I'd ask here.

henryiii: Isn't the windows test failing because of this? Or is it just tripping a unit test?

uranusjr:
The test fails because it cannot find a line in the logs; cowsay itself seems to run fine.

src/pipx/commands/run.py
uranusjr: > Why are these not annotated NoReturn instead of None, by the way?

Probably some legacy reasons, don’t know.

> Isn't the windows test failing because of this? Or is it just tripping a unit test?

The test fails because it cannot find a line in the logs; cowsay itself seems to run fine.

itsayellow: Pretty sure I wrote the `None` return values because I was not aware there was a `NoReturn`.  Feel free to fix that.

henryiii:
What's the status of this? :)

uranusjr:
Continuing in #615.

PR #606, opened 2021-01-22
Drop .exe appending on Windows

uranusjr:
This makes `Venv.run_app()` and code path along it accept an additional argument, so we can distinguish between the "real" name (filename) of an app, and the name supplied by the user.
    
This makes the code simpler, espacially when dealing with the proposed `[pipx.run]` entry points.

### DIFF
--- a/src/pipx/commands/run.py
+++ b/src/pipx/commands/run.py
@@ -67,9 +67,11 @@ def run(
             )
         )
 
-    if WINDOWS and not app.endswith(".exe"):
-        app = f"{app}.exe"
-        logger.info(f"Assuming app is {app!r} (Windows only)")
+    if WINDOWS:
+        app_filename = f"{app}.exe"
+        logger.info(f"Assuming app is {app_filename!r} (Windows only)")
+    else:
+        app_filename = app
 
     pypackage_bin_path = get_pypackage_bin_path(app)
     if pypackage_bin_path.exists():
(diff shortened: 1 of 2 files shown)


src/pipx/venv.py
```diff
@@ -327,8 +327,8 @@ def list_installed_packages(self) -> Set[str]:
         pip_list = json.loads(cmd_run.stdout.strip())
         return set([x["name"] for x in pip_list])
 
-    def run_app(self, app: str, app_args: List[str]) -> NoReturn:
-        exec_app([str(self.bin_path / app)] + app_args)
+    def run_app(self, app: str, filename: str, app_args: List[str]) -> NoReturn:
+        exec_app([str(self.bin_path / filename)] + app_args)
```
itsayellow, line 331: `app` doesn't seem to be used in this function at all anymore.

src/pipx/commands/run.py
```diff
@@ -140,7 +144,7 @@ def _download_and_run(
         is_main_package=True,
     )
 
-    if not (venv.bin_path / app).exists():
+    if not (venv.bin_path / app_filename).exists():
```
itsayellow, line 147: The following `PipxError` seems to be the only remaining necessary use of `app`.  Is it useful to have both `app` and `app_filename` just so that the error message doesn't suffix the app with ".exe"?  Or am I missing another use for the two separate variables?

src/pipx/venv.py
uranusjr: This argument will be needed to query the `[pipx.run]` hook so I leave it in for now.

src/pipx/commands/run.py
itsayellow: OK, I see your other comment.  So it seems this is mainly a setup for your other PR?

uranusjr: I’m also thinking maybe eventually the `.exe` existence checks should be replaced by `shutil.which()` (and `app_filename` can be dropped entirely after that), but that needs a wider discussion and I’m probably not going to spend much energy on it unless someone complains about pipx assuming `.exe`.

itsayellow: OK.  This change seems innocuous to me, so if it helps you to merge this it's fine with me.

itsayellow (approved):

PR #607, opened 2021-01-22
One windows check

itsayellow:

## Summary of changes
This just centralizes the code to check if we are running on Windows to `constants.py`.

Previously, there were a couple of different places where the code checked if we were on Windows.  A bunch of modules were even importing from `utils.py`, which in turn was importing from `constants.py`!!

## Test plan

Tested by running
```
```


itsayellow:
@uranusjr looks like you had the same idea in #606 😄

gaborbernat (approved):

PR #610, opened 2021-01-23
Fix show/hide cursor for older Windows

### LINKED ISSUE #603: Windows CMD - colors and Symbols are not displayed correctly
**Describe the bug**
Using Windows 10 colors and symbols are not displayed.

**How to reproduce**
Install pipx with Python 3.8 on Windows 10 like described in the documentation.
Use any pipx command with cmd.

**Expected behavior**
Symbol and Colors are displayed like in a bash shell.

What I currently get is:

2021-01-20_14h04_14

itsayellow:

## Summary of changes
Closes #603 .

On Windows, this PR adds and uses the method of showing/hiding the cursor outlined in https://stackoverflow.com/a/10455937

## Test plan

I don't have an older copy of Windows, so if somebody could test this with an old Windows (before Windows 10 TH2 (v1511)) I would be much appreciative.

From #603, all one would need to do would be to see if there are visible ANSI escape codes printed (e.g. `[?25l`,` [?25h`) when executing pipx commands like:

```
pipx --version
```

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,6 +1,6 @@
 dev
 
-- Fix cursor show/hide to work with older versions of Windows.
+- Fix cursor show/hide to work with older versions of Windows. (#610)
 
 0.16.0.0
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,6 @@
 dev
 
+- Fix cursor show/hide to work with older versions of Windows.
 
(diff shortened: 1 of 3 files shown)


itsayellow:
Testing by @carlijoy in #603 indicates that this fixes the problems due to `show_cursor` and `hide_cursor`.

uranusjr (approved):

PR #612, opened 2021-01-24
Add text color capability for Windows

itsayellow:

## Summary of changes
With colorama it is ridiculously easy to support text color output on Windows, so I added it with this PR.  It adds polish and brings the Windows experience in line with other platforms.

## Test plan

Tested by running
```
pipx list --verbose
```
On Windows both the bold list formatting and the green of the verbose output show up.  (They don't with existing pipx.)

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,6 @@
 dev
 
+- Support text colors on Windows.
 
 0.16.0.0
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,6 +1,6 @@
 dev
 
-- Support text colors on Windows.
+- Support text colors on Windows. (#612)
 
 0.16.0.0
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,6 +31,7 @@ include_package_data = true
 zip_safe = true
 python_requires = >=3.6
 install_requires =
+    colorama>=0.4.4
     userpath>=1.4.1
     argcomplete>=1.9.4, <2.0
     packaging>=20.0
--- a/src/pipx/colors.py
+++ b/src/pipx/colors.py
@@ -1,7 +1,7 @@
 import sys
 from typing import Callable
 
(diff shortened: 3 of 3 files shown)


uranusjr:
colorama also handles tty detection so `init()` can be called unconditionally. But I guess there's no harm to check either.

itsayellow:
There's also this: https://stackoverflow.com/questions/12492810/python-how-can-i-make-the-ansi-escape-codes-to-work-also-in-windows/60194390#60194390
Which seems to indicate with a few lines of code using `ctypes` we don't need colorama.  It seems great but I wonder if colorama is providing us with something else valuable?

itsayellow:
To follow up on my last comment, this https://stackoverflow.com/a/36760881 gives some more information.  It seems that it's only turning on Windows' native ANSI-interpreter.

So the simple non-colorama code might not help people whose Windows doesn't have ANSI capabilities.

itsayellow:
> colorama also handles tty detection so `init()` can be called unconditionally. But I guess there's no harm to check either.

That's a good point.  I think I will leave it in.  If I read colorama's docs right, it's still trying to filter out all the color codes in a non-tty environment.  So since we don't add the codes in the first place in the that case, I like not having to filter `sys.stdout` and `sys.stderr` at all.

CarliJoy:
Current Result on Windows:

2021-01-27_14h51_03

itsayellow:
Thanks for the independent test, @CarliJoy , that seems to work as expected, the codes still visible are fixed in the other PR.

itsayellow:
> colorama also handles tty detection so `init()` can be called unconditionally. But I guess there's no harm to check either.

Another possible option would just be to use colorama for everything--so the "color functions" would always add the color escape codes, and we rely on colorama to remove them if we're not in a tty.

I didn't go this route just out of conservatism (original code works) and because I have no idea how much overhead this would add.

PR #614, opened 2021-01-26
Better Unicode / emoji capability detection

### LINKED ISSUE #73: Traceback when locale not set to UTF-8
On a machine where the locale has not been set to UTF-8 you get tracebacks when installing/running pipx. E.g. when installing:
```
  File "get-pipx.py.1", line 287, in <module>
    main()
  File "get-pipx.py.1", line 283, in main
    print(f"Enjoy! {'\u2728 \U0001f31f \u2728' if not WINDOWS else ''}")
UnicodeEncodeError: 'latin-1' codec can't encode character '\u2728' in position 7: ordinal not in range(256)
```
Or when installing a package:
```
ipx install poetry
zsh: correct 'pipx' to 'popd' [nyae]? n
2nstalling package 'poetry'       
  installed package poetry 0.12.11, Python 3.6.7
  These binaries are now globally available
    - poetry
  File "/home/pmav2/.local/bin/pipx", line 10, in <module>
    sys.exit(cli())
  File "/home/pmav2/.local/pipx/venvs/pipx-app/lib/python3.6/site-packages/pipx/main.py", line 932, in cli
    run_pipx_command(args)
  File "/home/pmav2/.local/pipx/venvs/pipx-app/lib/python3.6/site-packages/pipx/main.py", line 642, in run_pipx_command
    force=args.force,
  File "/home/pmav2/.local/pipx/venvs/pipx-app/lib/python3.6/site-packages/pipx/main.py", line 505, in install
    print(f"done! {stars}")
UnicodeEncodeError: 'latin-1' codec can't encode character '\u2728' in position 6: ordinal not in range(256)
```
Obviously, the problem is the special characters in the print statements. My guess is that you could check `sys.getdefaultencoding()` before you actually try to print them.

If you want to test this, on ubuntu you can set an "ansi" encoding with:
```
sudo update-locale LANG=en_US
```
and restore utf-8 with:
```
sudo update-locale LANG=en_US.UTF-8
```
You probably need to open a new terminal after each change though.

itsayellow:

## Summary of changes
Closes #73.

Previously we only disallowed emojis (Unicode characters) when we detected Windows, and allowed for all other platforms.  This ignored the fact that some shells in Windows are now perfectly happy displaying emojis/Unicode, while some non-Windows platforms' shells are not using a Unicode encoding and actually crash when we try to print emoji (See #73.)

This PR actually directly tests encoding our emoji unicode characters using `sys.stderr.encoding` to determine if it will work on a platform.

## Test plan

Uninstalling a nonexistent package will show the "sleep" emoji.  This should omit the emoji and not yield an error if non-Unicode encoding is used in the shell.  It should show an emoji if the shell has Unicode-capable encoding.

This PR:
```
(venv39) > pipx uninstall doesntexist           
Nothing to uninstall for doesntexist 😴
(venv39) > LANG=en_US pipx uninstall doesntexist
Nothing to uninstall for doesntexist 
```

pipx 0.16.0.0:
```
> pipx uninstall doesntexist
Nothing to uninstall for doesntexist 😴
> LANG=en_US pipx uninstall doesntexist
  File "/home/mclapp/.local/bin/pipx", line 8, in <module>
    sys.exit(cli())
  File "/home/mclapp/.local/lib/python3.9/site-packages/pipx/main.py", line 752, in cli
  File "/home/mclapp/.local/lib/python3.9/site-packages/pipx/main.py", line 234, in run_pipx_command
    return commands.uninstall(venv_dir, constants.LOCAL_BIN_DIR, verbose)
  File "/home/mclapp/.local/lib/python3.9/site-packages/pipx/commands/uninstall.py", line 27, in uninstall
    print(f"Nothing to uninstall for {venv_dir.name} {sleep}")
UnicodeEncodeError: 'latin-1' codec can't encode character '\U0001f634' in position 37: ordinal not in range(256)
```


itsayellow:
I moved the `EMOJI_SUPPORT` constant and detection into `emojies.py` because it seemed appropriate there, but if people like it to remain in `constants.py` I would be amenable to that.

uranusjr:
Is the plural of emoji *emojis* or *emojies*? The spelling is inconsistent between the commit message and the module name.

itsayellow:
> Is the plural of emoji _emojis_ or _emojies_? The spelling is inconsistent between the commit message and the module name.

Yeahhhhhh.  I was trying to ignore that 😛   

https://www.theatlantic.com/technology/archive/2016/01/whats-the-plural-of-emoji-emojis/422763/
> There is a correct answer, but no one has agreed on it yet.

It does seem that the plural is a choice between "emojis" and "emoji".  So I'm going to rename the module "emojis" without the "e", since "emojies" seems unloved by everybody.

PR #615, opened 2021-01-29
Run a command from a [pipx.run] entry point

uranusjr:
Supercedes #602.

---

## Test plan

Tested by running
```
pipx run --spec="build @ https://github.com/uranusjr/build/archive/pipx-run.zip" build --version
```

### DIFF
--- a/src/pipx/venv.py
+++ b/src/pipx/venv.py
@@ -1,9 +1,15 @@
 import json
 import logging
+import re
 import time
 from pathlib import Path
 from subprocess import CompletedProcess
-from typing import Dict, Generator, List, NoReturn, Set
+from typing import Dict, Generator, List, NoReturn, Optional, Set
+
+try:
+    from importlib.metadata import Distribution, EntryPoint
+except ImportError:
+    from importlib_metadata import Distribution, EntryPoint  # type: ignore
 
 from packaging.utils import canonicalize_name
 
@@ -33,6 +39,16 @@ from pipx.util import (
 from pipx.venv_inspect import VenvMetadata, inspect_venv
 
 logger = logging.getLogger(__name__)
+
+# This regex was only introduced into importlib.metadata in Python 3.9.
+_entry_point_value_pattern = re.compile(
+    r"""
+    ^(?P<module>[\w.]+)\s*
+    (:\s*(?P<attr>[\w.]+))?\s*
+    (?P<extras>\[.*\])?\s*$
+    """,
+    re.VERBOSE,
+)
 
 
 class VenvContainer:
@@ -327,8 +343,33 @@ class Venv:
         pip_list = json.loads(cmd_run.stdout.strip())
         return set([x["name"] for x in pip_list])
 
+    def _find_entry_point(self, app: str) -> Optional[EntryPoint]:
+        if not self.python_path.exists():
+            return None
+        site_packages = get_site_packages(self.python_path)
+        for dist in Distribution.discover(path=[str(site_packages)]):
+            for ep in dist.entry_points:
+                if ep.group == "pipx.run" and ep.name == app:
+                    return ep
+        return None
+
     def run_app(self, app: str, filename: str, app_args: List[str]) -> NoReturn:
-        exec_app([str(self.bin_path / filename)] + app_args)
+        entry_point = self._find_entry_point(app)
+
+        # No [pipx.run] entry point; default to run console script.
+        if entry_point is None:
+            exec_app([str(self.bin_path / filename)] + app_args)
+
+        # Evaluate and execute the entry point.
+        match = _entry_point_value_pattern.match(entry_point.value)
+        assert match is not None, "invalid entry point"
+        module, attr = match.group("module", "attr")
+        code = (
+            f"import sys, {module}\n"
+            f"sys.argv[0] = {entry_point.name!r}\n"
+            f"sys.exit({module}.{attr}())\n"
+        )
+        exec_app([str(self.python_path), "-c", code] + app_args)
 
     def has_app(self, app: str, filename: str) -> bool:
         if self._find_entry_point(app) is not None:
--- a/src/pipx/venv.py
+++ b/src/pipx/venv.py
@@ -330,6 +330,11 @@ class Venv:
     def run_app(self, app: str, filename: str, app_args: List[str]) -> NoReturn:
         exec_app([str(self.bin_path / filename)] + app_args)
 
+    def has_app(self, app: str, filename: str) -> bool:
+        if self._find_entry_point(app) is not None:
+            return True
+        return (self.bin_path / filename).is_file()
+
     def _upgrade_package_no_metadata(self, package: str, pip_args: List[str]) -> None:
         with animate(
             f"upgrading {full_package_description(package, package)}", self.do_animation
--- a/src/pipx/venv.py
+++ b/src/pipx/venv.py
@@ -40,7 +40,6 @@ from pipx.venv_inspect import VenvMetadata, inspect_venv
 
 logger = logging.getLogger(__name__)
 
-# This regex was only introduced into importlib.metadata in Python 3.9.
 _entry_point_value_pattern = re.compile(
     r"""
     ^(?P<module>[\w.]+)\s*
@@ -361,6 +360,8 @@ class Venv:
             exec_app([str(self.bin_path / filename)] + app_args)
 
         # Evaluate and execute the entry point.
+        # TODO: After dropping support for Python < 3.9, use
+        # "entry_point.module" and "entry_point.attr" instead.
         match = _entry_point_value_pattern.match(entry_point.value)
         assert match is not None, "invalid entry point"
         module, attr = match.group("module", "attr")
--- a/src/pipx/venv.py
+++ b/src/pipx/venv.py
@@ -345,8 +345,11 @@ class Venv:
     def _find_entry_point(self, app: str) -> Optional[EntryPoint]:
         if not self.python_path.exists():
             return None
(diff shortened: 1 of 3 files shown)


uranusjr:
TODO: Write documentation on this.

src/pipx/venv.py
```diff
@@ -34,6 +40,16 @@
 
 logger = logging.getLogger(__name__)
 
+# This regex was only introduced into importlib.metadata in Python 3.9.
```
henryiii, line 43: You could use `sys.version_info < (3, 9)` instead of the try/except above, then require importlib_metadata for < 3.9 instead of < 3.8. Just a thought.

uranusjr: The regex supplied in `importlib.metadata` on >=3.9 is private; `importlib.metadata` uses it to implement additional properties on the `EntryPoint` class so `module` `attr` `extras` can be accessed directly. I’d prefer not to use it since it makes the fallback implementation needlessly complicated.

uranusjr: I’ve rewritten the comment to better describe the intention.

itsayellow:
Can you give an example of what a `setup.cfg` would look like to define a pipx run entry point?

uranusjr:
This is the configuration I’m using to test this (against pypa/build):

https://github.com/uranusjr/build/blob/5bcd9f2fc6b3560e68e907352064466fe3253de2/setup.cfg#L39-L43

itsayellow:
OK, thanks.

So the idea is that one would need to make sure that the packages PyPI Package Name is the name of the pipx.run endpoint?  Like:
```
pipx.run =
    <pypi_package_name> = <package_name>.__main__:entrypoint
```

Is there a way to eliminate `<pypi_package_name>` in the example?  So that people have one less opportunity to make a mistake and make this stop working for them, when really there's only one possible value for that string?

uranusjr:
Unfortunately, each entry point must have a group and a name, so it’s not possible to write

```ini
[options.entry_points]
pipx.run = <import-path>
```

The best possible is something like

```ini
[options.entry_points]
pipx =
   run = <import-path>
```

or (using an empty string as key)

```ini
[options.entry_points]
pipx.run =
    = <import-path>
```

henryiii:
That's how entry points work. What you would like to do is write:

```cfg
[options.entry_points]
console_scripts =
     build = build.__main__:entrypoint
```

But that would be too common of a name and would break things. So now you can write:

```
[options.entry_points]
console_scripts =
    pyproject-build = build.__main__:entrypoint
pipx.run =
    build = build.__main__:entrypoint
```

Now only "pyproject-build" shows up in the console, but `pip run build` works. It's exactly the same thing as what you'd put in `console_scripts`. By the way, there's no requirement that a package only supply it's PyPI username as a package (even though _very_ common - think of something like pkg_resources from setuptools), and you could even have a custom entrypoint that could only be accessed by `run --spec build my-special-pipx-command` or something.

henryiii:
Using run or an empty string (or anything not unique to the package) would likely be a disaster. If you have a dependency that _also_ had a pipx run entry point, the two would be merged into an unordered list (I believe - that's how Numba works). You _want_ the similarity to `console_scripts`.

henryiii:
You are also not limited to matching the package name; you could add multiple commands. For example, if a project did not want to provide console_scripts (say it only wanted to provide a `python -m ` interface (cough, `rich`, cough, cough), it could provide a series of `pipx.run` entry points, not just it's package name, accessible via `--spec <package_name> <pipx.run entrypoint name>`.

uranusjr:
importlib.metadata has the information which entry point comes from which package, so it's possible to choose the correct entry point when there are multiple matches. The implementation would be a bit more complicated than the current PR, but it's doable.

henryiii:
I didn't know that; good to know! I still think that the current implementation is correct, since you are literally just proving a `console_scripts` for `pipx run` only, but some of my arguments are no longer valid. I still think the strongest point is the symmetry with `console_scripts` - authors don't need to learn something new.

uranusjr:
Yeah, I think the current approach is more balanced, simple to implement while also easy enough for the user to use. The user experience is not perfect, but improvements would also bring drawbacks that may prove problematic.

itsayellow:
> Yeah, I think the current approach is more balanced, simple to implement while also easy enough for the user to use. The user experience is not perfect, but improvements would also bring drawbacks that may prove problematic.

That sounds right, I was just wondering if there was an easy way to eliminate what could be a possible gotcha for package creators.  Naming it properly should probably be pointed out in the docs.

itsayellow:
So, just to double-check on how the interface is working...

If `build` has a `pipx.run` endpoint, then I'm assuming the following will not work:
```
pipx run bUiLd --help
```
It will download and install `build` properly from PyPI (I believe) but it won't match the likely `pipx.run` endpoint name of "build".

I feel like we should canonicalize the package name and/or endpoint name before matching them because of this, and any random capitalization or punctuation for example.

uranusjr:
I believe the same is true with console scripts as well though? I agree canonicalisation is the right thing to do, but it is arguably a separate enhancement.

itsayellow:
> I believe the same is true with console scripts as well though? I agree canonicalisation is the right thing to do, but it is arguably a separate enhancement.

The difference to my mind is that console scripts are between the packager and the user only, but here pipx is part of it.  And the other parts of pipx behave in a way that different capitalizations, punctuations don't matter if they are equivalent canonically.

uranusjr:
I filed #618 for the name normalisation issue.

itsayellow:
> TODO: Write documentation on this.

Does this mean you're working on documentation for this PR?

uranusjr:
I’m not actively writing the documentation for this.

src/pipx/venv.py
```diff
@@ -327,8 +342,40 @@ def list_installed_packages(self) -> Set[str]:
         pip_list = json.loads(cmd_run.stdout.strip())
         return set([x["name"] for x in pip_list])
 
+    def _find_entry_point(self, app: str) -> Optional[EntryPoint]:
+        if not self.python_path.exists():
+            return None
+        site_packages = get_site_packages(self.python_path)
+        for dist in Distribution.discover(path=[str(site_packages)]):
```
itsayellow, line 349: How about limiting our check for the `pipx.run` entrypoint to only the main package?

Otherwise if we found a dependency with a `pipx.run` entrypoint having the app name, we would match it instead.  This sounds like a dumb thing to happen, but I've seen dumb things in the past (like a package using two dependencies that both use the same name for console scripts).

uranusjr: Good idea, added.

uranusjr: Hmm, this does not really work well after I think about this more, because `pipx run` does not really know what the “main package” is in a cached venv. It’s wrong to use the `app` argument, that would make usages like `pipx run tox4 --spec "tox>=4a0"` fail.

itsayellow: Don't we have that information in the metadata?  e.g. `self.main_package_name`.

itsayellow: You could even use `pipx.venv_inspect.get_dist` using `self.main_package_name` as the first argument.

uranusjr: Oh, I didn’t know that! Let me try it out.

uranusjr: That seems to work. PR updated.

itsayellow: Is `importlib.metadata.Distribution.discover()` documented anywhere?  I'm only finding it by browsing the `importlib.metadata` source.  It would probably be useful to use in `pipx.venv_inspect`.

itsayellow: Also, just out of curiosity, is there a way you would get multiple `dists` from from `Distribution.discover()` if you searched for the name of the package?

itsayellow (approved):
LGTM.  Of course still need to fix the conflict but otherwise, nice!

src/pipx/venv.py
uranusjr: I don’t think `Distribution.discover()` is documented. The author told me about it when I was investigating the possibility of using importlib.metadata in pip (to replace `pkg_resources`).

> Is there a way you would get multiple dists from from Distribution.discover() if you searched for the name of the package?

Yes, because it is possible to have multiple installations of a package in different `sys.path` entries. For example:

```
pip install build
pip install --user build
```

would give you two `build` distributions, one under the system site-packages, and the other in the [per-user site-packages](https://www.python.org/dev/peps/pep-0370/). Another possible case is [enabling system site-packages](https://www.python.org/dev/peps/pep-0405/#isolation-from-system-site-packages) in a virtual environment

henryiii:
Adding docs in "developing for pipx" would be nice.

bfox85 (commented):
Good

src/pipx/venv.py
```diff
@@ -327,8 +342,43 @@ def list_installed_packages(self) -> Set[str]:
...
+    def _find_entry_point(self, app: str) -> Optional[EntryPoint]:
+        if not self.python_path.exists():
+            return None
+        dists = Distribution.discover(
+            name=self.main_package_name,
+            path=[str(get_site_packages(self.python_path))],
+        )
+        for dist in dists:
+            for ep in dist.entry_points:
+                if ep.group == "pipx.run" and ep.name == app:
```
chrysle, line 354: @uranusjr I've just encountered a case where the entry point would not be found due to a path given as app. Can we just drop the requirement that the entry point name must exactly match the application name, and pick any entry point found under the `pipx.run` table?

henryiii: What if there is more than one? Say "a" depends on "b", and they both have entry points? (If there's only one, seems fine, just like we already do for one script)

PR #619, opened 2021-02-02
Mention that pipx requires pip 19+ to install

uranusjr:
Technically pipx *can* be installed with earlier pip versions if directly from wheel, or from source with setuptools 30.3+ and wheel installed, but that description is probably too confusing for most.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ pipx ensurepath
 
 Upgrade pipx with `brew update && brew upgrade pipx`.
 
-Otherwise, install via pip:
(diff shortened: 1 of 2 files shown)


tucked (approved):

itsayellow (changes requested):
Could you also make the same change in the Install portion of `README.txt` please.

uranusjr:
Thanks, added.

itsayellow (approved):
awesome, thanks.

PR #624, opened 2021-02-07
Speed up `pipx list` by removing multiprocessing code

### LINKED ISSUE #616: De-parallelize `pipx list`

`pipx list` uses `multiprocessing.Pool` to parallelize venv analyses.  On my system, this **always takes longer** than if I force the code to run sequentially by inserting the line `Pool = None` before the `if Pool:` statement.

I've tested with 59 venvs, and I've tested with only a few venvs, and always the sequential for-loop is faster than the `multiprocessing.Pool` method.

This could be because I have an SSD drive, but it makes me wonder in general if using multiprocessing for this is worth it, or if it's often worse than the simpler sequential method.

On my test case with 59 pipx-installed packages:
`time pipx list`: 3.13s (multiprocessing code) 0.88s (sequential code)

On my test case with 7 pipx-installed packages:
`time pipx list`: 1.84s (multiprocessing code) 0.15s (sequential code)

I'm curious if other people find the same behavior on other systems.

Get rid of `multiprocessing` code in `command/list_packages.py`, only use the fallback sequential code.

itsayellow:

## Summary of changes
Since we are not compute-limited in `pipx list`, parallel code is not appropriate, and in fact makes running pipx list noticeably slower than purely sequential code.

See #616 for some timing benchmarks

This PR removes the multiprocessing code for `pipx list`, and makes it always use fully sequential code.

Closes #616 

## Test plan

Tested by running

Unix / macOS;
```shell
> time pipx list
```

Windows:
```shell
> Measure-Command { pipx list }
```

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -3,7 +3,7 @@ dev
 - Fix cursor show/hide to work with older versions of Windows. (#610)
 - Support text colors on Windows. (#612)
 - Better platform unicode detection to avoid errors and allow showing emojis when possible. (#614)
-- Sped up `pipx list`.
+- Sped up `pipx list` (#624).
 
 0.16.0.0
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -3,6 +3,7 @@ dev
 - Fix cursor show/hide to work with older versions of Windows. (#610)
 - Support text colors on Windows. (#612)
 - Better platform unicode detection to avoid errors and allow showing emojis when possible. (#614)
+- Sped up `pipx list`.
 
 0.16.0.0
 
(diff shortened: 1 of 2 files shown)


uranusjr:
Semi-related, it seems like multiprocessing was used because the implementation used to call `pip list`. This also left a superfulous `python_path` argument along the call stack that can be removed in a follow-up refactoring.

uranusjr:
I’ve filed #626 to remove the unused variable.

itsayellow:
> Semi-related, it seems like multiprocessing was used because the implementation used to call `pip list`. This also left a superfulous `python_path` argument along the call stack that can be removed in a follow-up refactoring.

Ahh, that makes more sense.  Thanks for going back in the history and finding that out.

itsayellow:
This seems to be the actual commit when @cs01 made `list_packages()` parallel: af61325a0e6d1ac67959e8e28ee0a6d31d9a1eed

PR #626, opened 2021-02-08
Remove unused python_path and related checks

uranusjr:
See https://github.com/pipxproject/pipx/pull/624#issuecomment-774955205

### DIFF
--- a/src/pipx/commands/common.py
+++ b/src/pipx/commands/common.py
@@ -213,7 +213,6 @@ def get_package_summary(
     return (
         _get_list_output(
             python_version,
-            python_path,
             package_metadata.package_version,
             package,
             new_install,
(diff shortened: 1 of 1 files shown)


itsayellow:
I think this code is actually doing something useful, and maybe it shouldn't be removed.  (I may have added it?)  It's not that we need the python path for the listing process, but it's one of the checks that `pipx list` does to make sure that each venv doesn't have any problems.

uranusjr:
At least the check in `_get_list_output` can be removed though. It is unreachable because all invalid cases are caught in the other check.

uranusjr:
I’ve restored the check in `get_package_summary`; only the check in `_get_list_output` is removed in this PR now.

itsayellow (approved):
Right, this looks good.

itsayellow:
BTW, I'm letting you merge your own PRs once they've been approved.  That's the convention we've been using.

PR #634, opened 2021-02-20
In `reinstall-all` if shared_lib python is invalid, create it instead of upgrading to avoid error.

### LINKED ISSUE #633: Updating Python from 3.9.1_4 to 3.9.1_8 causes No such file or directory
Yesterday, I ran brew upgrade. After that, my shell (xonsh), as installed by pipx, stopped running. It would fail with:

```
login: /Users/jaraco/.local/bin/xonsh: No such file or directory
```

So I load up bash and inspect:

```
$ ls -la ~/.local/pipx/venvs/xonsh/bin/python3.9
lrwxr-xr-x   1 jaraco  primarygroup    69 Jan  5 11:39 python3.9 -> /Users/jaraco/.local/homebrew/Cellar/python@3.9/3.9.1_4/bin/python3.9
```

It seems `python3.9` no longer points to a valid Python, and thus all pipx commands are now broken.

I was able to `reinstall-all`, which worked, but also emitted an error:

```
$ pipx reinstall-all
Failed to upgrade shared libraries
  File "/Users/jaraco/.local/homebrew/lib/python3.9/site-packages/pipx/shared_libs.py", line 92, in upgrade
    upgrade_process = run_subprocess(
  File "/Users/jaraco/.local/homebrew/lib/python3.9/site-packages/pipx/util.py", line 123, in run_subprocess
    completed_process = subprocess.run(
  File "/Users/jaraco/.local/homebrew/Cellar/python@3.9/3.9.1_8/Frameworks/Python.framework/Versions/3.9/lib/python3.9/subprocess.py", line 501, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/Users/jaraco/.local/homebrew/Cellar/python@3.9/3.9.1_8/Frameworks/Python.framework/Versions/3.9/lib/python3.9/subprocess.py", line 947, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/Users/jaraco/.local/homebrew/Cellar/python@3.9/3.9.1_8/Frameworks/Python.framework/Versions/3.9/lib/python3.9/subprocess.py", line 1819, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: '/Users/jaraco/.local/pipx/shared/bin/python'
uninstalled oathtool! ✨ 🌟 ✨
...
done! ✨ 🌟 ✨
$ ls -la ~/.local/pipx/venvs/xonsh/bin/python3.9
lrwxr-xr-x  1 jaraco  primarygroup  69 Feb 19 09:25 /Users/jaraco/.local/pipx/venvs/xonsh/bin/python3.9 -> /Users/jaraco/.local/homebrew/Cellar/python@3.9/3.9.1_8/bin/python3.9
```

I have pipx 0.16.0.0.

Is it the case that a `reinstall-all` is expected to be needed between patch updates to Python? I haven't experienced this issue on my other mac where I have Homebrew installed in the canonical location. Is there anything that one can do to avoid this failure when running `brew upgrade`?

itsayellow:

## Summary of changes

Closes #633.

#554 introduced a regression where `pipx reinstall-all` would try to upgrade the shared_libs without verifying that they were still valid (i.e. their python link still existed).  If the python did not exist, this would output an error.

This PR first checks if the shared_libs are valid: if they are, it forces an upgrade, if they are not, it instead creates (or re-creates) the shared libs.

## Test plan

Tested by running
```
rm ~/.local/pipx/shared/bin/python3
ln -s /path/that/does/not/exist ~/.local/pipx/shared/bin/python3
pipx reinstall-all
```

This PR:
(no error)

pipx 0.16.0.0:
```
Failed to upgrade shared libraries
  File "/usr/local/Cellar/pipx/0.16.0.0/libexec/lib/python3.9/site-packages/pipx/shared_libs.py", line 92, in upgrade
    upgrade_process = run_subprocess(
  File "/usr/local/Cellar/pipx/0.16.0.0/libexec/lib/python3.9/site-packages/pipx/util.py", line 123, in run_subprocess
    completed_process = subprocess.run(
  File "/usr/local/Cellar/python@3.9/3.9.1_8/Frameworks/Python.framework/Versions/3.9/lib/python3.9/subprocess.py", line 501, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/local/Cellar/python@3.9/3.9.1_8/Frameworks/Python.framework/Versions/3.9/lib/python3.9/subprocess.py", line 947, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/local/Cellar/python@3.9/3.9.1_8/Frameworks/Python.framework/Versions/3.9/lib/python3.9/subprocess.py", line 1819, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: '/Users/mclapp/.local/pipx/shared/bin/python'

[...]
```

### DIFF
--- a/src/pipx/commands/reinstall.py
+++ b/src/pipx/commands/reinstall.py
@@ -84,7 +84,10 @@ def reinstall_all(
     skip: Sequence[str],
 ) -> ExitCode:
     """Returns pipx exit code."""
-    pipx.shared_libs.shared_libs.upgrade(verbose=verbose)
+    if pipx.shared_libs.shared_libs.is_valid:
+        pipx.shared_libs.shared_libs.upgrade(verbose=verbose)
+    else:
+        pipx.shared_libs.shared_libs.create(verbose=verbose)
 
     failed: List[str] = []
     for venv_dir in venv_container.iter_venv_dirs():
--- a/src/pipx/commands/reinstall.py
+++ b/src/pipx/commands/reinstall.py
@@ -84,10 +84,7 @@ def reinstall_all(
     skip: Sequence[str],
 ) -> ExitCode:
     """Returns pipx exit code."""
-    if pipx.shared_libs.shared_libs.is_valid:
-        pipx.shared_libs.shared_libs.upgrade(verbose=verbose)
-    else:
-        pipx.shared_libs.shared_libs.create(verbose=verbose)
+    pipx.shared_libs.shared_libs.upgrade(verbose=verbose)
 
     failed: List[str] = []
     for venv_dir in venv_container.iter_venv_dirs():
--- a/src/pipx/shared_libs.py
+++ b/src/pipx/shared_libs.py
@@ -73,6 +73,10 @@ class _SharedLibs:
     def upgrade(
         self, *, pip_args: Optional[List[str]] = None, verbose: bool = False
     ) -> None:
+        if not self.is_valid:
+            self.create(verbose=verbose)
+            return
+
         # Don't try to upgrade multiple times per run
         if self.has_been_updated_this_run:
             logger.info(f"Already upgraded libraries in {self.root}")
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -5,7 +5,7 @@ dev
 - Better platform unicode detection to avoid errors and allow showing emojis when possible. (#614)
 - Don't emit show cursor or hide cursor codes if STDERR is not a tty. (#620)
 - Sped up `pipx list` (#624).
-- For `reinstall-all`, fixed bug where missing python executable would cause error.
+- For `reinstall-all`, fixed bug where missing python executable would cause error. (#634)
 
 0.16.0.0
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -5,6 +5,7 @@ dev
 - Better platform unicode detection to avoid errors and allow showing emojis when possible. (#614)
 - Don't emit show cursor or hide cursor codes if STDERR is not a tty. (#620)
 - Sped up `pipx list` (#624).
+- For `reinstall-all`, fixed bug where missing python executable would cause error.
 
 0.16.0.0
 


src/pipx/commands/reinstall.py
```diff
@@ -84,7 +84,10 @@ def reinstall_all(
     skip: Sequence[str],
 ) -> ExitCode:
     """Returns pipx exit code."""
-    pipx.shared_libs.shared_libs.upgrade(verbose=verbose)
+    if pipx.shared_libs.shared_libs.is_valid:
```
cs01, line 87: What do you think about moving this check to the intervals of the upgrade method? That was any other calls to upgrade wouldn’t be subject to this issue either.

itsayellow: Yeah I thought about that.  It's probably good to do.

I hesitated to do that because it makes the code inside of `shared_libs.py` look almost like a circular reference (even though it's not) because `create` falls back to `upgrade`, and then with that code `upgrade` would fall back to `create`.  It ultimate works out ok in fact.

itsayellow: OK, made that fix. 👍

src/pipx/shared_libs.py
```diff
@@ -73,6 +73,10 @@ def needs_upgrade(self) -> bool:
     def upgrade(
         self, *, pip_args: Optional[List[str]] = None, verbose: bool = False
     ) -> None:
+        if not self.is_valid:
```
uranusjr, line 76: `is_valid` would still return True if a Windows venv’s underlying interpreter goes away. I wonder if we should actually try to execute the virtual environment instead (e.g. run `python -c "import sys; print(sys.executable)"`). This can probably be fixed separately though (when someone complains about it).

itsayellow: > Also `create()` already has an `is_valid` guard.

I think that should be ok, do you see a problem?  I put this one in `upgrade` for when external code calls `upgrade` itself, not for when `create()` calls `upgrade()`.

uranusjr: If you ask me, `create()` probably shouldn’t have a `is_valid` guard, since it feels weird to me calling `create()` may not actually create anything. But removing that guard probably has consequences.

itsayellow: Ah that's a different matter.  I think it was meant as an optimization to save time.  We use `shared_libs.create()` every time we create a new venv.  I think probably this is meant to prevent having to recreate the shared_libs if it's already fine.

We could move the guard to the calling code in `venv.py`, which would functionally be the same thing just factored differently.  Personally I'm ok either way.

uranusjr: Let’s keep this guard here and aim to remove the one in `create()` then.

uranusjr (approved):

PR #643, opened 2021-03-08
Naming fixes

itsayellow:

## Summary of changes
When working on other branches I noticed some areas of the code that could be made clearer through renaming.  Adding them in their own PR to avoid cluttering future PRs.

Renaming clarifications:
* `package` -> `package_name`
* `get_package_summary` -> `get_venv_summary` (returns info about the entire venv, not just one package)

Also, mypy seemed to occasionally actually crash due to the `RelevantSearch` definition in `util.py`.  This was either because of its use of existing keywords (`re`) or because it was a class defined inside a function, or possibly both together.  I've broken the class definition of `RelevantSearch` out to the top-level of the module and avoided existing keywords.

## Test plan

Tested by running
```
```

### DIFF
--- a/src/pipx/commands/common.py
+++ b/src/pipx/commands/common.py
@@ -156,18 +156,18 @@ def _symlink_package_apps(
             )
 
 
-def get_package_summary(
+def get_venv_summary(
     venv_dir: Path,
     *,
-    package: str = None,
+    package_name: str = None,
     new_install: bool = False,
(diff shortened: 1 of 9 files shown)


decodedfile:
Duplicate of #

itsayellow:
I tried to especially modify all the `package` variable names adjacent to `package_or_url` to help illustrate the distinction.

PR #648, opened 2021-03-19
Add type annotations

itsayellow:

## Summary of changes

Adding some missing or corrected type annotations found with `mypy --strict`.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -123,7 +123,13 @@ def lint(session):
     session.run("isort", "--check", "--diff", "--profile", "black", *files)
     session.run("black", "--check", *files)
     session.run("flake8", *files)
-    session.run("mypy", *files)
+    session.run(
+        "mypy",
+        "--strict-equality",
+        "--no-implicit-optional",
+        "--warn-unused-ignores",
+        *files,
+    )
     session.run("check-manifest")
     session.run("python", "setup.py", "check", "--metadata", "--strict")
 
--- a/src/pipx/commands/common.py
+++ b/src/pipx/commands/common.py
@@ -187,7 +187,7 @@ def get_venv_summary(
 
     package_metadata = venv.package_metadata[package_name]
 
-    if package_metadata.package_version is None:
+    if package_metadata.package_version == "":
         return (
             f"   package {red(bold(package_name))} {red('is not installed')} "
             f"in the venv {venv_dir.name}",
--- a/src/pipx/commands/common.py
+++ b/src/pipx/commands/common.py
@@ -159,7 +159,7 @@ def _symlink_package_apps(
 def get_venv_summary(
     venv_dir: Path,
     *,
-    package_name: str = None,
+    package_name: Optional[str] = None,
     new_install: bool = False,
     include_injected: bool = False,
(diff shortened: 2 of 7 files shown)


cs01:
Cool, I never saw that flag before. Should it be added to the nox file as well?

itsayellow:
> Cool, I never saw that flag before. Should it be added to the nox file as well?

We still don't satisfy all of what `mypy --strict` reports.  For some things we might have to end up adding more `type: ignore` for where a type doesn't exist.  

For example, whatever argparse returns from `ArgumentParser.add_subparsers()` is a bit of a mystery to me.  I tried to track it down before, but I'm not sure there's a public type for it (I think there may be some odd internal class that is not meant to be used outside of the argparse package.)

Other than that there are some interesting cases that would require some work to make pass.

`--strict` effectively is like including a list of other mypy switches.  We could always add some of those individually.
* --warn-unused-configs
* --disallow-any-generics
* --disallow-subclassing-any
* --disallow-untyped-calls
* --disallow-untyped-defs
* --disallow-incomplete-defs
* --check-untyped-defs
* --disallow-untyped-decorators
* --no-implicit-optional
* --warn-redundant-casts
* --warn-unused-ignores
* --warn-return-any
* --no-implicit-reexport
* --strict-equality

Another thing I don't understand is what type parameters `CompletedProcess` should get.
```
src/pipx/venv.py:433: error: Missing type parameters for generic type "CompletedProcess"
```

itsayellow:
Ah!  I finally found where the source for the stdlibs.  Here's the source for CompletedProcess:
https://github.com/python/typeshed/blob/a68fb507889062601d345a28e95fb5dc0cec1e0b/stdlib/subprocess.pyi#L34

That helps a lot.

itsayellow:
Looks like adding a type parameter to `CompletedProcess` only works on python3.9.  That's another thing we'd have to do to satisfy `--strict` type-checking, have workarounds for different python versions.

itsayellow:
Also, there is a type that `ArgumentParser.add_subparsers()` returns (`_SubParsersAction`) but it is officially "undocumented".
https://github.com/python/typeshed/blob/a68fb507889062601d345a28e95fb5dc0cec1e0b/stdlib/argparse.pyi#L446

cs01:
Ah okay. Yeah that makes sense. Types in Python can be pretty tricky, especially since some of the functionality changes across Python versions.

itsayellow:
In this PR I could add a couple of the mypy switches above to our nox.

What do you think about adding these?
* `--strict-equality` - Prohibit equality, identity, and container checks for non-overlapping types (inverse: --no-strict-equality)
* `--no-implicit-optional` - Don't assume arguments with default values of None are Optional (inverse: --implicit-optional)
* `--warn-unused-ignores` - Warn about unneeded '# type: ignore' comments (inverse: --no-warn-unused-ignores)

cs01:
NIce, I’m all for it. The stricter the better IMO.

itsayellow:
I just found `mypy --warn-unreachable`.  Very handy!  It lets you know if some code is "inferred to be unreachable".

PR #651, opened 2021-03-21
Fix cleanup of app binaries for venvs without pipx_metadata.json

itsayellow:

## Summary of changes
Found this code bug with `mypy --warn-unreachable`.

This PR fixes a bug with `pipx uninstall` that would leave app binaries leftover in `~/.local/bin` if the venv has no metadata file.

We were testing the wrong object to determine if metadata was present in `uninstall.py`.  `venv.pipx_metadata.main_package` is in fact always not `None`, (it's always `PackageInfo`) so the else (for missing metadata case) was impossible to execute.  Thus the special cleanup code for the non-metadata case to remove app binaries never executed.

I fixed it by changing the test for `None` to the object `venv.pipx_metadata.main_package.package`.

In addition, I updated `venv.get_venv_metadata_for_package` to add its currently-necessary second argument.  (This code must have been unreachable for a while to not get flagged that it is missing an argument.)

## Test plan

Tested by running
```shell
pipx install pycowsay
rm ~/.local/pipx/venvs/pycowsay/pipx_metadata.json
pipx uninstall pycowsay
```

This PR:
```shell
> ls -l ~/.local/bin/pycowsay
ls: /Users/mclapp/.local/bin/pycowsay: No such file or directory
```

pipx 0.16.1.0
```shell
> ls -l ~/.local/bin/pycowsay                     
lrwxr-xr-x  1 mclapp  staff  53 Mar 19 22:54 /Users/mclapp/.local/bin/pycowsay@ -> /Users/mclapp/.local/pipx/venvs/pycowsay/bin/pycowsay

```

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,7 +1,7 @@
 dev
 
 - Make sure log files are utf-8 encoded to preven Unicode encoding errors from occurring with emojis. (#646)
-- Fixed old regression that would prevent pipx uninstall from cleaning up linked binaries if the venv was old and did not have pipx metadata.
+- Fixed old regression that would prevent pipx uninstall from cleaning up linked binaries if the venv was old and did not have pipx metadata. (#651)
 
 0.16.1.0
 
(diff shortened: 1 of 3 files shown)


itsayellow:
I suppose I should be a "grown up" and add a test for this large block of code that isn't being tested...

PR #655, opened 2021-03-25
Add more types to codebase, update pre-commit mypy options

itsayellow:

## Summary of changes
This adds more types to more successfully satisfy `mypy --strict`.  It also brings the mypy options in `pre-commit-config.yml` in sync with the mypy options in `noxfile.py`.

Using string literal types is a useful way to take advantage of some parametric types without breaking previous python versions runtimes.  (e.g. `"Callable[str]"`)

I used the underscore-marked private object `argparse._SubParsersAction` to type some functions.  I have seen other projects use the same type, so I decided it was probably ok.

Unfortunately, `argcomplete` is untyped and does tricks to the argparse objects that are not type-announced, so I couldn't find a way to properly type a few subparser objects without extensive gymnastics, so I let them be.

## Test plan

Tested by running
```
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,3 +32,4 @@ repos:
   rev: v0.800
   hooks:
   - id: mypy
+    args: ['--strict-equality','--no-implicit-optional','--warn-unused-ignores']
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -13,7 +13,7 @@ import textwrap
 import time
 import urllib.parse
 from pathlib import Path
-from typing import Dict, List
+from typing import Any, Callable, Dict, List
 
 import argcomplete  # type: ignore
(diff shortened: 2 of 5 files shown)


itsayellow:
6 remaining issues preventing us passing `mypy --strict`
```shell
> mypy --strict src/pipx
src/pipx/commands/run.py:207: error: Returning Any from function declared to return "str"
src/pipx/main.py:331: error: Function is missing a type annotation for one or more arguments
src/pipx/main.py:362: error: Function is missing a type annotation for one or more arguments
src/pipx/main.py:405: error: Function is missing a type annotation for one or more arguments
src/pipx/main.py:424: error: Function is missing a type annotation for one or more arguments
src/pipx/main.py:551: error: Function is missing a type annotation for one or more arguments
Found 6 errors in 2 files (checked 26 source files)
```

PR #674, opened 2021-04-18
Make clear_line() in animate.py work reliably

### LINKED ISSUE #673: clear_line() isn't working on Windows Powershell
**Describe the bug**

I think this may be an "upstream" bug, but I just wanted to note it.

If `colorama.init()` has been executed in pipx (normal operation), executing `pipx.animate.clear_line()` does not actually erase the current line on Powershell in Windows.

I commented out the `colorama.init()` statement and on Windows powershell, the `clear_line()` functions properly and erases the final animation frame.

**How to reproduce**

On Windows Powershell:
```shell
pipx install nox
pipx reinstall-all
```

What is output is that "nox" is short enough that one can tell that "upgrading shared libraries" is not erased before "uninstalled nox!" (see the "ies" still visible past the stars):

```shell
> pipx reinstall-all
uninstalled nox! ✨ 🌟 ✨ies
  installed package nox 2020.12.31, Python 3.9.4
    - nox.exe
    - tox-to-nox.exe
done! ✨ 🌟 ✨
>
```

itsayellow:

## Summary of changes
Closes #673 

I noticed that `clear_line()` in `animate.py` wasn't working on Windows.  

It appears that the code wasn't structured properly.  The escape code sequence used is supposed delete everything from the current cursor position to the end of the line, and at the time it was invoked the cursor was at the end of the line, so it only deleted from there to the end of the line, i.e. deleted not much of anything.

I modified the code so that `clear_line()` first prints a linefeed character, positioning the cursor at the start of the line before executing the escape code to clear the line from cursor to end of line.  We always wanted to print a linefeed anyway when we executed `clear_line()`, so it worked out well.  The result is just putting linefeed before `CLEAR_LINE` instead of after.

In addition, I had to add a `sys.stderr.flush()` to the `animate` context manager code.  With the new `clear_line()`, the buffering tended to just show an empty line instead of the non-empty characters.  (This was possibly because `clear_line()` is actually working now.)  Flushing after printing the animation frame and before waiting ensures that the frame will be visible during the wait.

### DIFF
--- a/src/pipx/animate.py
+++ b/src/pipx/animate.py
@@ -68,8 +68,6 @@ def animate(
     finally:
         event.set()
         clear_line()
-        sys.stderr.write("\r")
-        sys.stdout.write("\r")
 
 
 def print_animation(
@@ -95,8 +93,8 @@ def print_animation(
                 cur_line = f"{message:.{max_message_len}}{s}"
 
             clear_line()
-            sys.stderr.write("\r")
             sys.stderr.write(cur_line)
+            sys.stderr.flush()
             if event.wait(period):
                 break
 
--- a/src/pipx/animate.py
+++ b/src/pipx/animate.py
@@ -128,5 +128,6 @@ def show_cursor() -> None:
 
 
 def clear_line() -> None:
+    """Clears current line and positions cursor at start of line"""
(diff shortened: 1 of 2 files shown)


cs01:
Awesome

PR #680, opened 2021-04-29
Change `pipx list` warnings and info to output on stderr

### LINKED ISSUE #678: pipx list --json can print non-json to stdout
When nothing is installed, `pipx list --json` prints `nothing has been installed with pipx 😴` to stdout, and that's not valid json.

_Originally posted by @liskin in https://github.com/pipxproject/pipx/issues/660#issuecomment-828473735_

In addition, any warnings about the venvs are also printed to stdout, which is also non-valid json!

itsayellow:

## Summary of changes
Closes #678 

I changed all informational messages from `pipx list` to output to stderr.  Actual warning messages use `logger.warning()`.

This needs to be fixed for `pipx list --json` to work properly and output valid json.

Let me know if there's any reason that a user might want these info messages on stdout, perhaps for a normal non-json `pipx list`.  I personally don't believe so, but I might be wrong. 🙂

## Test plan

Tested by running
```
```

### DIFF
--- a/src/pipx/commands/common.py
+++ b/src/pipx/commands/common.py
@@ -168,22 +168,25 @@ def venv_health_check(
     if not python_path.is_file():
         return (
             VenvProblems(invalid_interpreter=True),
-            f"   package {red(bold(venv_dir.name))} has invalid interpreter {str(python_path)}",
+            f"{hazard}  package {red(bold(venv_dir.name))} has invalid "
+            "interpreter {str(python_path)}",
         )
     if not venv.package_metadata:
         return (
             VenvProblems(missing_metadata=True),
-            f"   package {red(bold(venv_dir.name))} has missing internal pipx metadata.",
+            f"{hazard}  package {red(bold(venv_dir.name))} has missing "
+            "internal pipx metadata.",
         )
     if venv_dir.name != canonicalize_name(venv_dir.name):
         return (
             VenvProblems(bad_venv_name=True),
-            f"   package {red(bold(venv_dir.name))} needs its internal data updated.",
+            f"{hazard}  package {red(bold(venv_dir.name))} needs its "
+            "internal data updated.",
         )
     if venv.package_metadata[package_name].package_version == "":
         return (
             VenvProblems(not_installed=True),
-            f"   package {red(bold(package_name))} {red('is not installed')} "
+            f"{hazard}  package {red(bold(package_name))} {red('is not installed')} "
             f"in the venv {venv_dir.name}",
         )
     return (VenvProblems(), "")
--- a/src/pipx/commands/common.py
+++ b/src/pipx/commands/common.py
@@ -168,26 +168,26 @@ def venv_health_check(
     if not python_path.is_file():
         return (
             VenvProblems(invalid_interpreter=True),
-            f"{hazard}  package {red(bold(venv_dir.name))} has invalid "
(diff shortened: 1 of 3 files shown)


itsayellow:
A lot of times our warning messages have a ⚠️ emoji.  Should that be added to any of these?  Or does that make the text too busy?

itsayellow:
I added a hazard ⚠️ emoji to the specific venv warning messages.  Still not sure if the paragraphs at the end supplying guidance need that emoji or not.

e.g.
```
> pipx list
venvs are in /Users/mclapp/.local/pipx/venvs
apps are exposed on your $PATH at /Users/mclapp/.local/bin
⚠️ package nox has missing internal pipx metadata.
   package pylint 2.8.2, Python 3.9.4
    - epylint
    - pylint
    - pyreverse
    - symilar

One or more packages have a missing internal pipx metadata.
   They were likely installed using a pipx version before 0.15.0.0.
   Please uninstall and install these package(s) to fix.
```

src/pipx/commands/common.py
```diff
@@ -168,22 +168,25 @@ def venv_health_check(
...
             VenvProblems(missing_metadata=True),
-            f"   package {red(bold(venv_dir.name))} has missing internal pipx metadata.",
+            f"{hazard}  package {red(bold(venv_dir.name))} has missing "
+            "internal pipx metadata.",
         )
     if venv_dir.name != canonicalize_name(venv_dir.name):
         return (
             VenvProblems(bad_venv_name=True),
-            f"   package {red(bold(venv_dir.name))} needs its internal data updated.",
+            f"{hazard}  package {red(bold(venv_dir.name))} needs its "
```
cs01, line 183: Add leading white space

cs01 (approved):

src/pipx/commands/common.py
itsayellow: So I was trying to figure this out.  This actually looks really nice on macOS.  The word "package" lines up with the other "package" words on non-hazard lines if you space it this way.  The emoji just ends up to the left of consistently-tabbed text.

Unfortunately the spacing of emoji doesn't seem to be consistent across platforms, so it kind of looks crap on Windows for example. 😞

itsayellow: I'm going to try an experiment with a commit.  I won't merge until we can discuss...

itsayellow: haha that was tricky but I think it actually worked. 😆

itsayellow:
So what I did was to have the same exact text with the same initial spacing as before, but at the end of the printed string I added `\r{hazard}` so it linefeeds back to the beginning and prints a ⚠️ at the beginning.  That way the text maintains the same indent but the hazard appears to the left of the warning message text.

Seem ok?  I know it's a little tricky but I think it works. 🙂

macOS:
Screen Shot 2021-04-28 at 11 23 44 PM

Windows:
win_emoji_hazard

Linux:
ubuntu_emoji_hazard

cs01:
Nice! Works for me. 

How did you test on all the platforms? GitHub actions or did you run all that locally?

itsayellow:
> How did you test on all the platforms? GitHub actions or did you run all that locally?

I use macOS natively and use Parallels with a Windows VM and an Ubuntu VM.  It's really, really handy.

liskin:
Wow that was quick. ♥ 🎉

PR #681, opened 2021-04-29
Create valid json from `pipx list --json` when no venvs installed

### LINKED ISSUE #678: pipx list --json can print non-json to stdout

itsayellow:

## Summary of changes
Closes #678 (again 😄 )

An empty string is not valid json.  This makes sure that we print valid json to stdout for the case of no venvs installed, instead of nothing as we did before to stdout.

## Test plan

Tested by running the following with no venvs installed.
```shell
> pipx list --json > test.out
nothing has been installed with pipx 😴
> more test.out              
{
    "pipx_spec_version": "0.1",
    "venvs": {}
}
```

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,6 @@
 dev
 
+- Fixed `pipx list --json` to return valid json with no venvs installed.  Previously would return and empty string to stdout.
 
 0.16.2.1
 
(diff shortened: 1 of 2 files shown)


cs01 (approved):

ALJHONQ (commented):
Nc

PR #686, opened 2021-05-13
Local pypiserver default for tests

### LINKED ISSUE #248: Offline tests
Packaging for distros requires running the tests in a VM isolated from the internet to avoid any pollution.  Offline tests are needed for this.

However, the test suite can also be quicker if it uses a fake PyPI server that only contains the packages needed for testing.

pipenv project tests uses its own, as a fixture in the repo.

There are some other earlier standalone attempts

https://github.com/uranusjr/pytest-pypi-gateway

https://github.com/uranusjr/pytest-fakepi

https://pypi.org/project/pytest-pypi/

https://github.com/manahl/pytest-plugins has a devpi pytest plugin for https://github.com/devpi/devpi

An old style PyPI HTML `simple` index page.

itsayellow:

## Summary of changes
Closes #248 

*Updated after code changes since PR inception*

This implements running a local pypiserver in our tests, so we use a fixed set of package distribution files and do not have to depend on network access of pypi.org.

There is a master list in `testdata/tests_packages/primary_packages.txt` that defines all packages tests will need to install or inject.  From this master list, platform-specific and python-version-specific lists are made of all packages including dependencies necessary.  I made a nox session for generating the needed package lists (`create_test_package_list`) for all Python versions.  There is also a GitHub workflow that generates package lists needed for every platform and Python version combination.  It creates an artifact with all the platform-python-version-specific lists.

There is a cache directory that is created (`.pipx_tests/package_cache`) to store all of the wheel and tar.gz and zip files needed for the local pypi server for the tests.  At the beginning of the tests a download script makes sure the cache contains all the proper files, downloads missing ones, and deletes any no-longer-used files.  On our Github Workflow `tests` I made `.pipx_tests/package_cache` persistent so our Github CI tests do not need to repopulate it every time.

Performance has improved also, it seems that typically with a local pypi server the tests take about 2 minutes less than they used to.

The old style tests that use the internet to access pypi are still available using `nox -s tests_internet`. To use the old internet pypi.org with a plain pytest invocation use `pytest --net-pypiserver tests`.

## Test plan

Tested by running
```
nox -s tests
```


itsayellow:
Strictly speaking if the `.pipx_tests_cache` is not populated this methodology is not "offline".  But saving all of the *.whl, *.tar.gz, and *.zip files easily dominated the entire repo in file size.  I decided it wasn't worth it to have the tests be 100% off-line.

If `.pipx_tests_cache` is populated then these tests should function completely off-line.

Also, they make the tests completely deterministic, we don't have to worry about an upgrade of a dependency that still matches a package's version requirement changing what we're installing.

itsayellow:
Another minor point I just remembered.  The local pypiserver's log output is directed to a file in the top-level repo directory called `pypiserver.err`.  If somebody has a suggestion of a better place to create that file it would probably be good to remove the clutter of the main pipx directory.

itsayellow:
Right now to use the local pypi server one must invoke pytest using `pytest --pypiserver`, i.e. the default is the old pypi.org internet behavior.

One question is: should the default pytest behavior be to use the local pypiserver?  So instead of need a switch to use the local pypiserver, should one need a switch (like `--net-pypi` or similar) to cause the tests to go to pypi.org?

itsayellow:
Just changed the location of the local pypiserver log to be output at the top level of `.pipx_tests_cache`.

itsayellow:
This is a pretty major change, so I welcome any even general comments or questions.

One concern I had, it seems there is evidence that packagers of pipx (like linux distros, etc.) may use pytest alone (without nox) to run tests.  (See #682 and https://kojipkgs.fedoraproject.org//packages/pipx/0.16.2.1/1.fc35/data/logs/noarch/build.log).  That seems to suggest to me that we should make the pytest default to use the local pypiserver so our nox default and pytest default match.

In addition, I'm wondering if the pytest code should even attempt to download the necessary files into the cache directory if `--pypiserver` is enabled for pytest (and not handle that in nox as of now.)  Or would that be loading too advanced functionality into the pytest tests?

itsayellow:
Just changed the pylint default to use the local pypiserver to match the nox default.

itsayellow:
I've run into a wrinkle where different packages are necessary for `pyobj-framework-*` on the Github macOS machines (v10.15) vs. my macOS machine (v11.3.1).

I wonder if this means that the lists in `testdata/tests_packages/` need to have separate lists for **separate OS versions** besides the OS type and python version which is what is used to differentiate now.

itsayellow:
Added documentation to `contributing.md` explaining how to prepoulate the package cache manually in order to enable offline tests operation.

Also added a direct link from `changelog.md` to the offline test docs.

I think I'm almost ready to call this Ready for review.

itsayellow:
OK Ready for Review.  I think I like the way this shaped up 🙂

gaborbernat (approved):

itsayellow:
Ah, need to update for the new kaggle version.  Working on it...

itsayellow:
@cs01 you're welcome to review as much or as little as you want.  I know you were always looking for this kind of local pypi testing.  If you're good with @gaborbernat's review, just let me know and I'll merge.

cs01 (approved):

itsayellow:
booyah! 🎉

PR #691, opened 2021-05-21
Only install and import colorama on Windows. 

hwalinga:
Related: #612

Only install and import colorama on Windows. Importing and init call has no effect on other platforms, so no need to do it on other platforms.

## Summary of changes

Just check if on Windows, do import and call init. Also, colorama dep is now listed as a windows specific dep. 

## Test plan

Tested by running: 
```
pipx list --verbose
```

### DIFF
--- a/src/pipx/colors.py
+++ b/src/pipx/colors.py
@@ -1,11 +1,13 @@
 import sys
 from typing import Callable
 
-import colorama  # type: ignore
+from pipx.constants import WINDOWS
 
 PRINT_COLOR = sys.stdout.isatty()
 
-if PRINT_COLOR:
+if PRINT_COLOR and WINDOWS:
+    import colorama  # type: ignore
+
     colorama.init()
 
 
--- a/src/pipx/colors.py
+++ b/src/pipx/colors.py
@@ -1,12 +1,12 @@
 import sys
 from typing import Callable
 
-PRINT_COLOR = sys.stdout.isatty()
-
 try:
     import colorama  # type: ignore
 except ImportError:  # Colorama is Windows only package
     colorama = None
+
+PRINT_COLOR = sys.stdout.isatty()
 
 if PRINT_COLOR and colorama:
     colorama.init()
--- a/src/pipx/colors.py
+++ b/src/pipx/colors.py
@@ -1,13 +1,14 @@
 import sys
 from typing import Callable
 
-from pipx.constants import WINDOWS
-
 PRINT_COLOR = sys.stdout.isatty()
 
-if PRINT_COLOR and WINDOWS:
+try:
     import colorama  # type: ignore
+except ImportError:  # Colorama is Windows only package
+    colorama = None
 
+if PRINT_COLOR and colorama:
     colorama.init()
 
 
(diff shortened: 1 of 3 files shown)


itsayellow:
Does this have a measurable benefit?  (Startup time, etc.)

hwalinga:
Barely anything significant. But that is not the reason I took the effort to submit this PR. 

To give some background. I use poetry as my dependency manager and poetry will always uninstall colorama because `pylint` and `pytest` both list colorama as a Windows only dependency. Breaking pipx every time I use poetry. 

You can say this is a poetry bug, or that I just have to install pipx in its own environment, `pipx install pipx` is what I have now ; ). But listing colorama as a Windows only dependency is just more correct, just as `pylint` and `pytest` have done, and just makes sure `pipx` is a bit more robust on other platforms if there goes anything wrong with `colorama`.

itsayellow:
Ah ok, I didn't realize there was this side effect with poetry.  Your motivation makes more sense now.

It's too bad, because the whole point of colorama is that it works cross-platform and you don't have to conditionally install on Windows. 😕

src/pipx/colors.py
```diff
@@ -1,11 +1,13 @@
 import sys
 from typing import Callable
 
-import colorama  # type: ignore
+from pipx.constants import WINDOWS
```
itsayellow, line 4: Instead of getting the `WINDOWS` constant, I would just do the following:
```python
try:
    import colorama
except ImportError:
    colorama = None
```

Then just change `if PRINT_COLOR:` in the original code to `if PRINT_COLOR and colorama:`

That way we only have one platform check on whether or not colorama is installed in `setup.py`, and we always use colorama if it is installed.  And we don't have to worry about the possibility of our `WINDOWS` constant somehow being out of sync with `sys_platform="win32"`

itsayellow (changes requested):

hwalinga:
Yes, that is true, if other tools didn't specify colorama as Windows only and just ran it unconditionally Poetry would never decided to remove it. It's just that Python dependency management isn't as robust as it can be. Without that problem pipx might have never existed. 

Here is the issue from Poetry talking about that: 
https://github.com/python-poetry/poetry/issues/3139

src/pipx/colors.py
```diff
@@ -1,13 +1,14 @@
 import sys
 from typing import Callable
 
-from pipx.constants import WINDOWS
-
 PRINT_COLOR = sys.stdout.isatty()
```
itsayellow, line 4: Just a style nit: Could you place the `PRINT_COLOR` line after the `try`...`except` stanza (I like to put all the imports first as much as is possible.)  and I think that should do it.

itsayellow (approved):

jimxliu:
Hello! I came across the following error when I was trying to install `pipx 1.0.0` with `colorama 0.4.3` on **linux** (ubuntu). As stated in this PR, `>= 0.4.4` is only required in **windows**. 
```
  SolverProblemError

  Because pipx (1.0.0) depends on colorama (>=0.4.4)
   and my-app depends on colorama (0.4.3), pipx is forbidden.
  So, because my-app depends on pipx (1.0.0), version solving failed.
```
And the `pyproject.toml` looks like this 
```
[tool.poetry.dependencies]
colorama = "0.4.3"
python = "^3.7"
typer = "^=0.4.0"
pipx = "1.0.0"
virtualenv = "^20.8.1"
joblib = "^1.1.0"
```
system python 3.7

Any thoughts on why it didn't follow the linux requirement? Thank you!

PR #695, opened 2021-05-27
Replace pipxproject with pypa in docs and code

itsayellow:

## Summary of changes
I grep'ed for every instance of the text `pipxproject` in the docs and code, and replaced it with `pypa`.

## Test plan

Tested by running
```
```


uranusjr:
[The documentation nav bar still shows pipx-project/pipx](https://pypa.github.io/pipx/).

image

How can this be changed, a mkdocs rebuild?

itsayellow:
Thanks @uranusjr, yes docs need to be rebuilt.

I just manually triggered the Github workflow to rebuild the docs.  I think last night I assumed it would happen automatically, but in fact they are only rebuilt automatically for a release.

itsayellow:
Ah nope that didn't do it.  This text is `pipx-project`, and I grep'ed for `pipxproject` and didn't catch it.  Let me fix it.

itsayellow:
I think it's fixed now.  Thanks for the catch!

uranusjr:
Yup, thanks!

PR #706, opened 2021-06-14
Fix Python version inspection for pre-release

uranusjr:

## Summary of changes

pipx failed on my when I tried to run it against a pre-release Python:

```
$ ./python -c "import sys; print(sys.implementation)"
sys.version_info(major=3, minor=10, micro=0, releaselevel='beta', serial=2)
```

This fixes that.

## Test plan

Tested by running

1. Build a pre-release Python from source
2. `pipx install --python=path/to/prerelease/python cowsay`

### DIFF
--- a/src/pipx/venv_inspect.py
+++ b/src/pipx/venv_inspect.py
@@ -172,7 +172,11 @@ def fetch_info_in_venv(venv_python_path: Path) -> Tuple[List[str], Dict[str, str
         impl_ver = sys.implementation.version
         implementation_version = "{0.major}.{0.minor}.{0.micro}".format(impl_ver)
         if impl_ver.releaselevel != "final":
-            implementation_version += impl_ver.releaselevel[0] + impl_ver.serial
+            implementation_version = "{}{}{}".format(
+                implementation_version,
+                impl_ver.releaselevel[0],
+                impl_ver.serial,
+            )
 
         sys_path = sys.path
         try:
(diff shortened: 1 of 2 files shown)


src/pipx/venv_inspect.py
```diff
@@ -172,7 +172,11 @@ def fetch_info_in_venv(venv_python_path: Path) -> Tuple[List[str], Dict[str, str
         impl_ver = sys.implementation.version
         implementation_version = "{0.major}.{0.minor}.{0.micro}".format(impl_ver)
         if impl_ver.releaselevel != "final":
-            implementation_version += impl_ver.releaselevel[0] + impl_ver.serial
+            implementation_version = "{}{}{}".format(
```
itsayellow, line 175: Don't we need some periods `.` in this format string?

itsayellow:
Do newer python versions use integers in their version information?  I didn't know this, I thought the string type for versions was pretty python-universal.

uranusjr:
It seems like major, minor, micro, and serial components are all ints, only the release level part is str.

https://github.com/python/cpython/blob/689a84475e7b1da79d5ae82df67ab8897316f98c/Python/sysmodule.c#L2659-L2663

src/pipx/venv_inspect.py
uranusjr: This produces something like `3.10.0a1`, which is the preferred PEP 404 form. Dots can be added but will just be normalised away.

itsayellow: Oh I see.  You nested `implementation_version` instead of doing the `+=`.  OK, I understand now.

itsayellow (approved):
LGTM, thanks for keeping us up-to-date @uranusjr

PR #714, opened 2021-07-17
Use value of `#egg=` to determine package name if provided

nyuszika7h:

## Summary of changes
If `#egg=` is provided on the command line, it will be used to determine the package name, which is faster than trying to resolve it.

## Test plan

Tested by running
```
pipx install git+https://github.com/nyuszika7h/poetry.git@vcs-disable-hashes\#egg=poetry
```

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -2,6 +2,7 @@ dev
 
 - Fix to `pipx ensurepath` to fix behavior in user locales other than UTF-8, to fix #644. The internal change is to use userpath v1.6.0 or greater. (#700)
 - Fix virtual environment inspection for Python releases that uses an int for its release serial number. (#706)
+- Use package name from `#egg=` if present.
 
 0.16.3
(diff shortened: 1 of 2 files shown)


uranusjr:
-1 to this. The `egg=name` syntax is a non-standard setuptools and pip extension left from a now distant past, and is set to be deprecated for eventual removal in the future. New tools are expected to only support the PEP 508 `name @ url` syntax, which pipx does correctly.

A warning message telling the user the syntax is useless for pipx and and should to *not* be used would be a reasonable addition.

nyuszika7h:
I wasn't aware of that. However, `pipx install 'poetry @ git+https://github.com/nyuszika7h/poetry.git'` still seems to spend quite a long time on "determining package name".

uranusjr:
That I think we should improve (using basically your approach but parsing the requirement with `packaging`).

PR #716, opened 2021-07-19
improving installation instructions for windows.

sander76:
In my experience the windows instructions are minimal and not always correct (due to the nature of windows mostly I guess).
For this reason I am assuming installation will fail many times due to this, discouraging new users to actually use this nice tool.

This pull request should -hopefully- improve on this. I have now only updated the readme file and understand there is also some  things to change in the documentation, but I'd like to see where this goes first before I put energy in this.

Sander.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ WARNING: The script pipx.exe is installed in `<USER folder>\AppData\Roaming\Pyth
 ```
 
 If so, go to the mentioned folder, allowing you to run the pipx executable directly.
-Enter the following line (even if you did not get the warning.):
+Enter the following line (even if you did not get the warning):
 
 ```
 pipx ensurepath
@@ -67,7 +67,7 @@ Restart your terminal session and verify `pipx` does run.
 
 Upgrade pipx with `python3 -m pip install --user -U pipx`.
 
-### Shell completions.
+### Shell completions
 
 Shell completions are available by following the instructions printed with this command:
 ```
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ _For comparison to other tools including pipsi, see [Comparison to Other Tools](
 
 ## Install pipx
 
-On macOS:
+### On macOS
 
 ```
 brew install pipx
@@ -34,14 +34,40 @@ pipx ensurepath
 
 Upgrade pipx with `brew update && brew upgrade pipx`.
 
-Otherwise, install via pip (requires pip 19.0 or later):
+### On Linux, install via pip (requires pip 19.0 or later)
 
 ```
 python3 -m pip install --user pipx
 python3 -m pipx ensurepath
 ```
+Upgrade pipx with `python3 -m pip install --user -U pipx`.
+
+### On Windows, install via pip (requires pip 19.0 or later)
+
+```
+# If you installed python using the app-store, replace `python` with `python3` in the next line.
+python -m pip install --user pipx
+```
+
+It is possible (even most likely) the above finishes with a WARNING looking similar to this:
+
+```
+WARNING: The script pipx.exe is installed in `<USER folder>\AppData\Roaming\Python\Python3x\Scripts` which is not on PATH
+```
+
+If so, go to the mentioned folder, allowing you to run the pipx executable directly.
+Enter the following line (even if you did not get the warning.):
+
+```
+pipx ensurepath
+```
+
+This will add both the above mentioned path and the `%USERPROFILE%.local\bin` folder to your search path.
+Restart your terminal session and verify `pipx` does run.
 
 Upgrade pipx with `python3 -m pip install --user -U pipx`.
+
+### Shell completions.
 
 Shell completions are available by following the instructions printed with this command:
 ```


README.md
```diff
@@ -34,15 +34,41 @@ pipx ensurepath
...
 Upgrade pipx with `brew update && brew upgrade pipx`.
 
-Otherwise, install via pip (requires pip 19.0 or later):
+### On Linux, install via pip (requires pip 19.0 or later)
 
 ```
 python3 -m pip install --user pipx
 python3 -m pipx ensurepath
 ```
+Upgrade pipx with `python3 -m pip install --user -U pipx`.
```
gaborbernat, line 43: Perhaps we should replace --user with https://pypi.org/project/pipx-in-pipx/ 🤔

sander76: My mind is now in an infinite loop

sander76: Instead of adding documentation to this I thought it would even be better to have the installation ask for and actually adding the path variable instead of issuing a warning. Don't know if that's doable.

uranusjr: Or https://github.com/uranusjr/pipx-standalone 🙂

sander76: > 
> 
> Or we could incorporate https://github.com/uranusjr/pipx-standalone into the main repo and the release process 🙂

Scoop looks nice! But I would consider this an extra. Not as the default solution tbh.

gaborbernat: Yeah, it's also a Windows-only thing :thinking: while pipx-in-pipx is cross-platform compatible solution.

uranusjr: FWIW we have Homebrew five lines into the documentation, and it is very much not universal.

sander76: But can't we -in the meantime- just improve on the installation instruction of what we have now ? 
I haven't looked into pipx-in-pipx but it is -like scoop I guess- yet another dependency.

gaborbernat: It's actually not. It's just a wrapper to install pipx with pipx :blush: It just requires pip and python which all platforms should already have.

sander76: Ok. So what do you suggest? Do you prefer this to become the default way of installing pipx? 
In any case, this is out of my knowledge scope to have a proper opinion about.

My only fear is -right or wrong- we're solving one thing and potentially breaking a lot of other. Assuming we only have issues installing on windows..

gaborbernat: I personally would prefer it to become the standard :thinking:

sander76: Knowing your merits inside the python community I have full trust adopting your personal preference and will happily set aside my fears. 

But this is definitely not my choice to make. So close this PR then ?
I will start playing around with pipx-in-pipx.

sander76: Or should pipx fully absorb pipx-in-pipx ?

gaborbernat: I don't think pipx can absorh pipx-in-pipx, at best we'd look at pipx maintainers adopting it and moving it under PyPA. I don't think you need to abandon this PR, the changes are still great, all I'm saying is that perhaps we could switch from using ``--user`` to use ``pipx-in-pipx`` 🤔

gaborbernat (approved):

README.md
```diff
@@ -34,15 +34,41 @@ pipx ensurepath
...
+
+### On Windows, install via pip (requires pip 19.0 or later)
+
+```
+# If you installed python using the app-store, replace `python` with `python3` in the next line.
```
imba-tjd, line 48: Hi. I'm using the store version. I don't get the reason why `python3` matters here.

sander76: > Hi. I'm using the store version. I don't get the reason why `python3` matters here.

I guess at the time it was the thing to do for windows, still making a distinction between 2.7 and python3. I wonder if using `python3` still works or whether we need to change the documentation ? (I am not running windows anymore so unable to check)

imba-tjd: python and python3 is the same thing on win.

sander76: Alright. So I guess we should change this

imba-tjd: Besides, if you say *pipx ensurepath will add both the above mentioned path and the `%USERPROFILE%\.local\bin` folder to your search path*, why `python3 -m pipx ensurepath` wouldn't?

Honestly I prefer the previous guide. I don't see a difference between Linux and Windows installation. The `Otherwise` is enough.

Though there are indeed some differences between store and nornal version. For example you don't need `--user` when using store version.

sander76: The previous guide did not work. I can confirm this as I was on windows at that time. It could be that there is a difference in using python from the store or an install downloaded from pypi.org (which I did and also might have changed since this doc was written). In the end, the goal is getting your search paths right. So maybe there is not one solution that works in all windows situations.

imba-tjd: Then I bet that you met https://github.com/ofek/userpath/issues/25

If so, I would suggest reverting this PR.

sander76: Basically the new instruction tells you what to do when something goes wrong. Which, in my situation (and other's) was the case (and many times over and over again)
I don't see removing that does help improve the instruction. 

If you can confirm the old instructions worked for a store based installation, maybe that could be described as the happy/default flow. And then add an "if things go wrong" section describing a possible solution.

But as I am no longer on windows anymore I am no longer in the right position to do this and make the change. So maybe add another pr making the change ?

imba-tjd: I can confirm at my that time it didn't work for store version.

The point is that I don't believe *go to the mentioned folder, run the pipx executable directly* makes difference with `python -m pipx ensurepath`.
The reaon you succeed is because you said you changed store version to normal one. You can see my test at `userpath#25`.

I'm still using store version. However, I just tested and find `python -m pipx ensurepath` now works, so do `userpath append`. So I can't check whether *run pipx directly* makes difference, either.

PR #733, opened 2021-09-09
Add better handling for 'app not found'

### LINKED ISSUE #725: Run console scripts whose name doesn't match package name
For eg:

```console
$ pipx run textflip            
'textflip' executable script not found in package 'textflip'. Available
executable scripts: flip
```

So it does know about the script named `flip`, but I can't seem to figure out how to make it run that.

I've tried doing:

- `pipx run textflip.flip`
- `pipx run textflip --spec flip`
- `pipx run textflip flip` etc.

None of these seem to work, and I can't seem to find the relevant documentation.

tusharsadhwani:

## Summary of changes

Added one feature:
- If the package is not a runnable app, and there is only one runnable app in the package, run it with a note.

Improved the error message thrown when the package is not a runnable app, to include the command required to run the app.

Resolves #725


src/pipx/commands/run.py
```diff
@@ -146,12 +151,29 @@ def _download_and_run(
...
+
+        # If there's a single app inside the package, run that by default
+        if len(apps) == 1:
+            app = apps[0]
+            print(f"NOTE: running app {app!r} from {package_name!r}")
+            if WINDOWS:
+                app_filename = f"{app}.exe"
+                logger.info(f"Assuming app is {app_filename!r} (Windows only)")
+            else:
+                app_filename = app
```
uranusjr, line 163: This probably should not use `print`. Also, the Windows only line can just be combined to the previous log (something like *running app 'foo' (assuming 'foo.exe' on Windows) from 'foo-package'*).

gaborbernat (changes requested):
Needs a test plus changelog, plus what the other reviewer requested.

tusharsadhwani:
@gaborbernat I wasn't able to get most of the tests to pass on my machine even on master branch, so I can't reliably add a test here. Plus, I'm not entirely how to test this fix as well.

tusharsadhwani:
I can, however, provide a pretty good example of what was broken and is now fixed:

```console
$ pipx run textflip hello            
'textflip' executable script not found in package 'textflip'. Available
executable scripts: flip

$  . venv/bin/activate

$ pipx run textflip hello
oןןǝɥ
```

As for a package that has multiple runnable apps but neither of them matching the name of the package, I could find no examples sadly.

src/pipx/commands/run.py
tusharsadhwani: A slight concern about this: Shouldn't the user be notified that an app whose name is different from what he just typed is being run?
Changing the `print` to `logging.info` will stop it from showing up to the user. Is that intended?

uranusjr: Sorry, I was confused between different logger setups between pip and pipx (by default `INFO` is shown in pip, but `pipx` only shows `WARNING`). Using `print` is correct here.

gaborbernat (changes requested):
Fix merge conflicts and run tests again.

docs/changelog.md
```diff
@@ -1,14 +1,15 @@
 dev
 
 - Include machinery to build a manpage for pipx with [argparse-manpage](https://pypi.org/project/argparse-manpage/).
+- Fixed `pipx list` output phrasing to convey that python version displayed is the one with which package was installed.
```
tusharsadhwani, line 4:
```suggestion
```

docs/changelog.md
```diff
@@ -1,14 +1,15 @@
...
+- Add better handling for 'app not found' when a single app is present in the project, and an improved error message (#733)
 
 0.17.0
 
 - Support `pipx run` with version constraints and extras. (#697)
 
 0.16.5
 
-- Fixed `pipx list` output phrasing to convey that python version displayed is the one with which package was installed.
```
tusharsadhwani, line 11:
```suggestion
- Fixed `pipx list` output phrasing to convey that python version displayed is the one with which package was installed.
```

gaborbernat (approved):

Issue #125, opened 2019-03-26
pipx install not quite idempotent


chrish42:
Hi. I'm writing scripts to install multiple packages via pipx (thank you for writing it!). I want these scripts to be idempotent (i.e. running it a second time has no impact after running it once, and it still succeeds). I also want to know when there are errors during installation. However, the current way `pipx install` is implemented makes those two conflicting goals.

Running `pipx install foo` a second time does nothing more than running it just once (except print a message the second time saying "Not installing to existing directory ..."), which is good. However, the exit code for that second time is set to a non-zero value. So either I ignore the exit code, and then `pipx install foo` is idempotent, but I don't know if there was an error. Or I look at the exit code, and then I'll know if the install failed, but then running it a second time is not the same as once (because it fails with a non-zero exit code the second time).

cs01:
> However, the exit code for that second time is set to a non-zero value.

I'll change it to 0 if something is already installed. This is what `pip` does, and `pipx` should have done this from the start.

cs01:
this has been added to v0.13

wombelix:
> this has been added to v0.13

I'm using v0.16.4 and it looks like that this changes was reverted by PR https://github.com/pypa/pipx/pull/560
When I run `pipx install` for an already installed package, the return code is again `1` instead `0`.

`EXIT_CODE_INSTALL_VENV_EXISTS` in [pipx/constants.py](https://github.com/pypa/pipx/blob/72786fb873556f195e4ed2320327aa810e59a637/src/pipx/constants.py#L24) is set to `1` and used in [commands/install.py](https://github.com/pypa/pipx/blob/72786fb873556f195e4ed2320327aa810e59a637/src/pipx/commands/install.py#L56)

Example:
```
wombelix@geeko:~> pipx install poetry
'poetry' already seems to be installed. Not modifying existing installation in '/home/wombelix/.local/pipx/venvs/poetry'. Pass '--force' to force installation.

wombelix@geeko:~> echo $?
1
```

Expected was RC 0, similar to `pip`:
```
wombelix@geeko:~> pip install pipx
Defaulting to user installation because normal site-packages is not writeable
Requirement already satisfied: pipx in ./.local/lib/python3.8/site-packages (0.16.4)
Requirement already satisfied: packaging>=20.0 in /usr/lib/python3.8/site-packages (from pipx) (20.9)
[...]

PR #736, opened 2021-09-22
fix(install): Return Code 0 if venv already exists

wombelix:

## Summary of changes
Related Issue: https://github.com/pypa/pipx/issues/125

Restores functionality introduced in Commit: https://github.com/pypa/pipx/commit/11b853e9c6926b32133b27822516b2a5b4f35411

## Test plan

Tested by running and checking the return code
```
pipx install <alreadyinstalledpackage>
echo $?
# expected result: 0
```


gaborbernat (changes requested):
You'll need to validate with test, add changelog.

wombelix:
> You'll need to validate with test, add changelog.

Thanks @gaborbernat and apologize for not taking enough attention to the contributing guide earlier. I pushed a new commit that address the failed test, local run was now successful, also added an changelog entry.

docs/changelog.md
```diff
@@ -1,6 +1,7 @@
 dev
 
 - Fixed `pipx list` output phrasing to convey that python version displayed is the one with which package was installed. 
+- Fixed `pipx install` to provide return code 0 if venv already exists, similar behavior as pip. (#736)
```
uranusjr, line 4:
```suggestion
- Fixed `pipx install` to provide return code 0 if venv already exists, similarly to pip’s behavior. (#736)
```

wombelix: Thanks for the suggestion, updated and new commit pushed accordingly.

uranusjr (approved):

gaborbernat (approved):

PR #744, opened 2021-09-30
Build manpage with argparse-manpage

stefanor:

## Summary of changes

Add a script to generate a manpage with argparse-manpage.

In Debian we strive to have manpages for all command line utilities.

An explicitly written manpage usually gives the best result, but falls out of date quickly, unless maintained, and editing nroff source directly isn't fun. Sphinx has tooling for generating manpages from rst, but mkdocs doesn't seem to have any for manpages from markdown, yet. There are third-party tools that do this (ronn in the ruby world, and go-md2man in the Go world).

But argparse-manpage automatically generates a reasonable looking manpage, with minimal effort.

## Test plan
Tested by running
```
$ make man
$ man -l pipx.1
```

### DIFF
--- a/scripts/generate_man.py
+++ b/scripts/generate_man.py
@@ -1,0 +1,42 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+import os.path
+import textwrap
+
+from build_manpages.manpage import Manpage
+
+from pipx.main import get_command_parser
+
+
+def main():
+    parser = get_command_parser()
+    parser.man_short_description = parser.description.splitlines()[1]
+
+    manpage = Manpage(parser)
+    body = str(manpage)
+
+    # Avoid hardcoding build paths in manpages (and improve readability)
+    body = body.replace(os.path.expanduser("~"), "~")
+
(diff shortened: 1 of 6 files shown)


uranusjr:
Oops, conflicts.

scripts/generate_man.py
```diff
@@ -0,0 +1,52 @@
...
+        return "pipx"
+
+    def get_author(self):
+        return "Chad Smith and contributors"
+
+    def get_author_email(self):
+        return "grassfedcode@gmail.com"
+
+    def get_url(self):
+        return "https://pypa.github.io/pipx/"
```
stefanor, line 24: Instead of hardcoding these, I also played with extracting them from importlib_metadata, but that can only be done when it's installed, so this simpler approach seemed the right way to go.

stefanor: Actually, there is a lower level API that probably makes more sense to use, brb...

gaborbernat (changes requested):
You'll need to fix the CI :thinking:

gaborbernat (approved):

PR #745, opened 2021-09-30
Standalone generate docs

stefanor:
## Summary of changes

Allow `scripts/generate_docs.py` to be run directly on the source tree (with `PYTHONPATH=src`, for example).
This makes it easier to run during a Debian package build.

While we're here, update the makefile, the `docs` target wasn't updated after a noxfile reorg.

## Test plan
Tested by running
```
$ PYTHONPATH=src python3 scripts/generate_docs.py
$ make docs
```

### DIFF
--- a/makefile
+++ b/makefile
@@ -13,7 +13,7 @@ publish:
 	pipx run nox -s publish
 
 docs:
-	pipx run nox -s docs -r
+	pipx run nox -s build_docs -r
 
 watch_docs:
 	pipx run nox -s watch_docs -r
--- a/scripts/generate_docs.py
+++ b/scripts/generate_docs.py
@@ -3,6 +3,7 @@
 
 import os
 import subprocess
+import sys
 import textwrap
 from typing import Optional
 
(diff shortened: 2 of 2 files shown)


makefile
```diff
@@ -13,7 +13,7 @@ publish:
 	pipx run nox -s publish
 
 docs:
-	pipx run nox -s docs -r
+	pipx run nox -s build_docs -r
```
uranusjr, line 16: Hmm does this indicate nobody is actually using the makefile…?

uranusjr (approved):

makefile
cs01: I think everything in it was migrated to nox. If it wasn’t, it should be, and the makefile should be deleted.

gaborbernat (changes requested):
The CI is failing here.

cs01:
The makefile is now deprecated. The noxfile is the only way to guarantee (or at least come close to guaranteeing) a reproducible build with the same dependencies and environment (the pipx binary is on the user's path because `DOC_DEPENDENCIES = [".", "jinja2", "mkdocs", "mkdocs-material"]` contains `.` which installs pipx into the virtual environment nox runs the task in).

PR #751, opened 2021-10-12
Bump linters and release 0.16.5

gaborbernat:
Signed-off-by: Bernát Gábor <gaborjbernat@gmail.com>


src/pipx/version.py
```diff
@@ -1,2 +1,2 @@
-__version_info__ = (0, 16, 5, 1, "dev0")
+__version_info__ = (0, 16, 5)
```
uranusjr, line 1: Maybe we should skip 0.16.5 entirely and go directly to 0.16.5.1 on PyPI? This way we don’t need to force re-tag the Git commit, which might cause unwanted problems.

gaborbernat: That would not solve #773, I'm not expecting any unwanted problems, do you have any in mind?

uranusjr: If the goal is to solve #773, 0.16.5 should be released directly from that tag without any modification.

gaborbernat: Not necessarily and that would be an incorrect release because the tag was applied on a bad version number in the source code. IMHO the right way ahead is to move the tag.

PR #753, opened 2021-10-24
Update troubleshooting.md

pswalia2u:
Added ingore ssl/tls errors example.

## Test plan

Tested by running
```
pipx install termpair --pip-args '--trusted-host files.pythonhosted.org --trusted-host pypi.org --trusted-host pypi.python.org --trusted-host github.com'"
```


imba-tjd:
Hi. Would you mind elaborating this change? I don't think this should be covered in pipx.

PR #756, opened 2021-10-28
Fix python 3.6-3.8 compatibility error

alextremblay:

## Summary of changes
All PR CI checks are currently failing because a method call was added to the `pipx.commands.reinstall.reinstall` function which was only added in python 3.9

Replaced that method call with a python >=3.6 workaround

## Test plan

Tested by running
```
```

### DIFF
--- a/src/pipx/commands/reinstall.py
+++ b/src/pipx/commands/reinstall.py
@@ -27,7 +27,15 @@ def reinstall(
         print(f"Nothing to reinstall for {venv_dir.name} {sleep}")
         return EXIT_CODE_REINSTALL_VENV_NONEXISTENT
 
-    if Path(python).is_relative_to(venv_dir):
+    try:
+        # use PurePath.relative_to in a try block instead
+        # of PurePath.is_relative_to, for python 3.6-3.8 compatability
+        Path(python).relative_to(venv_dir)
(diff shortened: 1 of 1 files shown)


alextremblay:
@uranusjr Hello, would you be able to re-review this for me?

uranusjr (approved):

alextremblay:
What's the process for getting changes merged?

uranusjr:
First you’ll probably need to provide a changelog entry for this fix!

alextremblay:
hmmm.... but the bug i fixed doesn't exist in any published versions of pipx... does it still make sense to add a changelog entry?

If the upcoming version of pipx simultaniously introduced a bug and fixed that bug before release, should it still be added to the changelog?

cs01:
> hmmm.... but the bug i fixed doesn't exist in any published versions of pipx... does it still make sense to add a changelog entry?
> 
> If the upcoming version of pipx simultaniously introduced a bug and fixed that bug before release, should it still be added to the changelog?

No definitely not. Apologies for the slow review. Thank you for taking the time to make the fix, and taking the time to follow up.

uranusjr:
Oh! In that case I think this is ready. Thanks for the clarification!

PR #763, opened 2021-11-16
Replace `master` with `main`

NdagiStanley:

## Summary of changes

Updated the fllowing files to refer to `main` branch instead of `master` branch:

- https://github.com/pypa/pipx/blob/main/.github/workflows/publish_docs.yml
- https://github.com/pypa/pipx/blob/main/docs/comparisons.md#migrating-to-pipx-from-pipsi
- https://github.com/pypa/pipx/blob/main/noxfile.py#L86

## Test plan

Tested by running the following in a python virtual environment:
```
pip install nox
nox -s watch_docs
```

Any `nox` command that's dependent on the `noxfile.py` (which had the most edits in this PR) runs successfully.

### DIFF
--- a/.github/workflows/publish_docs.yml
+++ b/.github/workflows/publish_docs.yml
@@ -12,7 +12,7 @@ jobs:
     name: Deploy docs
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout master
+      - name: Checkout main
         uses: actions/checkout@v2
       - name: Set up Python ${{ env.default-python }}
         uses: actions/setup-python@v2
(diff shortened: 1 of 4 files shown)


uranusjr (approved):

NdagiStanley:
@uranusjr Should I fix the failing tests in another PR. I doubt I am responsible for making these fail.

uranusjr:
IIRC the failure is actually caused by a GitHub Action bug :(

uranusjr:
Merging since all changes here are purely cosmetic.

PR #785, opened 2022-01-01
Make the docs reproducible

stefanor:

## Summary of changes

Forwarded from: https://bugs.debian.org/995865

Whilst working on the Reproducible Builds effort [0] I noticed that python-pipx could not be built reproducibly.

This is because the generated manpages uses values from the build system, specifically the $HOME directory.

A patch is attached that replaces this with a literal "$HOME" string (which is arguably more informative...).

[0] https://reproducible-builds.org/

### DIFF
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -57,7 +57,7 @@ PIPX_DESCRIPTION = textwrap.dedent(
     Virtual Environment location is {str(constants.PIPX_LOCAL_VENVS)}.
     Symlinks to apps are placed in {str(constants.LOCAL_BIN_DIR)}.
 
-    """
+    """.replace(str(Path.home()), "$HOME")
 )
 PIPX_DESCRIPTION += pipx_wrap(
     """
@@ -104,7 +104,7 @@ INSTALL_DESCRIPTION = textwrap.dedent(
     The default python executable used to install a package is
     {DOC_DEFAULT_PYTHON} and can be overridden
     by setting the environment variable `PIPX_DEFAULT_PYTHON`.
-    """
+    """.replace(str(Path.home()), "$HOME")
 )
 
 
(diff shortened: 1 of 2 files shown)


src/pipx/main.py
```diff
@@ -57,7 +57,7 @@ def print_version() -> None:
     Virtual Environment location is {str(constants.PIPX_LOCAL_VENVS)}.
     Symlinks to apps are placed in {str(constants.LOCAL_BIN_DIR)}.
 
-    """
+    """.replace(str(Path.home()), "$HOME")
```
gaborbernat, line 60: Would this make sense on Windows too? I'm not a big fan of this change; I prefer the help message to show the actual paths, not some placeholder values.

gaborbernat (changes requested):

src/pipx/main.py
stefanor: It will *work* on Windows, but `$HOME` isn't expected to be set on Windows.

gaborbernat: That's what I mean. This change only makes sense on Debian. Overall I'm -1 on this change proposal. It's much better to show the path in the help message than add some placeholder value.

stefanor: Alternatively we could do the replacement in #744, the main motivation here was to make the manpages reproduceable.

There is a similar mechanism in generate_docs, already: https://github.com/pypa/pipx/blob/main/scripts/generate_docs.py#L23

gaborbernat:
Yeah let's not change it here just in 744.

gaborbernat:
The man page can be global, but the help message should be personalized.

stefanor:
Actually, I had already tried to do this in #744, so I'm confused about why this was necessary...

stefanor:
Resolved with c472921.

Issue #713, opened 2021-07-09
Add `pipx hook` command


paulmelnikow:
Hi! Thank you for this _fabulous_ tool!

**How would this feature be useful?**

Conda offers a command `conda shell.{bash|zsh|fish} hook` which can be invoked from scripts to bring the Conda environment into the current shell.

The boilerplate Conda adds to rcfiles actually invokes this command! But it's [also useful for scripts](https://erictleung.com/conda-in-subshell-script) which install a sequence of tools (Conda, pipx, Poetry) to bootstrap a developer's Python environment, with minimal interaction on their part.

I'm running `pipx ensurepath` from my bootstrap script, however it doesn't take effect until starting the next shell. Starting a new shell is a rather clunky thing to do in the middle of a shell script execution.

**Describe the solution you'd like**

A `pipx hook {bash|zsh|fish}` command which outputs `set -x PATH $PATH /Users/pnm/.local/bin` or something similar, which updates and exports `PATH` in the current shell.

Another alternative (a little less nice I guess) would be a command like `pipx bindir` which resolves pipx's config and just prints `/Users/pnm/.local/bin`.

**Describe alternatives you've considered**

My workaround is to hard-code ~/.local/bin in my bootstrap script. This works, however I'd have more confidence using a pipx command since I know it will honor pipx's config. It also makes the intention of my bootstrap script more self-documenting.

uranusjr:
I wonder if it’s better for pipx to only output the last part (`/Users/pnm/.local/bin` in your example) and let the surrounding script format the appropriate command. This way pipx doesn’t need to add any shell-specific stuff (which I’ve always felt is an anti-pattern).

paulmelnikow:
> I wonder if it’s better for pipx to only output the last part (`/Users/pnm/.local/bin` in your example) and let the surrounding script format the appropriate command. This way pipx doesn’t need to add any shell-specific stuff (which I’ve always felt is an anti-pattern).

Yep, something like `pipx bindir` or `pipx bindirs` would be just fine for my application. Since I'm running this from a shell I know the syntax I need.

[...]

PR #793, opened 2022-01-13
Add `pipx environment` command

meowmeowmeowcat:

## Summary of changes

Add `pipx environment` command.
See #713 

Example output:
```
$ pipx environment
PIPX_HOME=/home/user/.local/pipx
PIPX_BIN_DIR=/home/user/.local/bin
PIPX_SHARED_LIBS=/home/user/.local/pipx/shared
PIPX_LOCAL_VENVS=/home/user/.local/pipx/venvs
PIPX_LOG_DIR=/home/user/.local/pipx/logs
PIPX_TRASH_DIR=/home/user/.local/pipx/.trash
PIPX_VENV_CACHEDIR=/home/user/.local/pipx/.cache
```

```
$ pipx environment --value PIPX_HOME
/home/user/.local/pipx
```

## Test plan

Tested by running
```
pipx install .
pipx environment
pipx environment --value PIPX_HOME
pipx environment -v PIPX_LOCAL_VENVS
pipx environment --value=PIPX_BIN_DIR
```


src/pipx/commands/environment.py
```diff
@@ -0,0 +1,44 @@
...
+    elif "PIPX_LOCAL_VENVS" in value:
+        print(PIPX_LOCAL_VENVS)
+    elif "PIPX_LOG_DIR" in value:
+        print(PIPX_LOG_DIR)
+    elif "PIPX_TRASH_DIR" in value:
+        print(PIPX_TRASH_DIR)
+    elif "PIPX_VENV_CACHEDIR" in value:
+        print(PIPX_VENV_CACHEDIR)
+    else:
+        print("Variable not found.", file=sys.stderr)
```
cs01, line 41: This should just throw a `PipxError` with the text you want to displayed, and pipx will take care of the error message being printed to the right stream to the user. (see the bottom of `main.py` for `    except PipxError as e:`).

cs01, line 41: The logic in this block can be simplified by making a dict, then checking if `value` is in it, and if so, printing it

src/pipx/main.py
```diff
@@ -605,6 +607,21 @@ def _add_ensurepath(subparsers: argparse._SubParsersAction) -> None:
...
+            "Available variables: "
+            "PIPX_HOME, PIPX_BIN_DIR, PIPX_SHARED_LIBS, PIPX_LOCAL_VENVS, PIPX_LOG_DIR, "
+            "PIPX_TRASH_DIR, PIPX_VENV_CACHEDIR"
+        ),
+    )
+    p.add_argument(
+        "--value", "-v", metavar="VARIABLE", help="Print the value of the variable."
+    )
+
+
```
cs01, line 624: I think this is useful, but I am concerned there will be some confusion about whether these are settable by users or not. 

Can you add a note to the end of the actual output (in environment.py) mentioning only some of those are settable, and to see the list, see the output of `pipx --help`.

tests/test_environment.py
```diff
@@ -0,0 +1,27 @@
...
+def test_cli_with_args(monkeypatch, capsys):
+    assert not run_pipx_cli(["environment", "--value", "PIPX_HOME"])
+    assert not run_pipx_cli(["environment", "--value", "PIPX_BIN_DIR"])
+    assert not run_pipx_cli(["environment", "--value", "PIPX_SHARED_LIBS"])
+    assert not run_pipx_cli(["environment", "--value", "PIPX_LOCAL_VENVS"])
+    assert not run_pipx_cli(["environment", "--value", "PIPX_LOG_DIR"])
+    assert not run_pipx_cli(["environment", "--value", "PIPX_TRASH_DIR"])
+    assert not run_pipx_cli(["environment", "--value", "PIPX_VENV_CACHEDIR"])
+
+    run_pipx_cli(["environment", "--value", "SSS"])
```
cs01, line 25: Nice!

cs01 (changes requested):
Looking good! Thanks for doing this. Just needs some slight updates.

src/pipx/main.py
meowmeowmeowcat: I think only `PIPX_HOME` and `PIPX_BIN_DIR` are settable currently in the above list. Probably the note can display that they are settable directly, instead of telling users to run a new command?

cs01: > I think only `PIPX_HOME` and `PIPX_BIN_DIR` are settable currently in the above list. Probably the note can display that they are settable directly, instead of telling users to run a new command?

Sure that works

cs01 (approved):

Issue #792, opened 2022-01-04
`pipx list` should have an option to list packages only


njthomas:
**How would this feature be useful?**
Currently, `pipx list` outputs a long/verbose format by default:
```
$ pipx list
venvs are in /usr/home/njt/code/3p/pipx/venvs
apps are exposed on your $PATH at /usr/home/njt/code/3p/pipx/bin
   package pipx 0.17.0, installed using Python 3.8.12
    - pipx
   package virtualenv 20.12.1, installed using Python 3.8.12
    - virtualenv
```

This is hard to parse visually (and programatically, but one could argue that `--json` exists for that purpose, but more on that below).

There is a `--verbose` option available, but it adds information to an already verbose output format.

**Describe the solution you'd like**
I would like pipx to support a `--short` option to list the installed packages only:
```
$ pipx list --short
pipx
virtualenv
```

**Describe alternatives you've considered**

Ideally, `pipx list` by itself would have given the short listing by default, and `--verbose` would have given the extra information, but since this is already set, and users are used to the listing as it is, adding `--short` to do this would suffice.

**NB1:** Since pipx has not yet reached v1.0, I don't think it would be too big a deal to have `list` print the short output by default, and move the current default to `--verbose`, but I will leave that up to the maintainers to decide if they want to do that. I don't think it would break anything (users should be using `--json` to parse output anyway, and according to [Semantic Versioning rules](https://semver.org/#spec-item-4), major version zero allows for breaking changes.

**NB2:** I am aware that using `--json` and parsing the output will give the same effect (ie. `pipx list --json|jq -r '.venvs|keys[]'`), however, I submit to you that querying for the installed packages only **is a very common operation** and users should not have to resort to parsing JSON output to get this information.

**NB3:** I am okay with calling the option `--packages-only` or anything else, I just used `--short` as an example.

gaborbernat:
> **NB1:** Since pipx has not yet reached v1.0,

You missed that train by like 8 hours, https://twitter.com/gjbernat/status/1478297177604698115

> I would like pipx to support a `--short` option to list the installed packages only:

I'd be open to this, feel free to put in a PR with tests for this. I'd perhaps suggest to use PEP-508 formatting to display the version number in there, but that's just an option.

vrslev:
Maybe use flag “-q” instead?

cs01:
Closed by https://github.com/pypa/pipx/pull/804

PR #804, opened 2022-02-01
Add list --short option

q0w:

## Summary of changes
#792
Should `--short` flag have precedence over `--json` flag?
Should `--include-injected` flag be accepted with `--short`?
## Test plan

Tested by running
```
> pipx list --short
black 18.9b0
pipx 0.10.0
```

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,7 +1,7 @@
 dev
 
 - [docs] Fix the command for [installing development version](https://pypa.github.io/pipx/installation/#install-pipx-development-versions). (#801)
-- Add `list --short` option. (#804)
+- Add `list --short` option to list only package names. (#804)
 
 1.0.0
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,6 +1,7 @@
 dev
 
 - [docs] Fix the command for [installing development version](https://pypa.github.io/pipx/installation/#install-pipx-development-versions). (#801)
+- Add `list --short` option. (#804)
 
 1.0.0
 
--- a/tests/test_list.py
+++ b/tests/test_list.py
@@ -132,3 +132,15 @@ def test_list_json(pipx_temp_env, capsys):
         ),
         isort_package_ref,
     )
+
+
+def test_list_short(pipx_temp_env, monkeypatch, capsys):
+    assert not run_pipx_cli(["install", PKG["pycowsay"]["spec"]])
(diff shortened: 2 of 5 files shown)


cs01:
Cool, looks pretty good! Could you add example output to the summary?

> Should --short flag have precedence over --json flag?

These should be mutually exclusive. You should be able to do this with https://docs.python.org/3/library/argparse.html#mutual-exclusion.

> Should --include-injected flag be accepted with --short?

Probably not, since it wouldn't be very short in that case. Thoughts @njthomas?

cs01 (changes requested):
Cool, looks pretty good! 

Could you add example output to the summary?

> Should --short flag have precedence over --json flag?

These should be mutually exclusive. You should be able to do this with https://docs.python.org/3/library/argparse.html#mutual-exclusion.

> Should --include-injected flag be accepted with --short?

Probably not, since it wouldn't be very short in that case. Thoughts @njthomas?

docs/changelog.md
```diff
@@ -1,6 +1,7 @@
 dev
 
 - [docs] Fix the command for [installing development version](https://pypa.github.io/pipx/installation/#install-pipx-development-versions). (#801)
+- Add `list --short` option. (#804)
```
cs01, line 4: Could you briefly mention what this does? i.e. `Add `list --short` option, which only lists package names in list output

q0w: Done.

q0w:
> Could you add example output to the summary?
> These should be mutually exclusive. You should be able to do this with https://docs.python.org/3/library/argparse.html#mutual-exclusion.

Done.

tests/test_list.py
```diff
@@ -132,3 +132,15 @@ def test_list_json(pipx_temp_env, capsys):
...
+def test_list_short(pipx_temp_env, monkeypatch, capsys):
+    assert not run_pipx_cli(["install", PKG["pycowsay"]["spec"]])
+    assert not run_pipx_cli(["install", PKG["pylint"]["spec"]])
+    captured = capsys.readouterr()
+
+    assert not run_pipx_cli(["list", "--short"])
+    captured = capsys.readouterr()
+
+    assert "pycowsay 0.0.0.1" in captured.out
+    assert "pylint 2.3.1" in captured.out
```
cs01, line 146: nice!

cs01 (approved):
Look good, thank you!

q0w:
@cs01 but why did tests fail on macos?

PR #817, opened 2022-02-17
Update package metadata

ofek:

## Summary of changes

This implements [PEP 621](https://www.python.org/dev/peps/pep-0621/), obviating the need for `setup.py`, `setup.cfg`, and `MANIFEST.in`.

Support has not landed in `setuptools`, so builds will now use `hatchling`.

## Backend

Hatchling is the minimal build backend of [Hatch](https://github.com/ofek/hatch) v1.0.0 (v1 rewrite of CLI will come out of beta around PyCon). You can think of it as the same distinction between `flit-core` & `flit` or `poetry-core` & `poetry`, with `hatchling` being more similar to `flit-core` in that only accepted standards are used, like for dependency specification ([PEP 631](https://www.python.org/dev/peps/pep-0631/)). One of the central differentiators is the extensibility offered by its [plugin system](https://ofek.dev/hatch/latest/plugins/about/) (see [the showcase](https://github.com/ofek/hatch-showcase)).

Note that the only reason the version is 0.x is because 1.0.0 will drop support for Python 2. It is quite stable, and I'm in the process of switching all of our Python packages here at Datadog. Here's one of the repos that we just completely switched (~150 packages): https://github.com/DataDog/integrations-core

## Differences

In comparison to https://pypi.org/project/pipx/1.0.0/#files

### Wheel

Hatchling already supports the final draft of [PEP 639](https://www.python.org/dev/peps/pep-0639/) so license files are stored in `.dist-info/license_files/` rather than dumped directly in `.dist-info/`.

### Source distribution

Even though the old `MANIFEST.in` says otherwise, `*.md` files weren't actually included. Here's a before & after:

sdist


ofek:
There is a CI check that verifies source distributions contain everything in the repo, which is odd because the PyPI release definitely does not.

I can either remove that check (especially since [the tool](https://github.com/mgedmin/check-manifest) mentions a `MANIFEST.in` but that file no longer exists) or have it ship everything in the repo. Just let me know!

cs01:
The check can be removed now since we're not longer using MANIFEST

ofek:
Done!

pyproject.toml
```diff
@@ -1,6 +1,61 @@
 [build-system]
-requires = ["setuptools", "wheel"]
-build-backend = "setuptools.build_meta"
+requires = ["hatchling>=0.14.0"]
```
gaborbernat, line 2: Why not flit if we're going down this path? I know that has a better story for OS repackaging.

ofek: By OS repackaging, do you simply mean the availability of a system package for different Linux distributions like Debian? If so, as you might have seen in that other thread there is one for [Debian](https://salsa.debian.org/python-team/packages/hatchling) and the Fedora one is [nearly complete](https://github.com/ofek/hatch/issues/120#issuecomment-1043048369). For others like CentOS ([7 & 8-stream](http://mirror.centos.org/centos/7/os/x86_64/Packages/) / [9-stream](http://mirror.stream.centos.org/9-stream/AppStream/x86_64/os/Packages/)) I don't see `pipx` nor `flit`.

I appreciate Flit, and what it has accomplished. It essentially was the shining example of PEP 517, illustrating that `setuptools` was no longer a requirement for pure Python projects!

In many cases though, projects need more, which is where Hatchling comes in.

For example, many popular projects (including some you maintain 🙂) require [setuptools_scm](https://github.com/pypa/setuptools_scm) to use VCS tags exclusively for the project version. Hatchling has [a plugin](https://github.com/ofek/hatch-vcs) for this (despite the legacy naming, Ronny has taken great care to decouple `setuptools_scm` from `setuptools`). `pipx` may want this functionality in the future.

Or, if a project wanted to use [Mypyc](https://github.com/mypyc/mypyc) to compile code with the help of type hints, there is [a plugin](https://github.com/ofek/hatch-mypyc) for that too.

Or, if it is desired that the project page on PyPI displays the README and changelog (you'll often see them concatenated in `setup.py`), a custom [metadata hook](https://ofek.dev/hatch/latest/plugins/metadata-hook/) can easily be written to achieve this.

Essentially, I've put much effort into ensuring arbitrary use cases are easily supported.

ofek: > going ultra-simple and sticking to a pypa tool

1. Both use PEP 621 metadata so the "simplicity" for metadata is identical. Defining the contents of source distributions in Flit also requires [its own section](https://flit.readthedocs.io/en/latest/pyproject_toml.html#sdist-section). Finally, Flit does not currently support the layout of `pipx` because there is a strict expectation of a variable `__version__` at the package root:

   ```
   ofek@ozone ~\Desktop\code\pipx modernize-metadata $ python -m build
   ...
       raise NoVersionError('Cannot package module without a version string. '
   flit_core.common.NoVersionError: Cannot package module without a version string. Please define a `__version__ = "x.y.z"` in your module.
   ```
   If support is implemented, it would also require [its own section](https://github.com/pypa/flit/pull/382#issuecomment-874723398).
2. I get the sense that since Flit now lives under pypa/ it is being recommended here as the preferred or standard tool. If that is indeed what's happening (correct me if I misinterpreted), then I guess that's the future of Python packaging and I'll have to accept that 🙂 Though, that goes against everything in the discussion about the transfer: https://discuss.python.org/t/should-flit-become-a-pypa-project/11854

   > If PyPA did adopt Flit, I’d like to avoid any suggestion that it’s the ‘official’ or ‘standard’ tool for creating & publishing packages.

   Perhaps to be recognized as legitimate, backends should first apply/ask to be included in PyPA? Based on:

   > I think it would be good to have multiple backends under the PyPA umbrella, _precisely_ to avoid giving any impression that there’s a “PyPA preferred” option.

henryiii: No, but _this package is a PyPA project_. If something horrible breaks, we have a little more of a say in fixing it. There are also other PyPA projects planning on moving to Flit this year.

ofek: I see. Should we document the fact that PyPA projects must use a PyPA build backend somewhere?

henryiii: And there are a lot of repercussions for distro packagers that don't support PEP 517 - we are already having to convince them to include flit, also convincing them to include hatchling is likely not the correct approach. The correct approach is for them to use installer, but that's not ready yet. Maybe helping with that would really help out all PEP 517 backends (at least with pure wheels). See https://github.com/pypa/wheel/pull/436

conda-forge doesn't include hatch/hatchling, so this would immediately break the conda-forge recipe, while flit is available there. Etc.

henryiii: I am not saying that hatchling cannot be used by PyPA projects. I'm just giving an opinion and warning that at [brand new](https://pypistats.org/packages/hatchling) backend with seven dependencies is likely to be more risky than using something that has been around for years, has been involved in the PEP process, was popular enough to make it into the PyPA, and has zero dependencies. Maintaining a backend is hard, and a packages is only as compatible as its backend. There is a _lot_ involved, like PyPy support, and having lots of users and lots of experience working through all the exotic situations is quite important. I wish Hatchling the best, and was happy to add support for it to scikit-hep/cookie, but "it supports PEP 621" doesn't mean you can throw out the rest of the considerations of picking a good dependency. Flit has the added bonus of being "in house" for another PyPA project, and several projects are planning on going through the pain of conversion together. I'm quite sure Flit will continue to support new Python versions, new Python distributions (like Pyodide), etc. Maybe Hatch will too, but I'd like to see with more than 10 days of heavy use.

I am not a pipx maintainer, I do not have any say, I'm just providing some advice. The pipx maintainers might happy use Hatchling after reading this advice, and I would be happy for them. Though if it breaks GitHub Action's recipe, I'll be furious. ;)

Just curious, do you know why hatchling suddenly started getting more than 100 downloads per day 10 days ago? Did a major project pick it up?

ofek: > Just curious, do you know why hatchling suddenly started getting more than 100 downloads per day 10 days ago? Did a major project pick it up?

Yes, after I deemed it stable (esp. config options) I made an RFC internally proposing modernizing packaging (in particular for integrations) and we chose Hatchling over the other backends. As mentioned in the OP, this repo is almost done switching: https://github.com/DataDog/integrations-core

After the integration repos then I'll do the other Python projects we own, though those are mostly private so I cannot provide links.

gaborbernat: > By OS repackaging,

I've primarily been referring to ease of bootstrapping hatch within a Linux distribution 🤔

gaborbernat: I'd be happy to use hatch over flit if you could fix the Linux distribution bootstrap problem for Arch/RHEL/Debian maintainers.

henryiii: For reference, here's the plan for build: https://github.com/pypa/build/issues/394#issuecomment-1044550493

Don't forget conda-forge, WebAssembly efforts like Pyodide, etc.

layday: pipx isn't a bootstrapping dependency, not sure why that would matter here.

henryiii: > As in vendoring or is there some other way?

I don't think arbitrary backends need to support bootstrapping, and very much should not be vendoring.

gaborbernat: I think I got confused here, as long as flit is bootstrap-able (aka has no dependencies) and build project is using build, any distribution can use build + pip/installer to generate a OS package for any other build backed (including hatch).

pyproject.toml
```diff
@@ -1,6 +1,61 @@
...
+    "Virtual Environment",
+]
+authors = [
+    { name = "Chad Smith", email = "chadsmith.software@gmail.com" },
+]
+classifiers = [
+    "License :: OSI Approved :: MIT License",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3 :: Only",
```
gaborbernat, line 25: One feature gap compared to setuptools and `setup.cfg` I see here is that this file is now manually maintained and formatted while using `setuptools` one can use the excellent `setup-cfg-fmt` project for it. Does hatch have plans to add formatting support? Without it I'm tempted to keep using setuptools :thinking: Naturally if other maintainers even with that feature gap like setuptools I'm not going to oppose :+1:

ofek: I _suppose_ I could add a command for that but:

1. Something like that is indeed better suited as a small, separate project that can be used by all.
2. I would personally never use it because I would like to retain comments, and `tomlkit` is extremely buggy.

Are you saying that until there is a `pyproject.toml` linter you're not keen on using PEP 621?

gaborbernat: Indeed :blush:

gaborbernat (approved):

pyproject.toml
henryiii: Setup-cfg-fmt does a lot more than just formatting, it also keeps up the classifiers for new Python versions and cleans them up, sorts dependencies, etc. It also runs in pre-commit, can taplo?

It also strips comments, valid empty (sub) sections, and can’t be disabled for part of a file AFAIK…

pyproject.toml
henryiii: The original request was: “One other thing I've been wanting to do for a while is bring back the install script where you can run `curl ... | python3`. That is originally how pipx worked.“ - that’s part of where talk of bootstrapping came from, I think. Though that’s not quite bootstrapping.

Anyway, installer will really help for using arbitrary backends, since then distros can use the wheel to package without ever touching the backend.

henryiii:
I think hatchling should be added to conda-forge before this goes in, or installer. This will break the current conda-forge package. Flit(+core) are on conda forge, but not hatchling.

pyproject.toml
henryiii: > tomlkit is extremely buggy.
👍

An eventual tool should not be tied to a build backend, but be a generic tool; probably a pair of tools like taplo & (imaginary atm) pep621-fmt or something. Happy to have it live alongside hatch though. ;)

gaborbernat: > 2\. I would like to retain comments

I'm personally not a fan of comments in configuration files like these, so for me less of an issue. 😆

ofek:
Done!

- https://github.com/conda-forge/hatchling-feedstock
- https://anaconda.org/conda-forge/hatchling

henryiii:
I no longer have any objections; I don't see a strong reason to use hatch over flit for something this simple, but now I don't see a strong reason to use flit over hatch either. The addition of proper PyPy testing on Windows in Hatch is also really nice to see.

pyproject.toml
henryiii: There are some very good reasons to put comments in sometimes, like https://github.com/scikit-hep/awkward-1.0/blob/6d3c1d23d7ab492d0178b1798e179088a066e00a/pyproject.toml#L90-L157 - probably rare for the project metadata itself, but the other tool fields might need comments (flake8/pylint config being big ones for needed comments). Comments can be overdone, sure, but sometimes they are needed. setup-cfg-fmt doesn't support them for technical reasons, not primarily because they are a bad idea. I usually have managed without comments in setup.cfg, though, sometimes painfully with flake8, but really hope not to leak that issue over into pyproject.toml. :)

q0w:
> I no longer have any objections; I don't see a strong reason to use hatch over flit for something this simple, but now I don't see a strong reason to use flit over hatch either. The addition of proper PyPy testing on Windows in Hatch is also really nice to see.

Choosing `flit` you can stick with one and only one tool and spend your time to improve one and only one tool.

ofek:
> Choosing `flit` you can stick with one and only one tool and spend your time to improve one and only one tool.

By that logic we all should just contribute to `setuptools` 😉

henryiii:
>  we all should just contribute to `setuptools` 😉

Good luck, `setuptools` has 40K+ lines of Python code, `flit-core` has 2K. Most of the code is deeply entrenched in legacy usage and is a pain to touch, and it has a distutils / setuptools structure that is really irritating. In the future, other build systems might get there too, but we've learned a lot about limiting API in the years since (partially because we now have choice, and we don't need a "must be usable by everyone" build system).

I am a big fan of contributing to a project instead of starting a new one (which is why I'm part of many of the projects I'm part of!), but flit's goal is to stay simple, so that's why other "simple but not so simple" build systems can exist. That's why Flit can't support VCS versioning; there is no "build" step where you can add that.

I do think we are getting a few too many "simple" build systems, though - pdm's, hatchling, flit-core, trampolim, whey... I really like what PDM (and Hatch?) do by supporting _any_ PEP 621 build system inside the environment manager, rather than requiring their own (though PDM does have one).

ofek:
> I really like what PDM (and Hatch?) do by supporting _any_ PEP 621 build system inside the environment manager

Correct, Hatch does that too. Interoperability is very important to me https://ofek.dev/hatch/latest/meta/faq/#interoperability

q0w:
So maybe PDM instead of hatchling?😉

gaborbernat:
I'm definitely in the hatchling over pdm group as maintainer. 👌

pyproject.toml
abravalheri: > I would personally never use it because I would like to retain comments, and tomlkit is extremely buggy.

`tomlkit` has received a lot of improvements over the last 6 months. I personally have contributed several patches and @frostming has become a maintainer and has been doing an amazing job!

As an example, I can safely use it for [ini2toml](https://ini2toml.readthedocs.io/en/latest/) for translating the "AST" of `.ini` files to `.toml`.

So if its being a while that you last checked `tomlkit`, maybe it is a good time to have a look if it improved for your use cases?

gaborbernat: For what's worth I've been using `tomlkit` for https://github.com/gaborbernat/pyproject-fmt and works well for me 👍

abravalheri:
> Choosing `flit` you can stick with one and only one tool and spend your time to improve one and only one tool.

The whole point of PEP 517 was to make possible to have diversity on the ecosystem... This way of thinking seems to defy the standardisation efforts.

pyproject.toml
frostming: Run into this due to the ping, and I found @gaborbernat already made a [pyproject.toml formatter ](https://github.com/gaborbernat/pyproject-fmt), powered by `tomlkit`

layday:
I don't know if it's appropriate for me to review a PR here but I don't see anything wrong implementation-wise.

pyproject.toml
henryiii: I tried to use toolkit in cibuildwheel (at least 6 months ago, probably more), and found it too buggy to use; it was a rather weird format, but it was important to keep it, which it failed to do. I ended up just writing a manual writer, since it was a very well structured limited file. Great to hear it is seeing improvements! I know it was not 1.0.0 compliant either last I checked, but is now. :)

And @gaborbernat, @abravalheri, and @ofek have all made TOML formatter/validators. :) `pypyroject-fmt`, `pyproject-validate`, and `validate-pyproject`.

henryiii:
I don't think I need to say more than I've already said, no need for an explicit review. If this was my repo, I'd still be lightly biased to keep this "in house", and use flit-core - but it's practically no change, due to PEP 621, so it could even be done later if hatchling maintenance suddenly ends or something. Most users will use wheels, so various backends are pretty safe for pure Python projects. I don't have any reservations over using hatching. And the implementation looks good.

henryiii:
PS: just for completeness, since I mentioned it before, I did check GHA/Azure's wheel setup, and they use pip to install pipx, so they should not be affected.

cs01:
Thanks all for your participation. If any one is interested in becoming a maintainer, reviewing other pull requests, or triaging issues, please let me know! I can be reached at chadsmith.software@gmail.com.

cs01:
Somehow this broke CI https://github.com/pypa/pipx/runs/5339892213?check_suite_focus=true. Is anyone able to investigate/fix?

ofek:
Seems like a flaky test; I'd highly recommend https://github.com/pytest-dev/pytest-rerunfailures

cs01:
> Seems like a flaky test; I'd highly recommend https://github.com/pytest-dev/pytest-rerunfailures

Sounds like a good idea. If anyone has the bandwidth to try and fix this I would really appreciate it. I'm not sure when I can get to it unfortunately.

meowmeowmeowcat:
> Somehow this broke CI https://github.com/pypa/pipx/runs/5339892213?check_suite_focus=true. Is anyone able to investigate/fix?

Seems related to https://github.com/ipython/ipython/issues/13554

ofek:
@cs01 I opened https://github.com/pypa/pipx/pull/822

PR #822, opened 2022-02-28
Retry flaky tests

ofek:
Motivation: https://github.com/pypa/pipx/pull/817#issuecomment-1053944420

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -119,7 +119,7 @@ def tests_with_options(session, net_pypiserver):
         *pypiserver_option,
         "--cov=pipx",
         "--cov-report=",
-        "--reruns=5",
+        "--reruns=2",
         "--reruns-delay=1",
         *tests,
     )
@@ -150,7 +150,7 @@ def test_all_packages(session):
         "--show-capture=no",
         "--net-pypiserver",
         "--all-packages",
-        "--reruns=5",
+        "--reruns=2",
         "--reruns-delay=1",
         *tests,
     )
--- a/noxfile.py
+++ b/noxfile.py
@@ -114,7 +114,15 @@ def tests_with_options(session, net_pypiserver):
         refresh_packages_cache(session)
         pypiserver_option = []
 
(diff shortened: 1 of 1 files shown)


ofek:
CI failure due to: https://github.com/ipython/ipython/issues/13554

noxfile.py
```diff
@@ -142,6 +150,8 @@ def test_all_packages(session):
         "--show-capture=no",
         "--net-pypiserver",
         "--all-packages",
+        "--reruns=5",
```
uranusjr, line 153: Do we need this many? I’d start with 2.

cs01:
Before we go to retries, we should look to understand the root cause of the problem. I would rather hold off on landing this before we try other methods, since this is adding more tech debt. I appreciate the PR and if you can’t root cause it, that’s fine, but I’d prefer creating an issue instead and closing this PR, especially because CI tests still failed.

q0w:
https://github.com/pypa/pipx/pull/817#issuecomment-1054022864

ofek:
Basically, iPython needs to release the fix https://github.com/ipython/ipython/issues/13554#issuecomment-1053641599

I think retries are still a useful thing to add, though it's up to you!

uranusjr:
From experience, subprocess (also `os.fork` and `multiprocess` which all share a same vein under the hood) sometimes just fail like this due to OS internal states, and I’ve yet to see anyone come up with a good solution to make this always work. Retry is the usual solution for this kind of things.

cs01:
> From experience, subprocess (also `os.fork` and `multiprocess` which all share a same vein under the hood) sometimes just fail like this due to OS internal states, and I’ve yet to see anyone come up with a good solution to make this always work. Retry is the usual solution for this kind of things.

We have had zero of such issues with pipx so far. I just looked into the failures we're seeing a little, and the root cause is a repeatable python3.8/ipython issue. https://github.com/ipython/ipython/issues/12677

I appreciate the PR, but I am going to close it since I believe it is trying to solve a problem that does not exist, and is adding tech debt..

PR #824, opened 2022-03-04
Add a CACHEDIR.TAG to the cache directory.

KenMacD:

## Summary of changes

Tag the cache directory to prevent the files from being included in
archives and backups.

## Test plan
Tested by running `pipx` and verifying the file was created, then running it again to verify that the file isn't recreated. Also tested with the `cachdir-tag` library's method `is_tagged()`.


gaborbernat (changes requested):
You'll need to include a test for this 👍

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - [dev] Change github action job names
 - [docs] Add additional examples for installation from git repos
 - [packaging] Switch to [PEP 621](https://www.python.org/dev/peps/pep-0621/)
+- Add CACHEDIR.TAG to the cache directory
```
cs01, line 13: Can you add something like "For information about cache directory tags, see   https://bford.info/cachedir" like you did in the code? I've never seen this before, and I am not sure how many others haven't either.

uranusjr (approved):

Issue #219, opened 2019-09-18
pipx install doesn't prompt for credentials for private vcs url


alexkillen:
**Describe the bug**
Regular `pipx install` command doesn't work for private vcs urls that require authentication. Running `pipx install --spec git+https://<some_private_repo>.git <package_name>` does not prompt the user for username and password.  If the `--verbose` option is provided, it works fine.

**How to reproduce**
Run above command using a private vcs url (e.g. github, gitlab, bitbucket) and do **not** use `--verbose`.

**Expected behavior**
Should prompt user for credentials, even when `--verbose` is not used.

**Environment**
```console
$ python3 --version
Python 3.6.8
$ pipx --version
0.14.0.0
$ head -n 2 /etc/os-release
NAME="Ubuntu"
VERSION="18.04.3 LTS (Bionic Beaver)"
$ uname -srm
Linux 4.15.0-64-generic x86_64
```

cs01:
My guess is that this is because the pipx animation takes over the terminal and hides the prompt for the password. The animation runs until the subprocess command finishes. Does it just animate forever and never finish? 

The fix would be to not do the animation. Ways to do this:

* Use verbose flag as you discovered
* Add flag or env var to pipx to suppresses animation
* Never show animation
* Document problem and workaround in cli help text

alexkillen:
Yes, it animates forever and never completes.  As for a solution, a good compromise might be to disable animations when the `--spec` argument is used.  If that's not an option I'd lean towards a `--no-anims` type flag or env var to disable animations, as using `--verbose` is not that intuitive, I only discovered that worked while trying to debug the issue.

It would of course be nice if pipx could detect when user prompts occur and drop out of the animation when they do, but that sounds like it would require considerable effort.

btw, pipx is great, the Python ecosystem has needed this for years, too many people try to force pip and pypi into this role and it causes a lot of confusion. Keep up the great work.

cs01:
@uranusjr how does pipenv solve this?

alexkillen:
Had a quick check through the pipenv source and it looks like it's using this: https://github.com/sarugaku/vistir#spinner

cs01:
Thanks!

What happens when installing from a protected source with pipenv? Does it spin indefinitely or does it stop spinning and drop back to the user/password prompt? If it drops back to the password prompt i'll take a look around the sarugaku source code to see how it does it.

cs01:
[...]

PR #829, opened 2022-03-18
Add --shared to `pipx install` and `pipx runpip`.

Darsstar:
Hi,

Is this something you folks would be willing to maintain?

Alternative aproaches to make pipx more suitable for private registries you might prefer which I am willing to implement:
- make keyring special and always share it
- support PIPX_PYTHONPATH

If there is interest in maintaining anything like this I will add tests, if not I will maintain this as my private fork without going through the effort of adding tests.

* [x] I have written documentation
* [x] I have written tests

## Summary of changes
Add --shared to `pipx install` and `pipx runpip`.

The main use case for it is as explained in the help text: providing keyring backends while creating virtual environments while pip is looking at private repositories which require authentication.
PYTHONPATH is not an option since pipx explicitly removes it from the dict of environment variables to be passed to the subprocess.
Injecting it after it is created would ia a chicken and the egg problem.

It still won't work out of the box. But now it should no longer be a deal-breaker and instead be, assuming PyPI is reachable, a slight hassle after installing pipx. Documentation should be able to alleviate this further.

## Test plan

Tested by running
```
pipx install -i https://pypi.org/simple/ --shared keyring-subprocess
pipx install -i https://pypi.org/simple/ keyring
pipx inject -i https://pypi.org/simple/ --include-apps keyring keyring-subprocess-landmark
pipx inject -i https://pypi.org/simple/ keyring artifacts-keyring

# now we are good to go! This should list pip, wheel, setup-tools and keyring-subprocess 
pipx runpip --shared list

pipx install black
pipx install ...
```

While the following would achieve the same thing, I hope we can all agree the above is better in the sense that we don't use `pipx runpip` and are thus able to use `pipx upgrade`. (Also
```
pipx runpip --shared install keyring-subprocess
```


cs01:
Thanks for putting this together, and apologies for such a long wait time on the review (we are all volunteers so I hope you can understand).

I haven't used private pypi indexes before so I had to do a little reading to come up to speed on this, but I think I'm there now. For anyone else not familiar, keyring integrates with your OS's password manager, and pip will automatically load credentials for urls in saved in keyring [reference](https://pip.pypa.io/en/stable/topics/authentication/#keyring-support).

I see there is also a [netrc](https://pip.pypa.io/en/stable/topics/authentication/#netrc-support), which should be automatically used by pip. Is there a reason you aren't using this instead? (I'm guessing because storing your password in plaintext isn't desirable, but wanted to confirm this was considered as an option).

Darsstar:
Plaintext is one reason, yes. The other is that Azure DevOps Personal Access Tokens expire. 

Microsoft's artifacts-keyring keyring backend handles that expiration problem. 

Pipx solves weird dependecy conflicts I have been asked to help with. Artifacts-keyring reduces authentication related issues for me to help colleagues with by a lot.

Having these two (pipx & keyring) work together better nice for me, since I get to clean up our powershell script. (See the [keyring-subprocess](https://pypi.org/project/keyring-subprocess/) package's Readme if you are interested)
For other companies it might be the difference in whether they choose to use this excellent tool or not because they gave up sooner then I did.

Until I hear otherwise I'll assume I provided enough motivation for the feature and start working my way through the checklist.

Darsstar:
Tests, documentation and changelog enties are now all ready for review as well.

I'm also mention my related PR for Pip https://github.com/pypa/pip/pull/11029 in the hope that someone ends up there via this PR. As mentioned there I hope to create a PR that passes`--no-input` unless `--verbose` is passed as you mentioned in https://github.com/pypa/pipx/issues/219#issuecomment-570127281.

docs/installation.md
```diff
@@ -35,6 +35,50 @@ sudo PIPX_HOME=/opt/pipx PIPX_BIN_DIR=/usr/local/bin pipx install PACKAGE
...
+
+In order to have keyring provide credentials the [keyring](https://pypi.org/project/keyring/) package needs to be importable immediately after Pipx created a virtual environment using Python's venv module.
+
+For that first install [keyring-subprocess](https://pypi.org/project/keyring-subprocess/) since, unlike keyring, it has zero dependencies.
+
+Secondly we install keyring and inject keyring-subprocess-landmark and the desired keyring backend into its venv.
+
+```
+# install keyring-subprocess, which is somwhat magical but needs a keyring-subprocess executable 
+pipx install -i https://pypi.org/simple/ --share keyring-subprocess
```
jack1142, line 66:
```suggestion
pipx install -i https://pypi.org/simple/ --shared keyring-subprocess
```

Darsstar: I fixed both, thanks!

pfmoore:
Note that in the next release, pip will include a `--python` option and will be supported when run from a zipapp. Those two changes will open up opportunities for changing the current shared-libs approach (which has a negative effect of making the shared libraries visible in every application venv, which might be an issue for apps that, for example, depend on a specific version of pip, which is unlikely but possible).

As a result, making the existence of the shared libraries visible in the user interface may be something we want to avoid, as it will make it harder to change.

Darsstar:
I should probably repeat here what I said about this PR in my "vendor keyring & keyring-subprocess" PR: if that one gets merged this one would no longer be of any value to me. (And presumably potential Pipx users that want to retrieve credentials for private repositories with the keyring library.)

That is because it is more general: [pyprojectx](https://pyprojectx.github.io/), something a colleague pointed out as something we should take a look at, would become viable as soon as we updated Pip.

Darsstar:
Pip 23.1 included the `--keyring-provider` flag I contributed. Which makes this PR obsolete.

PR #851, opened 2022-06-12
add python 3.11 to ci workflows and tests

cs01:

## Summary of changes
Add python 3.11 to github workflows and test matrixes

## Test plan

CI tests pass


uranusjr:
I’ve made CI work, but tests are failing left and right 😆

meowmeowmeowcat:
> tests are failing

It's because pypisever doesn't work properly on Python 3.11.

meowmeowmeowcat:
Added in #917

PR #856, opened 2022-06-22
Fix examples in the docs

meowmeowmeowcat:

## Summary of changes
Fix `pipx run` examples in the docs and update Python versions used by `pipx install` examples
## Test plan

Tested by running
```
pipx run --python python3.9 black
```


cs01:
Looks good, thanks!

PR #871, opened 2022-07-13
Update `testdata/tests_packages` and generate missing ones

jack1142:

## Summary of changes

I regenerated files in `testdata/tests_packages` as I've noticed that it lacked package lists for Python 3.10 while running pipx's workflows. It seemed reasonable to just regenerate all of it rather than only attach new files. I also deleted the files that are no longer being generated by the workflow - `macos19-*.txt` and `*-python3.6.txt` files.

To fix an issue with the missing macOS Python 3.10 wheel for lxml 4.9.1 (that causes failures in installing jaraco-financial that is part of the test suite), I've manually altered the `macos20-python3.10.txt` and replaced `lxml==4.9.1` with `lxml==4.9.0`.
Alternatively, I could add it to `primary_packages.txt` and regenerate all tests_packages files.

Since I couldn't find an existing issue, I've also made one for tracking the lacking wheels for lxml 4.9.1 here:
https://bugs.launchpad.net/lxml/+bug/1981630

## Test plan

I would suggest that a maintainer:
1. Manually triggers a workflow run that generates package lists:
https://github.com/pypa/pipx/actions/workflows/create_tests_package_lists.yml
2. Once the workflow run finishes, downloads the created zip with artifacts.
3. Unpack it.
4. Compare it with the data that's in this PR and verify that it's equivalent.

Seems like the best way to ensure that what I've done was actually generated with that workflow, not manually.
There's only one expected difference related to lxml on macOS with Python 3.10:
```diff
diff --git a/testdata/tests_packages/macos20-python3.10.txt b/testdata/tests_packages/macos20-python3.10.txt
index bcd5a98d..cb005068 100644
--- a/testdata/tests_packages/macos20-python3.10.txt
+++ b/testdata/tests_packages/macos20-python3.10.txt
@@ -86,7 +86,7 @@ kaggle==1.5.12
 keyring==23.6.0
 keyrings.alt==4.1.0
 lazy_object_proxy==1.7.1
-lxml==4.9.1
+lxml==4.9.0
 matplotlib_inline==0.1.3
 mccabe==0.6.1
 mccabe==0.7.0
```

You should probably also verify that the `tests *` workflows actually use those files, as in they print a line like this:
```
Using the following file to specify needed package files:
    testdata\tests_packages\win-python3.10.txt
```
I already checked that myself too of course.


uranusjr:
I think we might want to keep macos19 for people running tests locally

jack1142:
> I think we might want to keep macos19 for people running tests locally

I added those back (without updating versions nor adding Python 3.10 for it) but note that macos19 applies to macOS 10.15 which will most likely be EOLed in just a few months (by the end of the year) sometime after macOS 13 gets released.

Versions currently supported by Apple are macOS 10.15, macOS 11, and macOS 12 (note that for macOS 12 there are currently no files generated; if you want me to add them, this could be arranged though).

uranusjr:
> note that for macOS 12 there are currently no files generated; if you want me to add them, this could be arranged though

Maybe it’s easier to add a nox script or pytest hook for testers to generate their own file if needed. This would avoid us needing to supply those files. Let’s do that separately though.

uranusjr (approved):

jack1142:
These are generated automatically when missing, it's just slower if they have to do that.

PR #874, opened 2022-07-19
Add `install-all` command

### LINKED ISSUE #687: Support provisioning of pipx-managed sites from the output of `pipx list --json`.
With the completion of #660 , the possibility now exists to re-create one or more pipx-managed environments based on the output of `pipx list --json`.  

This would be useful for backup or versioning across one or more machines.  For instance, I could use the output of `pipx list --json` from workstation A to quickly provision workstation B  or laptop C with a same/very similar set of tools. 

I don't have firm opinions on the API here, but i would kind of naively expect `pipx install-all my-pipx-list.json`, which creates the new subcommand `install-all`, with name parity with existing `-all` sub-command equivalents.

Tricky points I can think of: 
 - what if workstation B has a different version of one or more of the top-level packages in the exported list? who wins?
 - what if there are interpreter version incompatibilities in a top-level package or one or more (potentially pinned) deps or injected packages?
 - what about packages installed from local tarballs, etc?

JSON is supposed to be human readable, right? :)   I could certainly hack something together with `zsh` and  `jq` that walks the tree and installs root + injected, but it seems natural enough to absorb this into `pipx` itself.

dukecat0:

## Summary of changes

Closes #687

## Test plan

Tested by running
```
pipx install-all packages.json
```


A1vinSmith:
Thank you for this. Although, most of repos that I want to run it doesn't have the `packages.json`.
I want to run `pipx install-all requirements.txt`

uranusjr:
Even on where the manifest does exist, it’s still tedious for most users to have to dig out the file manually. It’s probably better for us to add something like `pipx freeze` (or export? something else?) to couple with this.

A1vinSmith:
> Even on where the manifest does exist, it’s still tedious for most users to have to dig out the file manually. It’s probably better for us to add something like `pipx freeze` (or export? something else?) to couple with this.

That sounds nice. Maybe even `pipx install-all-rtxt` that includes two commands above.

meowmeowmeowcat:
> `pipx install-all packages.json`

This `packages.json` is the output of `pipx list --json`, and it's a little bit different from `requirements.txt`, but I am +1 on supporting `requirements.txt`.

uranusjr:
Supporting requirements.txt would probably be unnecessarily complicated since the format lacks a lot of structure we need to for metadata besides just packages. I’d stick with just supporting our own JSON manifest.

meowmeowmeowcat:
> It’s probably better for us to add something like `pipx freeze` (or export? something else?) to couple with this.

I guess the output of `pipx freeze` would be quite similar to the output of `pipx list --json`, while `pipx freeze` doesn't have to contain info that wouldn't be used by `pipx install-all`.

uranusjr:
Yeah, and also it can have options like `--output` to make it easier to forward things to a file without shell syntax.

xeor:
This is exactly what I'm looking for! :)

Should python-version be read from the dump-file? Is it better to not include that and let the install command figure it out itself? What if the version doesnt exist on the system?

meowmeowmeowcat:
> Should python-version be read from the dump-file?

Some packages have limitations on the Python version, so python-version has to be read from the file.

> What if the version doesnt exist on the system?

They will get an error and then they can install the package manually with `pipx install <packages>`

halostatue:
This would solve most of what I would want in #952, but would it also automatically inject packages?

```
   package mkdocs 1.4.2, installed using Python 3.11.3
    - mkdocs
    Injected Packages:
      - mkdocs-material 9.1.8
      - mkdocs-mermaid2-plugin 0.6.0
      - mkdocs-redirects 1.2.0
      - mkdocs-simple-hooks 0.1.5
```

From my perspective as well, it would be useful to have *missing* Python version values assume the use of the (default) version, as most of the packages I install are simply done with `pipx install <package>` and haven’t yet found one that requires a specific Python (which is good, because I don’t actually plan on having multiple versions installed; I don’t develop software in Python, but I happily use many of the amazing tools).

meowmeowmeowcat:
> but would it also automatically inject packages?

Currently no, but I will implement it in the future.

gaborbernat (changes requested):
Merge conflicts 👍

gaborbernat:

Issue #705, opened 2021-06-13
pipx install "cannot find the file specified"


kcinlober:
when using `pipx install cowsay` (or any program) i get an error stating that the file cannot be found in the .local\pipx\shared directory (see --verbose code)

```
C:\WINDOWS\system32>pipx install cowsay --verbose
pipx >(setup:717): pipx version is 0.16.3
pipx >(setup:718): Default python interpreter is 'c:\\users\\nbreb\\appdata\\local\\programs\\python\\python39\\python3.exe'
pipx >(package_name_from_spec:323): Determined package name: cowsay
pipx >(package_name_from_spec:324): Package name determined in 0.0s
creating virtual environment...
pipx >(run_subprocess:135): running c:\users\nbreb\appdata\local\programs\python\python39\python3.exe -m venv --without-pip C:\Users\nbreb\.local\pipx\venvs\cowsay
creating shared libraries...
pipx >(run_subprocess:135): running c:\users\nbreb\appdata\local\programs\python\python39\python3.exe -m venv --clear C:\Users\nbreb\.local\pipx\shared
Error: [WinError 2] The system cannot find the file specified

pipx >(rmdir:43): removing directory C:\Users\nbreb\.local\pipx\venvs\cowsay
'c:\\users\\nbreb\\appdata\\local\\programs\\python\\python39\\python3.exe -m venv --clear
C:\\Users\\nbreb\\.local\\pipx\\shared' failed
```

I believe i installed pipx correctly (as per instructions) and same with pip (21.1.2)

itsayellow:
Hi, some questions:
1. Which version of Windows are you using?
2. Are you using a native Powershell, cmd.exe, MSYS2, or Cygwin to run pipx?
3. How was the python you're using installed, via a python.org installer?  Or through some other package manager?
4. Could you post the contents of the `cmd_*.log` file that pertains to this command?  It should be located in `C:\Users\nbreb\.local\pipx\logs`.  Sometimes there is more information in that than the `--verbose` output.

thanks!

itsayellow:
A quick thing to check, does `c:\users\nbreb\appdata\local\programs\python\python39\python3.exe` exist?

malteohlers:
@itsayellow I have run into the same issue on a newly installed Windows 10 machine:
```PS C:\Users\MalteØhlers\.local\pipx\venvs> pipx install cowsay --verbose
pipx >(setup:717): pipx version is 0.16.3
[...]

PR #879, opened 2022-07-29
Recommend windows users use the PEP 397 python launcher

rmcclellan-izo:
This is a doc-only change, fixing issue #705.

The current windows install instructions won't work for windows users using python.org installed versions of python below recent versions of 3.9 and 3.10, due to https://github.com/python/cpython/issues/87915 .  According to [this discussion](https://discuss.python.org/t/microsoft-store-package-does-not-add-the-py-exe-launcher/538/11),

> Right now, approx 95% of the downloads go through [python.org](http://python.org/).

Thus, most Windows users will have access to the [PEP 397](https://peps.python.org/pep-0397/) `py` launcher tool, and will _not_ have access to `python3`.  So `py` is probably a better choice for installation instructions.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -18,8 +18,8 @@ pipx ensurepath
 Otherwise, install via pip (requires pip 19.0 or later):
 
 ```
-python3 -m pip install --user pipx
-python3 -m pipx ensurepath
+py -3 -m pip install --user pipx
+py -3 -m pipx ensurepath
 ```
 
 ### Installation Options


docs/installation.md
```diff
@@ -18,8 +18,8 @@ pipx ensurepath
 Otherwise, install via pip (requires pip 19.0 or later):
 
 ```
-python3 -m pip install --user pipx
-python3 -m pipx ensurepath
+py -3 -m pip install --user pipx
+py -3 -m pipx ensurepath
 ```
```
uranusjr, line 23: This block is describing not just Windows but other platforms as well. You need to add a new paragraph insterad.

rmcclellan-izo: Thanks, I've added a new paragraph.

uranusjr (approved):

rmcclellan-izo:
Hi - I'm not sure why the tests are failing - it doesn't seem related to my doc-only changes.  Let me know if there's anything I can do to help move this PR forward :-)

uranusjr:
Looks unrelated, don’t worry about it.

PR #881, opened 2022-07-30
Match pip's behavior when package name looks like an archive

### LINKED ISSUE #641: Install a .whl file containing console entry point

If a user creates a `.whl` file that contains an [`entry_point` `console_script`](https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-console-scripts-entry-point), normally they can run:

```bash
(venv) $ pip install myproject-0.1.0-py3-none-any.whl
```

And then those entry points are available within the virtual environment.

But when I try to do something similar with `pipx`:

```
$ pipx install myproject-0.1.0-py3-none-any.whl                               
    ~/.local/pipx/logs/cmd_2021-03-02_14.39.35_pip_errors.log

    ERROR: Could not find a version that satisfies the requirement myproject-0-1-0-py3-none-any-whl
    ERROR: No matching distribution found for myproject-0-1-0-py3-none-any-whl

Error installing myproject-0-1-0-py3-none-any-whl.
```

It seems like the `.whl` file name is being normalized into a PyPI name?

Be able to run

```
$ pipx install myproject-0.1.0-py3-none-any.whl
```

And it installs the package into the `pipx` environment, making the console entry points available in the PATH.

meowmeowmeowcat:

## Summary of changes

Match pip's behaviour when package name ends with an archive extension such as `.whl`, `.tar.gz` (Treat it as a local path).

Closes #641

## Test plan

Tested by running
```
pipx install black-22.6.0-py3-none-any.whl --verbose
```


meowmeowmeowcat:
> It will still change to `black-22.6.0-py3-none-any-whl` if the file doesn't exist since it may be a valid project name.

Or we should directly raise file not found error?

src/pipx/package_specifier.py
```diff
@@ -36,6 +44,16 @@ def _split_path_extras(package_spec: str) -> Tuple[str, str]:
         return (package_spec, "")
 
 
+def _check_package_path(package_path: str) -> Tuple[Path, bool]:
+    pkg_path = Path(package_path)
+    try:
+        pkg_path_exists = pkg_path.exists()
```
uranusjr, line 50: I don’t think `exists()` raise `OSError`?

meowmeowmeowcat: I moved these few lines into a new function:

https://github.com/pypa/pipx/blob/fc9919d9da647ead6ee83fa9bde4879eb13a7517/src/pipx/package_specifier.py#L75-L78

meowmeowmeowcat: But actually it only returns True or False.

meowmeowmeowcat: https://docs.python.org/3/library/pathlib.html#methods

> Changed in version 3.8: [exists()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.exists), [is_dir()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.is_dir), [is_file()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.is_file), [is_mount()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.is_mount), [is_symlink()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.is_symlink), [is_block_device()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.is_block_device), [is_char_device()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.is_char_device), [is_fifo()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.is_fifo), [is_socket()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.is_socket) now return False instead of raising an exception for paths that contain characters unrepresentable at the OS level.

meowmeowmeowcat: Looks like it is: https://docs.python.org/3/library/pathlib.html#pathlib.Path.exists

uranusjr (approved):

src/pipx/package_specifier.py
```diff
@@ -20,6 +20,14 @@
 
 logger = logging.getLogger(__name__)
 
+# Copied from https://github.com/pypa/pip/blob/main/src/pip/_internal/utils/filetypes.py
+WHEEL_EXTENSION = ".whl"
+BZ2_EXTENSIONS: Tuple[str, ...] = (".tar.bz2", ".tbz")
+XZ_EXTENSIONS: Tuple[str, ...] = (".tar.xz", ".txz", ".tlz", ".tar.lz", ".tar.lzma")
+ZIP_EXTENSIONS: Tuple[str, ...] = (".zip", WHEEL_EXTENSION)
+TAR_EXTENSIONS: Tuple[str, ...] = (".tar.gz", ".tgz", ".tar")
+ARCHIVE_EXTENSIONS = ZIP_EXTENSIONS + BZ2_EXTENSIONS + TAR_EXTENSIONS + XZ_EXTENSIONS
```
uranusjr, line 29: pip supports a ton of files for backward compatibility, but I think we should only support the standard ones: `.whl`, `.tar.gz`, and `.zip`. The wheel extension is obvious, and `packaging` also only supports the two sdist extensions:

https://github.com/pypa/packaging/blob/97db7175675a618b78838ccdf4b78108d3eca6f3/src/packaging/utils.py#L128-L131

Issue #854, opened 2022-06-18
Parse 'pipx run' input and infer the app name like 'pipx install'


aaronsteers:
**How would this feature be useful?**

This would make `pipx run` immensely more user friendly and easier to type without resorting to a readme or scanning the `--help` for the correct run syntax.

**Background**

Given an app called `tap-mysql` currently, `pipx run tap-mysql --help` works great! It knows that the plugin's name is `tap-mysql` and it attempts to run a command with that same name (`tap-mysql --help` in this example),

Given an a more complex pip_url (like `tap-mysql==1.4` or `git+https://github.com/singer-io/tap-mysql.git`), could `pipx` doesn't use the detected package name to attempt the same. Instead, the package is installed and then the command will immediately fail. 

```
$ pipx run tap-mysql --help
usage: tap-mysql [-h] -c CONFIG...
... # works perfectly
$ pipx run tap-mysql==1.17.5 --help
'tap-mysql==1.17.5' executable script not found in package 'tap-mysql==1.17.5'. Available executable scripts: tap-mysql
```

So in theory, we could put those two things together and let this work out of box?: `pipx run <any_pip_url> --help`

**Describe the solution you'd like**

The requested feature improvement would be to use the detected package name as the app name if `--spec` is not detected and if the argument passed to `pipx run` was a complex pip_url, such as a git ref or a version-pinned ref.

Already `pipx` does have the logic to detect package name, and `pipx run` already accepts simple package names without having to separately pass the app name.

```
pipx run tap-mysql --help                                              # already works today
pipx run tap-mysql==1.17.5 --help                                      # would be new
pipx run git+https://github.com/singer-io/tap-mysql@main.git --help    # would be new
```

**Describe alternatives you've considered**

The workaround is to explicitly send the executable name and pip URL separately using `pipx run --spec <pip_url> <app_name>`. However, this is not very portable and since most pipx commands do not require remembering the `--spec`.

Instead of running like this:

```
pipx run git+https://github.com/singer-io/tap-mysql.git --help
```

or

```
pipx run tap-mysql==1.0.5 --help
```

We would need to add `tap-mysql` at the end also.

```
pipx run --spec=git+https://github.com/singer-io/tap-mysql.git tap-mysql --help
```

uranusjr:
This doesn’t need an option at all, pipx could use the same name-inferring logic in `pipx install` for this.

konstantinos-null:
[...]

PR #903, opened 2022-10-19
[#854] Add logic to use pipx run as install

konstantinos-null:

## Summary of changes
https://github.com/konstantinos-null/pipx/blob/64eea87720d0b1fe2e9277a3e1e813f492678004/src/pipx/commands/run.py#L51
https://github.com/konstantinos-null/pipx/blob/64eea87720d0b1fe2e9277a3e1e813f492678004/src/pipx/commands/run.py#L108

Thought process was that app was getting the value of the spec we used so moved the same logic fro install function here to tackle the issue. I don't know if the approach is correct so feedback is welcome!

## Test plan

Tested by running
```
python3 ../source_builds/pipx/src/pipx run --verbose git+https://github.com/ansible/ansible-lint.git@v6.8.1 --version
```

### DIFF
--- a/src/pipx/commands/run.py
+++ b/src/pipx/commands/run.py
@@ -10,7 +10,7 @@ from typing import List, NoReturn
 
 from pipx import constants
 from pipx.commands.common import package_name_from_spec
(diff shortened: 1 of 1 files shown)


gaborbernat:

PR #906, opened 2022-10-28
Warn users if '#egg' is detected in package

### LINKED ISSUE #875: Show warning when a URL with #egg is provided to pipx
**Describe the bug**

Exposing the shell commands of extra dependencies does not work when installing via git+ssh. It works when installing from a local directory.

**How to reproduce**

Create a git repository `git+ssh://git@gitlab.com/.../my-package.git` with a package whose extra dependency exposes a shell command. Then, run

    pipx install  --include-deps "git+ssh://git@gitlab.com/.../my-package.git#egg=my-package[extra-dependency]"

The shell command of the `extra-dependency` is not exposed, even though the extra dependency is installed in the venv.

**Expected behavior**

The shell command should be exposed.

sha016:

## Summary of changes
Displays a warning if '#egg' is in package_spec. Closes #875 

This is my first contribution here, let me know if there's any changes that should be made.

Thanks,
## Test plan

Tested by running
```
```

### DIFF
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -182,6 +182,17 @@ def run_pipx_command(args: argparse.Namespace) -> ExitCode:  # noqa: C901
     if "skip" in args:
         skip_list = [canonicalize_name(x) for x in args.skip]
 
+    if "package_spec" in args and "#egg" in args.package_spec:
+        logger.warning(
+            pipx_wrap(
+                f"""
+                '#egg' fragment in package ({args.package_spec})
+                This is obsolete but should not impact installation.
+                """,
+                subsequent_indent=" " * 4,
+            )
+        )
+
     if args.command == "run":
         package_or_url = (
             args.spec


src/pipx/main.py
```diff
@@ -182,6 +182,17 @@ def run_pipx_command(args: argparse.Namespace) -> ExitCode:  # noqa: C901
     if "skip" in args:
         skip_list = [canonicalize_name(x) for x in args.skip]
 
+    if "package_spec" in args and "#egg" in args.package_spec:
+        logger.warning(
+            pipx_wrap(
+                f"""
+                '#egg' fragment in package ({args.package_spec})
+                This is obsolete but should not impact installation.
```
uranusjr, line 190:
```suggestion
                '#egg' fragment in package ({args.package_spec})
                is obsolete and ignored.
```

uranusjr:
Can we have a test for this?

meowmeowmeowcat:
These few lines should also be changed: https://github.com/pypa/pipx/blob/4c0f2892c8f932f345a379b84bffe816c7c7206c/src/pipx/main.py#L175-L178

chrysle:
@uranusjr I would write a test. Should I open a separate PR?

guiillermo:
Go ahead

PR #909, opened 2022-11-02
Add support for virtualenv

### LINKED ISSUE #853: use virtualenv instead of python -m venv for ubuntu installation
I don't have sudo access on ubuntu 20.04  and i try to use pipx to setup some tools for me, e.g., `pipx install poetry`.
However, I got error 
```
The virtual environment was not created successfully because ensurepip is not
available.  On Debian/Ubuntu systems, you need to install the python3-venv
package using the following command.

    apt install python3.8-venv

You may need to use sudo with that command.  After installing the python3-venv
package, recreate your virtual environment.

Failing command: ['/home/username/.local/pipx/shared/bin/python3', '-Im', 'ensurepip', '--upgrade', '--default-pip']


'/usr/bin/python3 -m venv --clear /home/username/.local/pipx/shared' failed
```

Later I figured out that i need to use `virtualenv` instead of `/usr/bin/python3 -m venv`., so i first run 
` virtualenv --clear /home/username/.local/pipx/shared`, then run `pipx install poetry`, everything worked. 

So the feature i am requesting is to add suppor to use `virtualenv` tool directly so pipx works on ubuntu without this error of `ensurepip not found`

dukecat0:

## Summary of changes

Use virtualenv to create virtual environment when venv is not available.

Closes #853

TODO:

* [ ] Allow users to pass a flag (`--use-virtualenv`) for using virtualenv to create the virtual environment.

## Test plan

Tested by running
```
pipx install black --verbose
```

### DIFF
--- a/src/pipx/shared_libs.py
+++ b/src/pipx/shared_libs.py
@@ -41,12 +41,27 @@ class _SharedLibs:
 
     def create(self, verbose: bool = False) -> None:
         if not self.is_valid:
-            with animate("creating shared libraries", not verbose):
-                create_process = run_subprocess(
-                    [DEFAULT_PYTHON, "-m", "venv", "--clear", self.root]
-                )
-            subprocess_post_check(create_process)
-
+            try:
+                with animate("creating shared libraries using venv", not verbose):
+                    create_process = run_subprocess(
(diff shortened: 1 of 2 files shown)


gaborbernat:

PR #930, opened 2023-01-22
Testing: add git to PATH

### LINKED ISSUE #303: Testing: fixture to temporarily add certain utilities to binary path

Currently, with a package that requires compilation, tests that install it fail on macOS (such as regex dependency inside of black).

Additionally, it would be helpful to add `git` and other VCS tools so that we could test installing package specifications with VCS urls.

The solution that @chrahunt mentioned (https://github.com/pipxproject/pipx/pull/298#issuecomment-568642820), quoted below:

1. Have a session-scoped fixture which creates a temporary directory
1. Populate the directory with symlinks to whitelisted build tools identified by shutil.which like gcc/clang
1. Depend on that fixture here, adding the temporary directory to PATH

Previously, we just added some paths to the bin path (see PR #269 to fix Issue #266) but this resulted in undesired side-effects and was removed (#298, #299 )

dukecat0:

## Summary of changes

Add git to PATH during testing, so we can test installing packages with VCS urls.
Closes #303 

## Test plan

Tested by running
```
nox -s tests
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -144,7 +144,7 @@ def utils_temp_dir(tmp_path_factory):
     for util in utils:
         util_path = Path(shutil.which(util))
         try:
-            Path(tmp_path / util_path.name).symlink_to(util_path)
+            (tmp_path / util_path.name).symlink_to(util_path)
         except FileExistsError:
             pass
     return tmp_path
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,3 +1,4 @@
+import ctypes
 import os
 import shutil
 import subprocess
@@ -142,7 +143,12 @@ def utils_temp_dir(tmp_path_factory):
     git_path = shutil.which("git")
     tmp_path = tmp_path_factory.mktemp("session_utilstempdir")
     try:
-        Path(tmp_path / "git").symlink_to(Path(git_path))
+        if WIN:
+            ctypes.windll.kernel32.CreateSymbolicLinkA(
+                os.path.join(tmp_path, "git"), git_path, 0
+            )
+        else:
+            Path(tmp_path / "git").symlink_to(Path(git_path))
     except FileExistsError:
         pass
     return tmp_path
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -139,14 +139,15 @@ def pipx_session_shared_dir(tmp_path_factory):
 
 @pytest.fixture(scope="session")
 def utils_temp_dir(tmp_path_factory):
-    git_path = shutil.which("git.exe" if WIN else "git")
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -139,14 +139,15 @@ def pipx_session_shared_dir(tmp_path_factory):
...
     tmp_path = tmp_path_factory.mktemp("session_utilstempdir")
-    try:
-        Path(tmp_path / "git.exe" if WIN else tmp_path / "git").symlink_to(
-            Path(git_path)
-        )
-    except FileExistsError:
-        pass
+    utils = ["git"]
+    for util in utils:
+        util_filename = f"{util}.exe" if WIN else util
```
uranusjr, line 145: Not needed, `shutil.which` do this automatically.

tests/conftest.py
```diff
@@ -59,7 +62,7 @@ def pipx_temp_env_helper(pipx_shared_dir, tmp_path, monkeypatch, request):
     #   which make tests fail (e.g. on Github ansible apps exist in /usr/bin)
     monkeypatch.setenv("PATH_ORIG", str(bin_dir) + os.pathsep + os.getenv("PATH"))
     monkeypatch.setenv("PATH_TEST", str(bin_dir))
-    monkeypatch.setenv("PATH", str(bin_dir))
+    monkeypatch.setenv("PATH", str(bin_dir) + os.pathsep + str(utils_temp_dir))
```
uranusjr, line 65:
```suggestion
    monkeypatch.setenv("PATH", str(bin_dir / utils_temp_dir))
```

tests/conftest.py
```diff
@@ -134,8 +137,23 @@ def pipx_session_shared_dir(tmp_path_factory):
...
 
 
+@pytest.fixture(scope="session")
+def utils_temp_dir(tmp_path_factory):
+    tmp_path = tmp_path_factory.mktemp("session_utilstempdir")
+    utils = ["git"]
+    for util in utils:
+        util_path = Path(shutil.which(util))
+        try:
+            Path(tmp_path / util_path.name).symlink_to(util_path)
```
uranusjr, line 147:
```suggestion
           (tmp_path / util_path.name).symlink_to(util_path)
```

tests/conftest.py
dukecat0: Seems like this will make the tests fail?

uranusjr: Why? But eh this ugly construct is also used by existing code so maybe it’s better to be consistent and do the refactoring in some other PR.

dukecat0: The `bin_dir` is the location for apps installed by `pipx install`, so these two directories should be separated.

uranusjr (approved):

PR #939, opened 2023-02-11
Added `with-suffix` option for `pipx inject` command

### LINKED ISSUE #521: Need consistent `inject --include-apps` behavior with suffix
**Describe the bug**

When injecting into a venv with a suffix using `--include-apps`, currently the injected apps do not take the venv sufix.  This seems inconsistent to me.

We should at least decide intentionally what the proper behavior should be.

**How to reproduce**

```
pipx install pycowsay --suffix _2
pipx inject pycowsay_2 black --include-apps
```
**Expected behavior**

I would assume if that we would end up with a `black_2` app to match the venv, but in fact we end up with a `black` app.

chrysle:

## Summary of changes

Added a `--with-suffix` option to `pipx inject` command to add the venv suffix to injected apps. Closes #521
Since the `pipx uninject` command doesn't work with suffixes, maybe an option should also be added there. 

## Test plan

Tested by running
```
python3 src/pipx install pycowsay --suffix "_2" 
python3 src/pipx inject pycowsay_2 black --include-apps --with-suffix
nox -s tests-3.8
```


chrysle:
@meowmeowmeowcat Could you review this? I also don't exactly know what is wrong with the linter.

chrysle:
@uranusjr Could you have a look at this please?

PR #953, opened 2023-03-16
Added `--upgrade` option for `pipx install`

### LINKED ISSUE #741: Add unified install/upgrade command

When you install/upgrade pip, it has a unified command line:

```
python3 -m pip install --user -U foo
```

But for `pipx`, need to have different command, depending on if you want to upgrade or install:

```
pipx install foo
pipx upgrade foo
```

For the end user instructions, it would be convenient to have a single command that does install or upgrade to the latest version

```
pipx install foo --upgrade
pipx install foo -U
```

Alternative instructions having two commands, is not that straight forward:

Consider this in the application readme:

----

To install latest application foo, run the following:

```sh
pipx install foo

# or

pipx upgrade foo
```

chrysle:

## Summary of changes

Fixes #741 

## Test plan

Tested by running
```
pipx install black
pipx install black --upgrade # -U also works
nox -s tests-3.8
```

### DIFF
--- a/src/pipx/commands/install.py
+++ b/src/pipx/commands/install.py
@@ -3,6 +3,7 @@ from typing import List, Optional
 
 from pipx import constants
 from pipx.commands.common import package_name_from_spec, run_post_install_actions
+from pipx.commands.upgrade import upgrade as upgrade_package
 from pipx.constants import EXIT_CODE_INSTALL_VENV_EXISTS, EXIT_CODE_OK, ExitCode
 from pipx.util import pipx_wrap
 from pipx.venv import Venv, VenvContainer
@@ -21,6 +22,7 @@ def install(
     force: bool,
     include_dependencies: bool,
     suffix: str = "",
+    upgrade: bool,
 ) -> ExitCode:
     """Returns pipx exit code."""
     # package_spec is anything pip-installable, including package_name, vcs spec,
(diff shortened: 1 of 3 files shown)


src/pipx/commands/install.py
```diff
@@ -40,44 +42,55 @@ def install(
...
-        run_post_install_actions(
-            venv,
-            package_name,
-            local_bin_dir,
-            venv_dir,
-            include_dependencies,
             force=force,
+            include_injected=True,
+            verbose=verbose,
         )
```
meowmeowmeowcat, line 53: We should probably keep installing the package even if it is not installed. (Currently it will exit with an error when the package is not installed.)

meowmeowmeowcat (changes requested):

src/pipx/commands/install.py
chrysle: The change is placed. I had to add `-f` to the test however because the assert won't work with existing installations.

chrysle: I think `--force` should automatically be passed when upgrading?

meowmeowmeowcat: Yeah, I agree with this.

chrysle: Done.

chrysle:
@meowmeowmeowcat Do you have any other suggestions or is this production-ready?

meowmeowmeowcat (commented):
LGTM, but I am not sure if other maintainers have any other opinions on this.

uranusjr:
Personally I don’t feel this is needed. The reasoning of the linked issue is basically “pip has this” and IMO is very weak. But if others feel strong enough I won’t block the feature.

chrysle:
> The reasoning of the linked issue is basically “pip has this” and IMO is very weak. 

I think most people working with pipx already know how to handle pip. Won't it help them if they find pipx behaving similarly to what they're used from pip? 

In my opinion the option would be a good addition to the (rather unusual?) `upgrade` command.

uranusjr:
> rather unusual?

If you look across all package managers, I believe `[manager] upgrade` is more popular than `[manager] [add] --upgrade`. At most 50-50.

chrysle:
> If you look across all package managers, I believe `[manager] upgrade` is more popular than `[manager] [add] --upgrade`. At most 50-50.

Even if this is the case, it's still quite common. I would think there is nothing that spoke against a little usability improvement? Since this isn't anything that aims to replace the `upgrade` command.

chrysle:
@uranusjr Could this be merged or closed soon? Otherwise there'll be a conflict with my work on #971

PR #957, opened 2023-03-18
fix: same interpreter from venv to get pkg name

### LINKED ISSUE #956: inject command uses wrong python interpreter for local packages
**Describe the bug**
While trying to `pipx inject` a local wheel with specific Python version constraints into a compatible environment (i.e. with a valid interpreter for those constraints), I keep getting the following error:

```
ERROR: Package 'my-local-package' requires a different Python: 3.11.1 not in '<3.11,>=3.8'
```
Here are the relevant versions and constraints 👇 

- **package (to be injected) interpreter constraints**:  <3.11, >=3.8
- **environment interpreter**: 3.8
- **pipx interpreter**: 3.11.1

After seeing the error, it became clear that somehow in the middle of the injection process, pipx was trying to install the package with a different interpreter for some reason. After some debugging, I noticed that when the package is a local distribution we need to install it (without its dependencies) into a temporary venv in order to extract its name. However, when we do it, we end up using a default Python interpreter, instead of the one associated with the venv here 👇 

https://github.com/pypa/pipx/blob/726eaf88077b65069fc7143f72bd2a1153ac35aa/src/pipx/commands/inject.py#L46-L51

I locally replaced `venv.python` (which maps to a default interpreter in this case) with `venv.python_path` (which maps to venv-specific interpreter) and it worked 😃 

**How to reproduce**

Verbose logs for the inject command 

```
pipx >(_parsed_package_to_package_or_url:128): cleaned package spec: /path/to/my-local-package-1.0.0-py3-none-any.whl
pipx >(run_subprocess:173): running /opt/homebrew/Cellar/pipx/1.1.0/libexec/bin/python3.11 -m venv --without-pip /var/folders/vj/chbx1tms2c93nmnnp2gw4wh80000gq/T/tmpe5g_foll
pipx >(run_subprocess:173): running /var/folders/vj/chbx1tms2c93nmnnp2gw4wh80000gq/T/tmpe5g_foll/bin/python -c import sysconfig; print(sysconfig.get_path('purelib'))
pipx >(run_subprocess:173): running /Users/me/.local/pipx/shared/bin/python -c import sysconfig; print(sysconfig.get_path('purelib'))
pipx >(run_subprocess:173): running /var/folders/vj/chbx1tms2c93nmnnp2gw4wh80000gq/T/tmpe5g_foll/bin/python --version
determining package name from '/path/to/my-local-package-1.0.0-py3-none-any.whl'...
pipx >(run_subprocess:173): running /var/folders/vj/chbx1tms2c93nmnnp2gw4wh80000gq/T/tmpe5g_foll/bin/python -m pip list --format=json
pipx >(run_subprocess:173): running /var/folders/vj/chbx1tms2c93nmnnp2gw4wh80000gq/T/tmpe5g_foll/bin/python -m pip install --no-dependencies /path/to/my-local-package-1.0.0-py3-none-any.whl
Looking in indexes: https://pypi.org/simple
Processing /path/to/my-local-package-1.0.0-py3-none-any.whl
ERROR: Package 'my-local-package' requires a different Python: 3.11.1 not in '<3.11,>=3.8'
pipx >(subprocess_post_check:207): '/var/folders/vj/chbx1tms2c93nmnnp2gw4wh80000gq/T/tmpe5g_foll/bin/python -m pip install --no-dependencies /path/to/my-local-package-1.0.0-py3-none-any.whl' failed
Cannot determine package name from spec '/path/to/my-local-package-1.0.0-py3-none-any.whl'. Check package spec for errors.
```

**Expected behavior**

pipx should be able to inject dependencies into existing environments if their interpreters are compatible.

lecardozo:

## Summary of changes
Fixes #956 
- Use venv-specific interpreter for package name extraction when injecting local "pip-installable" packages

## Test plan

Tested by running the following locally
```shell
pipx install jupyterlab --python python3.8

# The package must be any wheel with interpreter constraints
# that are compatible with the venv's in which it's going to be injected
# but doesn't match pipx's default interpreter
pipx inject jupyterlab package-1.0.0-py3-none-any.whl
```

### DIFF
--- a/src/pipx/commands/inject.py
+++ b/src/pipx/commands/inject.py
@@ -1,3 +1,4 @@
+import os
 import sys
 from pathlib import Path
 from typing import List, Optional
@@ -47,7 +48,10 @@ def inject_dep(
     #   zip file, or tar.gz file.
     if package_name is None:
         package_name = package_name_from_spec(
-            package_spec, str(venv.python_path), pip_args=pip_args, verbose=verbose
+            package_spec,
+            os.fspath(venv.python_path),
+            pip_args=pip_args,
+            verbose=verbose,
         )
 
     venv.install_package(
--- a/src/pipx/commands/inject.py
+++ b/src/pipx/commands/inject.py
@@ -47,7 +47,7 @@ def inject_dep(
     #   zip file, or tar.gz file.
     if package_name is None:
         package_name = package_name_from_spec(
-            package_spec, venv.python_path, pip_args=pip_args, verbose=verbose
+            package_spec, str(venv.python_path), pip_args=pip_args, verbose=verbose
         )
 
     venv.install_package(
--- a/src/pipx/commands/inject.py
+++ b/src/pipx/commands/inject.py
@@ -47,7 +47,7 @@ def inject_dep(
     #   zip file, or tar.gz file.
     if package_name is None:
         package_name = package_name_from_spec(
-            package_spec, venv.python, pip_args=pip_args, verbose=verbose
+            package_spec, venv.python_path, pip_args=pip_args, verbose=verbose
         )
 
     venv.install_package(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## dev
 
 - Fallback to user's log path if the default log path (`$PIPX_HOME/logs`) is not writable to aid with pipx being used for multi-user (e.g. system-wide) installs of applications
+- Fix wrong interpreter usage when injecting local pip-installable dependencies into venvs
 
 ## 1.2.0
 


src/pipx/commands/inject.py
```diff
@@ -47,7 +47,7 @@ def inject_dep(
     #   zip file, or tar.gz file.
     if package_name is None:
         package_name = package_name_from_spec(
-            package_spec, venv.python, pip_args=pip_args, verbose=verbose
+            package_spec, venv.python_path, pip_args=pip_args, verbose=verbose
```
chrysle, line 50: `venv.python_path` seems to be of type `Path`, which breaks the lint because `package_name_from_spec` requires `str`. You should convert `Path` or add the possibility to pass it in the function itself.

lecardozo: Thanks for the heads up, @chrysle! Just fixed it 😃

src/pipx/commands/inject.py
```diff
@@ -47,7 +47,7 @@ def inject_dep(
     #   zip file, or tar.gz file.
     if package_name is None:
         package_name = package_name_from_spec(
-            package_spec, venv.python, pip_args=pip_args, verbose=verbose
+            package_spec, str(venv.python_path), pip_args=pip_args, verbose=verbose
```
uranusjr, line 50: Better to use `os.fspath` instead

lecardozo: done 😃

uranusjr (approved):

PR #960, opened 2023-03-24
Update examples.md

### LINKED ISSUE #941: Pip args do not behave similarly for `pipx install` and `pipx run`
**Describe the bug**
When installing a package with `pipx install`, which requires packages from both the public registry and a private one, it works:
`pipx install <package> --pip-args="--index-url=<private_url> --extra-index-url https://pypi.python.org/simple"`

When trying to run the same package using `pipx run`, it does not:
`pipx run <package> --pip-args="--index-url=<private_url> --extra-index-url https://pypi.python.org/simple"`

Result:
```bash
    ERROR: Could not find a version that satisfies the requirement <private_package> (from versions: none)
    ERROR: No matching distribution found for <private_package>
Error installing <package> from spec '<local_wheel>'.
```

Should i supply arguments in a different way here?

meowmeowmeowcat:

## Summary of changes
Closes #941
Update a few examples in `examples.md`.

## Test plan

Tested by running
```
```

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -7,13 +7,14 @@ pipx install --python python3.8 pycowsay
 pipx install git+https://github.com/psf/black
 pipx install git+https://github.com/psf/black.git@branch-name
 pipx install git+https://github.com/psf/black.git@git-hash
+pipx install git+ssh://<username>@<private-repo-domain>/<path-to>/<package.git>
 pipx install https://github.com/psf/black/archive/18.9b0.zip
 pipx install black[d]
 pipx install 'black[d] @ git+https://github.com/psf/black.git@branch-name'
 pipx install --suffix @branch-name 'black[d] @ git+https://github.com/psf/black.git@branch-name'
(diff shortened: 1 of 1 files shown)


docs/examples.md
```diff
@@ -7,13 +7,14 @@ pipx install --python python3.8 pycowsay
 pipx install git+https://github.com/psf/black
 pipx install git+https://github.com/psf/black.git@branch-name
 pipx install git+https://github.com/psf/black.git@git-hash
+pipx install git+ssh://<username>@<private-repo-domain>/<path-to>/<package.git>
```
uranusjr, line 10:
```suggestion
pipx install git+ssh://<username>@<private-repo-domain>/<path-to-package.git>
```

I don’t think there’s really much reason to split these?

NeffIsBack: The reason why I split it that I was not sure if it is clear that you have to include the `package.git` in the path. But do what you think is best.

PR #970, opened 2023-04-16
Override shared libs directory when installing/running pip/wheel

### LINKED ISSUE #951: `pipx run wheel` / `install wheel` doesn't work
**Describe the bug**

`pipx run wheel` doesn't seem to be supported. I think it's because `wheel` is filtered out as part of the default environment.

~~A possible fix would be to add the entrypoint for pipx to wheel;~~ otherwise I think special handling would have to be included to see if 'wheel' is specifically asked for.

**How to reproduce**
`pipx run wheel` reports no executables found.

```console
$ docker run --rm -it python bash
# pip uninstall wheel
# pip install pipx
# pipx run wheel
'wheel' executable script not found in package 'wheel'.
Available executable scripts:

```

**Expected behavior**
It should get wheel.

I opened an example of the entrypoint in wheel, but I realized that `pipx install wheel` should work too, and it doesn't:

```console
# pipx install wheel

No apps associated with package wheel or its dependencies. If you are attempting to install a library, pipx should not be used. Consider using pip or a similar tool
instead.
```

So the fix needs to be here. Note that wheel isn't meant to be used as a library (at least yet), only as a CLI app (or a setuptools helper, of course). And the next version will be gaining a very useful CLI command (`tags`).

It also seems it's filtered from git?

```console
$ pipx run --spec git+https://github.com/henryiii/wheel.git@henryiii/fix/pipxrun wheel tag --help

Cannot determine package name from spec 'git+https://github.com/henryiii/wheel.git@henryiii/fix/pipxrun'. Check package spec for errors.
```

That might be related to the fact the version number is the same, and it doesn't reinstall the package, though.

chrysle:

## Summary of changes

Fixes #951 

## Test plan

Tested by running
```
python src/pipx run wheel
python src/pipx install wheel
nox -s tests-3.8
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## dev
 
+- Ignore shared libs directory on installing/running pip/wheel
 - Support including requirements in scripts run using `pipx run` (#916)
 - Pass `pip_args` to `shared_libs.upgrade()`
 - Fallback to user's log path if the default log path (`$PIPX_HOME/logs`) is not writable to aid with pipx being used for multi-user (e.g. system-wide) installs of applications
--- a/src/pipx/commands/install.py
+++ b/src/pipx/commands/install.py
@@ -30,6 +30,12 @@ def install(
         package_name = package_name_from_spec(
             package_spec, python, pip_args=pip_args, verbose=verbose
         )
+
+    override_shared = False
+
+    if package_name in ["pip", "wheel"]:
+        override_shared = True
+
     if venv_dir is None:
         venv_container = VenvContainer(constants.PIPX_LOCAL_VENVS)
         venv_dir = venv_container.get_venv_dir(f"{package_name}{suffix}")
@@ -56,7 +62,7 @@ def install(
             return EXIT_CODE_INSTALL_VENV_EXISTS
 
     try:
-        venv.create_venv(venv_args, pip_args)
+        venv.create_venv(venv_args, pip_args, override_shared)
         venv.install_package(
             package_name=package_name,
             package_or_url=package_spec,
--- a/src/pipx/venv.py
+++ b/src/pipx/venv.py
@@ -156,24 +156,26 @@ class Venv:
         else:
             return self.pipx_metadata.main_package.package
 
-    def create_venv(self, venv_args: List[str], pip_args: List[str]) -> None:
+    def create_venv(self, venv_args: List[str], pip_args: List[str], override_shared: bool = False) -> None:
         with animate("creating virtual environment", self.do_animation):
-            cmd = [self.python, "-m", "venv", "--without-pip"]
+            cmd = [self.python, "-m", "venv"]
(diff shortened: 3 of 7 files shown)


src/pipx/commands/install.py
```diff
@@ -30,6 +30,12 @@ def install(
         package_name = package_name_from_spec(
             package_spec, python, pip_args=pip_args, verbose=verbose
         )
+
+    override_shared = False
+
+    if package_name in ["pip", "wheel"]:
+        override_shared = True
```
uranusjr, line 37: `pipx install pip` does not really make practical sense IMO so I’d recommend only handle `wheel` until someone proposes a use case.

chrysle: Thanks for the review. Updated.

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## dev
 
+- Ignore shared libs directory on installing/running pip/wheel
```
uranusjr, line 3:
```suggestion
- Ignore shared libs directory on installing/running `wheel`
```

src/pipx/venv_inspect.py
```diff
@@ -69,6 +69,7 @@ def get_apps(dist: metadata.Distribution, bin_path: Path) -> List[str]:
 
     sections = {"console_scripts", "gui_scripts"}
     # "entry_points" entry in setup.py are found here
+    # print([x for x in bin_path.iterdir()])
```
uranusjr, line 72: Accidentally left over?

chrysle: Yes, thanks for catching!

meowmeowmeowcat (approved):

meowmeowmeowcat:
Thanks!

src/pipx/venv.py
```diff
@@ -156,24 +156,28 @@ def main_package_name(self) -> str:
         else:
             return self.pipx_metadata.main_package.package
 
-    def create_venv(self, venv_args: List[str], pip_args: List[str]) -> None:
+    def create_venv(
+        self, venv_args: List[str], pip_args: List[str], override_shared: bool = False
+    ) -> None:
```
uranusjr, line 161: Can we have a docstring describing `override_shared`?

chrysle: Added.

graingert:
since 23.1 pip doesn't need wheel anymore https://pip.pypa.io/en/stable/news/#v23-1

> When the wheel package is not installed, pip now uses the default build backend instead of setup.py install and setup.py develop for project without pyproject.toml. ([#8559](https://github.com/pypa/pip/issues/8559))

so probably the fix is to remove wheel from the shared libraries

gaborbernat (approved):

gaborbernat (changes requested):
Can you fix the merge conflicts?

PR #972, opened 2023-04-23
📝 Update installation method for linux users

oviniciusfeitosa:

## Test plan

Tested by running
```
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,8 +38,8 @@ Upgrade pipx with `brew update && brew upgrade pipx`.
 ### On Linux, install via pip (requires pip 19.0 or later)
 
 ```
-python3 -m pip install --user pipx
-python3 -m pipx ensurepath
+sudo apt install pipx
+pipx ensurepath
 ```
 
 Upgrade pipx with `python3 -m pip install --user --upgrade pipx`.
--- a/README.md
+++ b/README.md
@@ -35,7 +35,6 @@ pipx ensurepath
 
 Upgrade pipx with `brew update && brew upgrade pipx`.
 
-
 ### On Linux
 
 -  Ubuntu 23.04 or above
--- a/README.md
+++ b/README.md
@@ -36,18 +36,19 @@ pipx ensurepath
 Upgrade pipx with `brew update && brew upgrade pipx`.
 
 
-### On Linux, install via pip (requires pip 19.0 or later)
+### On Linux
 
 -  Ubuntu 23.04 or above
 
-```sh
+```
+sudo apt update
 sudo apt install pipx
 pipx ensurepath
 ```
 
 -  Ubuntu 22.04 or below
 
-```sh
+```
 python3 -m pip install --user pipx
 python3 -m pipx ensurepath
 ```
--- a/README.md
+++ b/README.md
@@ -35,11 +35,21 @@ pipx ensurepath
 
 Upgrade pipx with `brew update && brew upgrade pipx`.
 
+
 ### On Linux, install via pip (requires pip 19.0 or later)
 
-```
+-  Ubuntu 23.04 or above
+
+```sh
 sudo apt install pipx
 pipx ensurepath
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,9 +35,19 @@ pipx ensurepath
 
 Upgrade pipx with `brew update && brew upgrade pipx`.
 
+
 ### On Linux, install via pip (requires pip 19.0 or later)
```
chrysle, line 39: This should be changed

README.md
```diff
@@ -35,9 +35,19 @@ pipx ensurepath
...
 
+-  Ubuntu 23.04 or above
+
+```sh
+sudo apt install pipx
```
chrysle, line 44:
```suggestion
sudo apt update
sudo apt install pipx
```

README.md
```diff
@@ -35,9 +35,19 @@ pipx ensurepath
 
 Upgrade pipx with `brew update && brew upgrade pipx`.
 
+
 ### On Linux, install via pip (requires pip 19.0 or later)
 
+-  Ubuntu 23.04 or above
+
+```sh
```
chrysle, line 43: I think you should adapt your style to the rest of the docs and remove the `sh`.

README.md
```diff
@@ -35,9 +35,19 @@ pipx ensurepath
...
+sudo apt install pipx
+pipx ensurepath
 ```
+
+-  Ubuntu 22.04 or below
```
chrysle, line 48:
```suggestion
Or, and for other distributions, install via pip:
```
Installing via `pip` is still recommended, since new versions of `pipx` will be directly uploaded to PyPi and be pip-installable, in contrast to the version freeze in Ubuntu.
I'd tend to remove this line.

README.md
```diff
@@ -35,9 +35,19 @@ pipx ensurepath
 
 Upgrade pipx with `brew update && brew upgrade pipx`.
 
+
 ### On Linux, install via pip (requires pip 19.0 or later)
 
+-  Ubuntu 23.04 or above
```
chrysle, line 41:
```suggestion
Ubuntu 23.04 and above provide a packaged version of pipx:
```

chrysle:
Gentle ping @oviniciusfeitosa

README.md
dukecat0: See #982. Installing via pip may not work. Or maybe we can suggest users use pipx through zipapp?
```
curl -L https://github.com/pypa/pipx/releases/latest/download/pipx.pyz -o pipx.pyz
alias pipx="/path/to/python/interpreter pipx.pyz"
```

README.md
```diff
@@ -35,7 +35,18 @@ pipx ensurepath
 
 Upgrade pipx with `brew update && brew upgrade pipx`.
 
-### On Linux, install via pip (requires pip 19.0 or later)
+
```
dukecat0, line 38:
```suggestion

```

README.md
chrysle: Is this better?

README.md
uranusjr: Sueggesting the pyz is probably the best thing to do. Instead of `alias` (which needs to be added to rc) it’s probably better to suggest adding a script fix somewhere in PATH that does `exec ..../python ..../pipx.pyz`

chrysle: Where should the zipapp be stored? And it must be updated by hand, this should probably be included in the script fix. 

Also, I don't think it should be suggested to tweak the PATH (if that is what you want).

uranusjr: It can be added to `~/.local/bin`. `pipx ensurepath` would add that to PATH.

chrysle: Do you think of a script approach like in #849 ?

dukecat0: Maybe we can merge this PR first, and leave all these works in the follow-up PRs? :)

koterpillar: FYI, `pipx.pyz` contains a shellbang header so it can be made executable directly.

PR #1001, opened 2023-06-15
Confirm XDG base directory specification

### LINKED ISSUE #722: XDG compatibility
`pipx` installs stores all it's files to `~/.local/pipx`, but that's not what [XDG spec says][xdg-spec].

These should be probably changed to match:

- `~/.local/pipx/logs` -> `$XDG_STATE_HOME/pipx/logs`
- `~/.local/pipx/shared` -> `$XDG_DATA_HOME/pipx/shared`
- `~/.local/pipx/venvs` -> `$XDG_DATA_HOME/pipx/venvs`

venvs could be `$XDG_STATE_HOME/pipx/venvs`, but I looked virtualenvs are stored in `~/.local/share/virtualenvs`.

when implementing, then for safe migration purposes, `pipx` should use old paths if it finds the parent directory (`~/.local/pipx`) present, otherwise use the new standards confirming paths.

[xdg-spec]: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

memchr:

## Summary of changes

Changed default data and log directories specified by the [XDG spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) using [platformdirs](https://pypi.org/project/platformdirs/).

- `~/.local/pipx/logs` -> `$XDG_STATE_HOME/pipx/logs`
- `~/.local/pipx/shared` -> `$XDG_DATA_HOME/pipx/shared`
- `~/.local/pipx/venvs` -> `$XDG_DATA_HOME/pipx/venvs`
- `~/.local/pipx/.trash` -> `$XDG_DATA_HOME/pipx/trash`
- `~/.local/pipx/.cache` -> `$XDG_CACHE_HOME/pipx`

If `~/.local/pipx` exists, revert to old behavior.

Fixes #722

## Test plan

Tested by running
```
# platform linux - python 3.11.3, pytest 7.3.2, pluggy-1.0.0
python -m venv venv
. venv/bin/activate
nox -s tests-3.11
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Fallback to user's log path if the default log path (`$PIPX_HOME/logs`) is not writable to aid with pipx being used for multi-user (e.g. system-wide) installs of applications
 - Fix wrong interpreter usage when injecting local pip-installable dependencies into venvs
 - add pre-commit hook support
+- Confirm the XDG base directory specification
 
 ## 1.2.0
 
--- a/docs/how-pipx-works.md
+++ b/docs/how-pipx-works.md
@@ -2,17 +2,17 @@
 
 When installing a package and its binaries (`pipx install package`) pipx will
 
-- create directory `~/.local/pipx/venvs/PACKAGE`
-- create or re-use a shared virtual environment that contains shared packaging libraries `pip`, `setuptools` and `wheel` in `~/.local/pipx/shared/`
+- create directory `~/.local/share/pipx/venvs/PACKAGE`
+- create or re-use a shared virtual environment that contains shared packaging libraries `pip`, `setuptools` and `wheel` in `~/.local/share/pipx/shared/`
 - ensure all packaging libraries are updated to their latest versions
-- create a Virtual Environment in `~/.local/pipx/venvs/PACKAGE` that uses the shared pip mentioned above but otherwise is isolated (pipx uses a [.pth file]( https://docs.python.org/3/library/site.html) to do this)
+- create a Virtual Environment in `~/.local/share/pipx/venvs/PACKAGE` that uses the shared pip mentioned above but otherwise is isolated (pipx uses a [.pth file]( https://docs.python.org/3/library/site.html) to do this)
 - install the desired package in the Virtual Environment
-- expose binaries at `~/.local/bin` that point to new binaries in `~/.local/pipx/venvs/PACKAGE/bin` (such as `~/.local/bin/black` -> `~/.local/pipx/venvs/black/bin/black`)
+- expose binaries at `~/.local/bin` that point to new binaries in `~/.local/share/pipx/venvs/PACKAGE/bin` (such as `~/.local/bin/black` -> `~/.local/share/pipx/venvs/black/bin/black`)
 - As long as `~/.local/bin/` is on your PATH, you can now invoke the new binaries globally
 
 When running a binary (`pipx run BINARY`), pipx will
 
-- create or re-use a shared virtual environment that contains shared packaging libraries `pip`, `setuptools` and `wheel` in `~/.local/pipx/shared/`
+- create or re-use a shared virtual environment that contains shared packaging libraries `pip`, `setuptools` and `wheel` in `~/.local/share/pipx/shared/`
 - ensure all packaging libraries are updated to their latest versions
 - create a temporary directory (or reuse a cached virtual environment for this package) with a name based on a hash of the attributes that make the run reproducible. This includes things like the package name, spec, python version, and pip arguments.
 - create a Virtual Environment inside it with `python -m venv`
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -62,7 +62,7 @@ Example configuration for use of the code linter [yapf](https://github.com/googl
 
 The default binary location for pipx-installed apps is `~/.local/bin`. This can be overridden with the environment variable `PIPX_BIN_DIR`.
 
-pipx's default virtual environment location is `~/.local/pipx`. This can be overridden with the environment variable `PIPX_HOME`.
+pipx's default virtual environment location is `$XDG_DATA_HOME/pipx`, which is typically `~/.local/share/pipx`, and for compatibility reasons, if `~/.local/pipx` exists, it will be used as the default location instead. This can be overridden with the `PIPX_HOME` environment variable.
 
 As an example, you can install global apps accessible by all users on your system with the following command (on MacOS, Linux, and Windows WSL):
 
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -74,10 +74,10 @@ Reference: [pip Environment Variables](https://pip.pypa.io/en/stable/user_guide/
 
 ## `pipx` log files
 Pipx records a verbose log file for every `pipx` command invocation.  The logs
-for the last 10 `pipx` commands can be found in `$PIPX_HOME/logs` or user's log path
+for the last 10 `pipx` commands can be found in `$XDG_STATE_HOME/pipx/logs` or user's log path
 if the former is not writable by the user.
 
-For most users this location is `~/.local/pipx/logs`, where `~` is your home
+For most users this location is `~/.local/state/pipx/logs`, where `~` is your home
 directory.
 
 ## Debian, Ubuntu issues
--- a/src/pipx/commands/environment.py
+++ b/src/pipx/commands/environment.py
@@ -25,7 +25,7 @@ def environment(value: str) -> ExitCode:
     }
     if value is None:
         for env_variable in environment_variables:
-            print(f"{env_variable}={environment_variables[env_variable]}")
+            print(f"{env_variable}={environment_variables[env_variable] or ''} ")
         print("")
         print("Only PIPX_HOME and PIPX_BIN_DIR can be set by users in the above list.")
     elif value in environment_variables:
--- a/src/pipx/constants.py
+++ b/src/pipx/constants.py
@@ -4,20 +4,31 @@ import sysconfig
 from pathlib import Path
 from textwrap import dedent
 from typing import NewType, Optional
+from platformdirs import user_cache_path, user_data_path, user_log_path
 
 DEFAULT_PIPX_HOME = Path.home() / ".local/pipx"
 DEFAULT_PIPX_BIN_DIR = Path.home() / ".local/bin"
-PIPX_HOME = Path(os.environ.get("PIPX_HOME", DEFAULT_PIPX_HOME)).resolve()
-PIPX_LOCAL_VENVS = PIPX_HOME / "venvs"
-PIPX_LOG_DIR = PIPX_HOME / "logs"
-DEFAULT_PIPX_SHARED_LIBS = PIPX_HOME / "shared"
-PIPX_TRASH_DIR = PIPX_HOME / ".trash"
+
+if DEFAULT_PIPX_HOME.exists() or os.environ.get("PIPX_HOME") is not None:
+    PIPX_HOME = Path(os.environ.get("PIPX_HOME", DEFAULT_PIPX_HOME)).resolve()
+    PIPX_LOCAL_VENVS = PIPX_HOME / "venvs"
+    PIPX_LOG_DIR = PIPX_HOME / "logs"
+    DEFAULT_PIPX_SHARED_LIBS = PIPX_HOME / "shared"
+    PIPX_TRASH_DIR = PIPX_HOME / ".trash"
+    PIPX_VENV_CACHEDIR = PIPX_HOME / ".cache"
+else:
+    PIPX_HOME = user_data_path("pipx")
+    PIPX_LOCAL_VENVS = PIPX_HOME / "venvs"
+    PIPX_LOG_DIR = user_log_path("pipx")
+    DEFAULT_PIPX_SHARED_LIBS = PIPX_HOME / "shared"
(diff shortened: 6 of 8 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - Fallback to user's log path if the default log path (`$PIPX_HOME/logs`) is not writable to aid with pipx being used for multi-user (e.g. system-wide) installs of applications
 - Fix wrong interpreter usage when injecting local pip-installable dependencies into venvs
 - add pre-commit hook support
+- Confirm the XDG base directory specification
```
chrysle, line 8: This should be somewhat more descriptive:

```suggestion
- Move `pipx` paths to ensure compatibility with the XDG base directory specification
```

src/pipx/commands/environment.py
```diff
@@ -25,7 +25,7 @@ def environment(value: str) -> ExitCode:
     }
     if value is None:
         for env_variable in environment_variables:
-            print(f"{env_variable}={environment_variables[env_variable]}")
+            print(f"{env_variable}={environment_variables[env_variable] or ''} ")
```
chrysle, line 28: Why did you add this?

src/pipx/constants.py
```diff
@@ -4,20 +4,31 @@
 from pathlib import Path
 from textwrap import dedent
 from typing import NewType, Optional
+from platformdirs import user_cache_path, user_data_path, user_log_path
```
chrysle, line 7: I'm sure this will break the style guide.. Can you run `isort` or another tool on that?

src/pipx/constants.py
```diff
@@ -4,20 +4,31 @@
...
 
 DEFAULT_PIPX_HOME = Path.home() / ".local/pipx"
 DEFAULT_PIPX_BIN_DIR = Path.home() / ".local/bin"
```
chrysle, line 10: This isn't the default any more, is it? Can you change the fallback mechanism?

src/pipx/constants.py
memchr: Both isort and black just add a new line before line 7.

src/pipx/commands/environment.py
memchr: Remnants of older iteration

src/pipx/constants.py
memchr: What do you suggest? Change it to `user_data_path("pipx")` or remove this constant?

chrysle: > Change it to `user_data_path("pipx")`

That would be my suggestion.

src/pipx/constants.py
```diff
@@ -5,19 +5,32 @@
...
+
+if FALLBACK_PIPX_HOME.exists() or os.environ.get("PIPX_HOME") is not None:
+    PIPX_HOME = Path(os.environ.get("PIPX_HOME", FALLBACK_PIPX_HOME)).resolve()
+    PIPX_LOCAL_VENVS = PIPX_HOME / "venvs"
+    PIPX_LOG_DIR = PIPX_HOME / "logs"
+    DEFAULT_PIPX_SHARED_LIBS = PIPX_HOME / "shared"
+    PIPX_TRASH_DIR = PIPX_HOME / ".trash"
+    PIPX_VENV_CACHEDIR = PIPX_HOME / ".cache"
+else:
+    PIPX_HOME = user_data_path("pipx")
```
chrysle, line 22: I suggest doing it like this now:

```suggestion
    PIPX_HOME = DEFAULT_PIPX_HOME
```

chrysle (approved):
LGTM. @uranusjr could you take a look?

uranusjr:
How does this affect existing installations? I feel we should add something in the documentation for this. Even if existing installations are not affected at all (i.e. will continue to use the old paths), we should add some notes to guide migration if users want to use the new paths.

uranusjr:
Since paths in docs have been changed to the new locations, maybe we can frame this as a potential question a user may have, e.g. files are not at the expected locations. An entry in _Troubleshooting_ can explain why that is the case, it should not affect functionalities, provide a table to map old and new paths, and explain how to migrate if the user inclines (I assume by exporting and reinstalling all apps?)

memchr:
> I assume by exporting and reinstalling all apps?

Would it be better to add a new command/flag to handle the path change, or should it be done manually?

chrysle:
> Would it be better to add a new command/flag to handle the path change

I think that's not worth it... Most users will upgrade their `pipx` installations to latest anyway, I suppose. I suggest mentioning the version which introduces this in the note. For the minority who may not want to upgrade `pipx` because of various reasons, the maual way should be proposed (if even wanted).

docs/installation.md
```diff
@@ -71,6 +71,18 @@ sudo PIPX_HOME=/opt/pipx PIPX_BIN_DIR=/usr/local/bin pipx install PACKAGE
 # Example: $ sudo PIPX_HOME=/opt/pipx PIPX_BIN_DIR=/usr/local/bin pipx install cowsay
 ```
 
+!!! note
+
+    After version 1.2.0, the default pipx paths have been moved from `~/.local/pipx` to `$XDG_DATA_HOME/pipx`, which is typically `~/.local/share/pipx`, to ensure compatibility with the XDG base directory specification.
```
chrysle, line 76: I'd find a link quite useful.

```suggestion
    After version 1.2.0, the default pipx paths have been moved from `~/.local/pipx` to `$XDG_DATA_HOME/pipx`, which is typically `~/.local/share/pipx`, to ensure compatibility with the [XDG base directory specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#variables).
```

docs/troubleshooting.md
```diff
@@ -119,3 +119,21 @@ To clean up after this experiment:
...
+
+The default PIPX_HOME is `~/.local/pipx`, prior to the adoption of the XDG base
+directory specification after version 1.2.0. To maintain compatibility with older 
+versions, pipx will automatically detect the old paths and use them accordingly.
+For a map of old and new paths, See [Installation](installation.md#installation-options)
+
+To migrate from the old path to the new path, you can remove the `~/.local/pipx` directory and 
+reinstall all packages.
+
+for example:
```
chrysle, line 133:
```suggestion
For example, on Linux systems, you could read out `pipx`'s package information in JSON via `jq` (which you might need to install first):
```

docs/installation.md
```diff
@@ -71,6 +71,18 @@ sudo PIPX_HOME=/opt/pipx PIPX_BIN_DIR=/usr/local/bin pipx install PACKAGE
 # Example: $ sudo PIPX_HOME=/opt/pipx PIPX_BIN_DIR=/usr/local/bin pipx install cowsay
 ```
 
+!!! note
+
+    After version 1.2.0, the default pipx paths have been moved from `~/.local/pipx` to `$XDG_DATA_HOME/pipx`, which is typically `~/.local/share/pipx`, to ensure compatibility with the [XDG base directory specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#variables).
```
uranusjr, line 76: How does Windows apply here?

memchr: I can't test on Windows. Where does pipx store its files on Windows?

uranusjr: Worthwhile to ask `platformdirs`. Or maybe they have this in the documentation already.

uranusjr: Come to think of it Macs are also an issue. Maybe instead of explicitly mentioning XDG this part should just redirect users to platformdirs documentation as the canonical source.

memchr: > Come to think of it Macs are also an issue. Maybe instead of explicitly mentioning XDG this part should just redirect users to platformdirs documentation as the canonical source.

I don't have a Mac and I'm not familiar with it.

I can only test this PR on Linux, I'd be happy if someone could help and test the PR on other platforms. 

Perhaps it's best to limit this change to Linux for now, if it can't be tested.

uranusjr: The code is fine, only the documentation is problematic.

docs/installation.md
```diff
@@ -71,6 +71,18 @@ sudo PIPX_HOME=/opt/pipx PIPX_BIN_DIR=/usr/local/bin pipx install PACKAGE
...
+    
+    | Old Path               | New Path                     |
+    | ---------------------- | ---------------------------- | 
+    | `~/.local/pipx/.trash` | `$XDG_DATA_HOME/pipx/trash`  |
+    | `~/.local/pipx/shared` | `$XDG_DATA_HOME/pipx/shared` |
+    | `~/.local/pipx/venvs`  | `$XDG_DATA_HOME/pipx/venv`   |
+    | `~/.local/pipx/.cache` | `$XDG_CACHE_HOME/pipx`       |
+    | `~/.local/pipx/logs`   | `$XDG_STATE_HOME/pipx/log`   |
```
chrysle, line 84: How about linking to [this part of the platformdirs documentation](https://platformdirs.readthedocs.io/en/latest/api.html#platforms) in the above paragraph? That contains all relevant platform-specific information.

And I'd suggest replacing `~` in the table with something like `HOMEPATH`.

uranusjr: `~` seems better to me, it’s more or less commonly understood as to be the home directory (even to Windows users; I think Powershell can handle this shorthand), but something like `HOMEPATH` is not universal.

docs/installation.md
memchr: Sorry for the late reply. Can you tell me the default pipx on Windows and Mac **before** this change, so I can update this documentation accordingly?

uranusjr: Before the change they are exactly the same as on Linux!

memchr: Understood, then I think the current change to the documentation is fine.

docs/installation.md
```diff
@@ -62,7 +62,7 @@ Example configuration for use of the code linter [yapf](https://github.com/googl
 
 The default binary location for pipx-installed apps is `~/.local/bin`. This can be overridden with the environment variable `PIPX_BIN_DIR`.
 
-pipx's default virtual environment location is `~/.local/pipx`. This can be overridden with the environment variable `PIPX_HOME`.
+pipx's default virtual environment location is typically `~/.local/share/pipx` on Linux/Unix `%USERPROFILE%\AppData\Local\pipx` on Windows and `~/Library/Application Support/pipx` on Mac OS , and for compatibility reasons, if `~/.local/pipx` exists, it will be used as the default location instead. This can be overridden with the `PIPX_HOME` environment variable.
```
uranusjr, line 65:
```suggestion
pipx's default virtual environment location is typically `~/.local/share/pipx` on Linux/Unix `%USERPROFILE%\AppData\Local\pipx` on Windows and `~/Library/Application Support/pipx` on macOS, and for compatibility reasons, if `~/.local/pipx` exists, it will be used as the default location instead. This can be overridden with the `PIPX_HOME` environment variable.
```

docs/how-pipx-works.md
```diff
@@ -2,17 +2,17 @@
 
 When installing a package and its binaries (`pipx install package`) pipx will
```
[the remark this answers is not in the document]
memchr: how about `When installing a package and its binaries on linux`

uranusjr (approved):

docs/installation.md
```diff
@@ -62,7 +62,7 @@ Example configuration for use of the code linter [yapf](https://github.com/googl
 
 The default binary location for pipx-installed apps is `~/.local/bin`. This can be overridden with the environment variable `PIPX_BIN_DIR`.
 
-pipx's default virtual environment location is `~/.local/pipx`. This can be overridden with the environment variable `PIPX_HOME`.
+pipx's default virtual environment location is typically `~/.local/share/pipx` on Linux/Unix `%USERPROFILE%\AppData\Local\pipx` on Windows and `~/Library/Application Support/pipx` on macOS, and for compatibility reasons, if `~/.local/pipx` exists, it will be used as the default location instead. This can be overridden with the `PIPX_HOME` environment variable.
```
dukecat0, line 65:
```suggestion
pipx's default virtual environment location is typically `~/.local/share/pipx` on Linux/Unix, `%USERPROFILE%\AppData\Local\pipx` on Windows and `~/Library/Application Support/pipx` on macOS, and for compatibility reasons, if `~/.local/pipx` exists, it will be used as the default location instead. This can be overridden with the `PIPX_HOME` environment variable.
```

dukecat0 (approved):

dukecat0:
Thanks for the PR!

Baerbeisser:
Hmm, having moved `~/.local/*` and `~/.config` to `~/Data` (i rebooted), pipx got dependency issues. So i deleted the now `~/Data/data/pipx` (data being share) and installing now succeeds, but an empty `~/.local/share/man` gets created and the binary symlinks get placed in `~/.local/bin`.

My XDG variables:
```
XDG_BIN_HOME=/home/me/Data/bin
XDG_CACHE_HOME=/home/me/Data/cache
XDG_CONFIG_HOME=/home/me/Data/config
XDG_DATA_HOME=/home/me/Data/data
XDG_STATE_HOME=/home/me/Data/state
```

memchr:
The [spec](https://specifications.freedesktop.org/basedir-spec/latest/) doesn't include XDG_BIN_HOME.

Baerbeisser:
Oh, right, it's a unofficial one. I also have XDG_LOG_HOME, because a lot of tools still use that (and i find it more convenient).

 Still, XDG_DATA_HOME is set; it shouldn't create a .local/bin.

Is there somewhere a config i need to update/delete?

PR #1002, opened 2023-06-18
Use py launcher to select Python interpreter

### LINKED ISSUE #940: Windows: Use the Python Launcher to Select Python Version
In macOS and Linux, pipx allows you to set the Python version using the `--python` option. For example, `pipx install --python python3.11 black`. However, in Windows, all Python executables are named `python.exe`. So the only solution I have found to tell pipx the interpreter is by giving the full path like:

```
pipx install --python 'C:\\Users\\dan\\AppData\\Local\\Programs\\Python\\Python311\\python.exe' black
```

The standard way to select a Python version in Windows is with the Python Launcher. For example, `py -3.11`, will launch Python 3.11. It is possible to use the launcher with pipx with `pipx install --python py black`. However, I don't see a way to tell `py` which version of Python to use.

In Windows on GitHub Actions, pipx is installed by default with Python 3.9. I would like to run the setup-python action and then tell pipx to use this later Python version.

`pipx install --python 3.11 black` would use `py -3.11` on Windows and `python3.11` in Linux/macOS. An alternative might be `--python py311` which is the format tox uses.

Reinstall pipx for each Python version, or use the full path to the Python executable.

I would be glad to help implement this if there is agreement that this would be a nice enhancement.

danyeaw:

## Summary of changes

This PR closes #940 by adding support to use the py launcher to select the Python interpreter when the `--python` option is provided.

The feature is used like `pipx install --python 3.10 cowsay` to install using the Python 3.10 interpreter. The only way to do this right now is by passing in the full path to the interpreter.

I also did some refactoring in nearby code to leave the codebase in better shape than I found it.

## Test plan

Tested by running
```
> pipx install --python 3.10 cowsay`
> pipx reinstall --python 3.11 cowsay`
> pipx uninstall cowsay`
> pipx run --python 3.9 cowsay hi`
```
`pipx install --python 3.100 cowsay` raises a FileNotFound error, which is similar to the behavior in Linux if you run `pipx install --python python3.100 cowsay`.

### DIFF
--- a/src/pipx/commands/run.py
+++ b/src/pipx/commands/run.py
@@ -188,7 +188,7 @@ def run(
         # we can't parse this as a package
         package_name = app
 
-    if constants.WINDOWS and python and not Path(python).is_file():
+    if python and not Path(python).is_file():
         py_launcher = find_py_launcher_python(python)
         if py_launcher:
             python = py_launcher
--- a/src/pipx/commands/run.py
+++ b/src/pipx/commands/run.py
@@ -288,7 +288,7 @@ def _get_temporary_venv_path(
     m.update(python.encode())
     m.update("".join(pip_args).encode())
     m.update("".join(venv_args).encode())
-    venv_folder_name = m.hexdigest()[0:15]  # 15 chosen arbitrarily
+    venv_folder_name = m.hexdigest()[:15]  # 15 chosen arbitrarily
     return Path(constants.PIPX_VENV_CACHEDIR) / venv_folder_name
 
 
--- a/src/pipx/commands/run.py
+++ b/src/pipx/commands/run.py
@@ -180,7 +180,6 @@ def run(
     package
     """
 
-    package_or_url = spec if spec is not None else app
     # For any package, we need to just use the name
     try:
         package_name = Requirement(app).name
@@ -198,6 +197,7 @@ def run(
     if content is not None:
         run_script(content, app_args, python, pip_args, venv_args, verbose, use_cache)
     else:
+        package_or_url = spec if spec is not None else app
         run_package(
             package_name,
             package_or_url,
--- a/src/pipx/commands/run.py
+++ b/src/pipx/commands/run.py
@@ -322,7 +322,7 @@ def _http_get_request(url: str) -> str:
         return res.read().decode(charset)
     except Exception as e:
         logger.debug("Uncaught Exception:", exc_info=True)
-        raise PipxError(str(e))
+        raise PipxError(str(e)) from e
 
 
 def _get_requirements_from_script(content: str) -> Optional[List[str]]:
@@ -356,7 +356,7 @@ def _get_requirements_from_script(content: str) -> Optional[List[str]]:
         try:
             req = Requirement(line_content)
         except InvalidRequirement as e:
-            raise PipxError(f"Invalid requirement {line_content}: {str(e)}")
+            raise PipxError(f"Invalid requirement {line_content}: {str(e)}") from e
 
         # Use the normalised form of the requirement,
         # not the original line.
--- a/src/pipx/commands/run.py
+++ b/src/pipx/commands/run.py
@@ -75,7 +75,7 @@ def run_script(
 ) -> NoReturn:
     requirements = _get_requirements_from_script(content)
     if requirements is None:
-        exec_app([str(python), "-c", content, *app_args])
+        exec_app([python, "-c", content, *app_args])
     else:
         # Note that the environment name is based on the identified
         # requirements, and *not* on the script name. This is deliberate, as
--- a/src/pipx/commands/run.py
+++ b/src/pipx/commands/run.py
@@ -14,7 +14,6 @@ from pipx import constants
 from pipx.commands.common import package_name_from_spec
 from pipx.constants import TEMP_VENV_EXPIRATION_THRESHOLD_DAYS, WINDOWS
 from pipx.emojis import hazard
-from pipx.interpreter import find_py_launcher_python
 from pipx.util import (
     PipxError,
     exec_app,
@@ -187,11 +186,6 @@ def run(
         # Raw URLs to scripts are supported, too, so continue if
         # we can't parse this as a package
         package_name = app
-
-    if python and not Path(python).is_file():
-        py_launcher_python = find_py_launcher_python(python)
-        if py_launcher_python:
-            python = py_launcher_python
 
     content = None if spec is not None else maybe_script_content(app, is_path)
     if content is not None:
--- a/src/pipx/commands/run.py
+++ b/src/pipx/commands/run.py
@@ -194,11 +194,7 @@ def run(
         if py_launcher:
             python = py_launcher
 
-    if spec is not None:
-        content = None
-    else:
-        content = maybe_script_content(app, is_path)
-
+    content = None if spec is not None else maybe_script_content(app, is_path)
     if content is not None:
         run_script(content, app_args, python, pip_args, venv_args, verbose, use_cache)
     else:
--- a/src/pipx/commands/run.py
+++ b/src/pipx/commands/run.py
@@ -14,6 +14,7 @@ from pipx import constants
 from pipx.commands.common import package_name_from_spec
 from pipx.constants import TEMP_VENV_EXPIRATION_THRESHOLD_DAYS, WINDOWS
 from pipx.emojis import hazard
+from pipx.interpreter import find_py_launcher_python
 from pipx.util import (
     PipxError,
     exec_app,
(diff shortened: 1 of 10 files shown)


src/pipx/interpreter.py
```diff
@@ -26,14 +27,22 @@ def has_venv() -> bool:
 # so we try to locate the system Python and use that instead.
 
 
+def find_py_launcher_python(python_version: Optional[str] = None) -> Optional[str]:
+    py = shutil.which("py")
+    if py and python_version:
+        py = subprocess.run(
+            ["py", f"-{python_version}", "-c", "import sys; print(sys.executable)"],
```
chrysle, line 34: Could you try using the path returned by `shutil`?

```suggestion
            [py, f"-{python_version}", "-c", "import sys; print(sys.executable)"],
```

danyeaw: Hi @chrysle, good catch, I updated it!

uranusjr:
Some thoughts:

1. There’s an unofficial `py` launcher that works cross platform. While it is not a part of the official distribution, it may be a good idea to extend this feature to all platforms since it costs us nothing.
2. I wonder if we should unify the value format a bit. For example, if `py` is present in `PATH` and the user provides `pythonX.Y`, maybe we could automatically strip `python` and try `py -X.Y`? Not sure.
3. On the flip side, if the user gives `X.Y` and `py` is not found, maybe we can automatically try `pythonX.Y` in Python? I am less in favour of this one; seems a bit too magical.

danyeaw:
Hi @uranusjr, thanks so much for the feedback! I haven't used py launcher in Linux before, but I can investigate adding support for it as well. Could that be part of a follow-up PR?

For the 2nd and 3rd points, I thought that your direction in the original issue was to avoid magic.

The nice thing about the 2nd option would be that a single command could be used across platforms, but no magic is fine as well.

uranusjr:
The second point is mainly to maintain existing usages. We already support `--python python.X.Y` and it would be nice to make it automatically work on Windows.

The UNIX launcher can be found here: https://github.com/brettcannon/python-launcher
It’s mostly compatible with the Windows one, so maybe the best way to approach this is to simply drop the Windows check in this PR and investigate further later.

src/pipx/commands/run.py
```diff
@@ -189,14 +188,16 @@ def run(
...
         package_name = app
 
-    if spec is not None:
-        content = None
-    else:
-        content = maybe_script_content(app, is_path)
+    if python and not Path(python).is_file():
+        py_launcher = find_py_launcher_python(python)
+        if py_launcher:
+            python = py_launcher
```
uranusjr, line 194:
```suggestion
        py_launcher_python = find_py_launcher_python(python)
        if py_launcher_python:
            python = py_launcher_python
```

For clarity, otherwise this reads like we’re assigning _the launcher_ to `python`.

src/pipx/main.py
```diff
@@ -181,6 +181,10 @@ def run_pipx_command(args: argparse.Namespace) -> ExitCode:  # noqa: C901
         logger.info(f"Virtual Environment location is {venv_dir}")
     if "skip" in args:
         skip_list = [canonicalize_name(x) for x in args.skip]
+    if "python" in args and WINDOWS and not Path(args.python).is_file():
+        py_launcher = find_py_launcher_python(args.python)
```
uranusjr, line 185: Similar to above

src/pipx/main.py
```diff
@@ -342,7 +346,8 @@ def _add_install(subparsers: argparse._SubParsersAction) -> None:
         default=DEFAULT_PYTHON,
         help=(
             "The Python executable used to create the Virtual Environment and run the "
-            "associated app/apps. Must be v3.6+."
+            "associated app/apps. The Python version for the Python Launcher. "
+            "Must be v3.6+."
         ),
```
uranusjr, line 351: The message reads like a salad of words and not really comprehensive to me. Could use a rewrite from scratch. (Same for other similar arguments)

src/pipx/commands/run.py
danyeaw: Thanks, I agree that is much more clear.

src/pipx/main.py
danyeaw: Good one, I updated the wording, hopefully this is better :+1:

danyeaw:
Hi team, is there anything else needed before we merge this?

src/pipx/commands/run.py
```diff
@@ -189,14 +188,16 @@ def run(
         # we can't parse this as a package
         package_name = app
 
-    if spec is not None:
-        content = None
-    else:
-        content = maybe_script_content(app, is_path)
+    if python and not Path(python).is_file():
```
dukecat0, line 191: I guess this check is unnecessary here?
See comment below

src/pipx/main.py
```diff
@@ -181,6 +181,10 @@ def run_pipx_command(args: argparse.Namespace) -> ExitCode:  # noqa: C901
         logger.info(f"Virtual Environment location is {venv_dir}")
     if "skip" in args:
         skip_list = [canonicalize_name(x) for x in args.skip]
+    if "python" in args and WINDOWS and not Path(args.python).is_file():
```
dukecat0, line 184: Repeated here

dukecat0 (changes requested):

src/pipx/commands/run.py
danyeaw: Hi @dukecat0, nice catch! It looks like I missed cleaning this up. I removed the logic in the run module entirely, since it is handled by main already. 👍

src/pipx/main.py
```diff
@@ -588,7 +592,10 @@ def _add_run(subparsers: argparse._SubParsersAction) -> None:
     p.add_argument(
         "--python",
         default=DEFAULT_PYTHON,
-        help="The Python version to run package's CLI app with. Must be v3.6+.",
+        help=(
+            "Python to run with. Possible values can be the executable name (python3.11), "
+            "the version to pass to py launcher (3.11), or the full path to the executable."
```
dukecat0, line 597: See above comment

src/pipx/main.py
```diff
@@ -512,8 +516,8 @@ def _add_reinstall_all(subparsers: argparse._SubParsersAction) -> None:
         "--python",
         default=DEFAULT_PYTHON,
         help=(
-            "The Python executable used to recreate the Virtual Environment "
-            "and run the associated app/apps. Must be v3.6+."
+            "Python to reinstall with. Possible values can be the executable name (python3.11), "
+            "the version to pass to py launcher (3.11), or the full path to the executable."
```
dukecat0, line 520: See above comment

src/pipx/main.py
```diff
@@ -484,8 +488,8 @@ def _add_reinstall(subparsers, venv_completer: VenvCompleter) -> None:
         "--python",
         default=DEFAULT_PYTHON,
         help=(
-            "The Python executable used to recreate the Virtual Environment "
-            "and run the associated app/apps. Must be v3.6+."
+            "Python to reinstall with. Possible values can be the executable name (python3.11), "
```
dukecat0, line 491: See above comment

src/pipx/main.py
```diff
@@ -341,8 +345,8 @@ def _add_install(subparsers: argparse._SubParsersAction) -> None:
         "--python",
         default=DEFAULT_PYTHON,
         help=(
-            "The Python executable used to create the Virtual Environment and run the "
-            "associated app/apps. Must be v3.6+."
+            "Python to install with. Possible values can be the executable name (python3.11), "
+            "the version to pass to py launcher (3.11), or the full path to the executable."
```
dukecat0, line 349: How about adding `Must be v3.7+`?

src/pipx/main.py
chrysle: I think better `v3.8+`, since version _3.7_ will (hopefully) be dropped soon.

chrysle: Thinking of maintenance, how about adding a constant like `MINIMUM_PYTHON_VERSION` to [constants.py](https://github.com/pypa/pipx/blob/main/src/pipx/constants.py), which could then be imported into `main.py`?

dukecat0: Good point!

danyeaw: Thanks for the suggestion @dukecat0 and @chrysle! I added a `MINIMUM_PYTHON_VERSION` constant.

uranusjr (approved):

dukecat0 (approved):

dukecat0:
Thanks!

danyeaw:
Thanks team for your reviews and getting this merged, I think this will be a nice improvement to pipx!

PR #1005, opened 2023-06-19
Drop Python 3.7

### LINKED ISSUE #1040: "Python 3.6 or later is required."
**Describe the bug**

pipx is documented to support Python >= 3.7:
https://github.com/pypa/pipx/blob/238ae43bf4fd1797154ee07a40d045b0e81757e8/pyproject.toml#L10

This test actually tests Python >=3.6:
https://github.com/pypa/pipx/blob/238ae43bf4fd1797154ee07a40d045b0e81757e8/src/pipx/__init__.py#L3-L8

Is the above test really needed, in addition to `requires-python`? Which corner case is it supposed to catch?

Could be fixed by https://github.com/pypa/pipx/pull/1005#issuecomment-1676411329.

chrysle:

## Summary of changes

Dropped support for Python 3.7 which is EOL on June 26, 2023.

Closes #1040 

## Test plan

Tested by running
```
```

### DIFF
--- a/.github/workflows/create_tests_package_lists.yml
+++ b/.github/workflows/create_tests_package_lists.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
 
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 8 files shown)


DimitriPapadopoulos:
I would recommend changing this too:
https://github.com/pypa/pipx/blob/238ae43bf4fd1797154ee07a40d045b0e81757e8/src/pipx/__init__.py#L3-L8

chrysle:
> I would recommend changing this too

Thanks, but I already did that.

dukecat0:
We may also need to remove this:
https://github.com/pypa/pipx/blob/b489691c61622c495ae8aeece2ef1624eb45310c/pyproject.toml#L27

dukecat0:
@chrysle I think it's time to merge this. Could you apply the above comment and resolve the conflict?

chrysle:
> @chrysle I think it's time to merge this. Could you apply the above comment and resolve the conflict?

Sure, done!

dukecat0 (approved):

PR #1006, opened 2023-06-19
(🐞) Don't use `repr` for paths in console output

### LINKED ISSUE #1003: (🐞) Do not use `repr` to format paths shown in console
```
> pipx install ...
...
⚠️  Note: 'C:\\Users\\amongus\\.local\\bin' is not on your PATH environment variable.
...
```

There shouldn't be double backslashes here, single only.

KotlinIsland:
* [🚀] I have added an entry to `docs/changelog.md`

## Summary of changes
Previously pipx would output things like `"File 'foo\\bar\\baz' not found."`, which isn't ideal.

- resolves #1003

## Test plan

Tested by running
```
> pipx install basedmypy
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Fallback to user's log path if the default log path (`$PIPX_HOME/logs`) is not writable to aid with pipx being used for multi-user (e.g. system-wide) installs of applications
 - Fix wrong interpreter usage when injecting local pip-installable dependencies into venvs
 - add pre-commit hook support
+- Don't show escaped backslashes for paths in console output 
 
 ## 1.2.0
 
--- a/src/pipx/commands/common.py
+++ b/src/pipx/commands/common.py
@@ -409,7 +409,7 @@ def warn_if_not_on_path(local_bin_dir: Path) -> None:
         logger.warning(
             pipx_wrap(
                 f"""
-                {hazard}  Note: {str(local_bin_dir)!r} is not on your PATH
+                {hazard}  Note: '{local_bin_dir}' is not on your PATH
                 environment variable. These apps will not be globally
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - Fallback to user's log path if the default log path (`$PIPX_HOME/logs`) is not writable to aid with pipx being used for multi-user (e.g. system-wide) installs of applications
 - Fix wrong interpreter usage when injecting local pip-installable dependencies into venvs
 - add pre-commit hook support
+- Don't use `repr` for paths in console output
```
chrysle, line 8: I don't think this is worth a changelog entry.

uranusjr: I would add an entry, but not like this. It should say something about not showing double backslashes on Windows instead.

src/pipx/commands/common.py
```diff
@@ -409,7 +409,7 @@ def warn_if_not_on_path(local_bin_dir: Path) -> None:
         logger.warning(
             pipx_wrap(
                 f"""
-                {hazard}  Note: {str(local_bin_dir)!r} is not on your PATH
+                {hazard}  Note: '{str(local_bin_dir)}' is not on your PATH
```
uranusjr, line 412:
```suggestion
                {hazard}  Note: '{local_bin_dir}' is not on your PATH
```

CHANGELOG.md
KotlinIsland: Done

src/pipx/commands/common.py
KotlinIsland: Opps, forgot to push 🤦

uranusjr (approved):

dukecat0 (approved):
Thanks!

PR #1015, opened 2023-07-03
chore: ruff moved to astral-sh

henryiii:
See https://github.com/scientific-python/cookie/pull/200.


chrysle:
Though this was not breaking the hook.

henryiii:
Yeah, GitHub redirects when a repo moves.

Issue #978, opened 2023-05-15
ensurepath falsely detecting C:\Users\...\AppData\Roaming\Python\Python310\Scripts as being in PATH on Windows


Vuizur:
As instructed by the terminal/tutorial on https://pypa.github.io/pipx/, I ran `.\pipx ensurepath` to add the PATH to Windows. I got this message:

```
Success! Added C:\Users\...\.local\bin to the PATH environment variable.
C:\Users\...\AppData\Roaming\Python\Python310\Scripts is already in PATH.

Consider adding shell completions for pipx. Run 'pipx completions' for
instructions.

You will need to open a new terminal or re-login for the PATH changes to take  
effect.

Otherwise pipx is ready to go! ✨ 🌟 ✨
```
However, the part about `C:\Users\...\AppData\Roaming\Python\Python310\Scripts` was not true, I restarted the terminal/my PC, but it wouldn't detect pipx, only adding the listed path manually fixed the problem. I did have some similar folders in my PATH (C:/Python10/Scripts), so maybe this could be the reason (and the regex to detect already existing values in PATH maybe is not specific enough).
(It did add `C:\Users\...\.local\bin` correctly though).

PS: Thanks a lot for maintaing this project!

uranusjr:
A way to reproduce the issue (with say a virtual machine) would be helpful.

Vuizur:
I looked now a bit into the pipx code and think the reason might be that if you execute a python script in a specific directory that is not in PATH and then call `print(userpath.in_current_path(os.getcwd()))` it still prints `True`.

Vuizur:
I opened an issue here: https://github.com/ofek/userpath/issues/45

dukecat0:
Closing this since the fix in userpath was merged.

Issue #1007, opened 2023-06-20
pipx installation instructions on Windows don't seem to work


markmelvin:
I love `pipx` and am recommending it to people using my packages, however the first two people that tried to install it failed miserably.

These people both have Windows 10, and Python 3.x. I believe Python was installed from the Python installer for all users and resides in:

```
<USER_folder>\AppData\Local\Programs\Python\Python311-32\python.exe
``` 

Installing `pipx` with the recommended command `python -m pip install --user pipx`, installs `pipx` into  `<USER_folder>\AppData\Roaming\Python\Python3x\Scripts` . Then going into that folder and running `pipx ensurepath` doesn't seem to do anything to put that folder on the PATH. Thus, from then on, running `pipx` just fails to find the executable.

For one user, we manually hacked their PATH to add the right folder to it. For the other user, we removed `pipx` and installed it without the `--user` flag to install it into their Local folder as opposed to the Roaming one and it works.

But the out-of-box experience for both users was that installing `pipx` didn't work and I needed to hold their hand to figure out what went wrong. Not sure, but this may be related to https://github.com/pypa/pipx/issues/978

gaborbernat:
> In theory this should be fixed once the next version of the userpath library gets published.

Issue #996, opened 2023-06-01
pipx ensurepath adds itself multiple times


tripleee:
**Describe the bug**

Running `pipx ensurepath` adds several redundant instances of the `$HOME/.local/bin` directory to my `PATH`. These are harmless as such, but somewhat annoying.

**How to reproduce**

`pipx ensurepath` and examine the end of recently modified files in your home directory.

```
cd
ls -art ~ | tail -n 2 | xargs tail -n 7
```

On a fresh Linux Mint installation, I see this:

```
==> .profile <==
# set PATH so it includes user's private bin if it exists
if [ -d "$HOME/.local/bin" ] ; then
    PATH="$HOME/.local/bin:$PATH"
fi

# Created by `pipx` on 2023-06-01 04:10:01
export PATH="$PATH:/home/tripleee/.local/bin"

==> .bashrc <==
  elif [ -f /etc/bash_completion ]; then
    . /etc/bash_completion
  fi
fi

# Created by `pipx` on 2023-06-01 04:10:01
export PATH="$PATH:/home/tripleee/.local/bin"
```

On a Mac with Zsh, where I installed `pipx` with `brew`, I see similar behavior with `.zprofile` and `.zshrc`

**Expected behavior**

1. If the file already has a stanza which (conditionally) adds `$HOME/.local/bin` to my `PATH`, ideally, do nothing.
2. Or, at the very least, wrap it so that it doesn't unnecessarily add the same directory to the `PATH` multiple times, like the code in the default `.profile` on Mint (and I imagine Ubuntu) already does.
3. The `export` is arguably useless; the system already exports `PATH`.
4. Ideally, don't modify both files. One or the other is sufficient.

There are complications around how the shell loads its startup files; what exactly it does out of the box and how exactly each OS or distro modifies these files to produce consistent and predictable behavior is rather a can of worms. The "belt and suspenders" approach of adding something to both the `.profile` (or `.bash_profile`, `.zprofile`, etc) and `.bashrc` (or `.zshrc` etc) might thus be defendable if the other symptoms can be avoided.

essn:
Bash on Linux - same thing. It actually overwrites my PATH because of the `export PATH=` statement.. Easy enough to fix - but frustrating. On login, I need to source `~/.bash_profile` in order to have my usual PATH.

chrysle:
> It actually overwrites my PATH because of the `export PATH=` statement..

[...]

PR #1021, opened 2023-07-22
Bump userpath to 1.9.0

dukecat0:

## Summary of changes

Bump userpath to 1.9.0

## Test plan

Tested by running
```
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## dev
 
+- Bump `userpath` to 1.9.0
 - Make usage message in `pipx run` show `package_or_url`, so extra will be printed out as well
 - Add `--force-reinstall` to pip arguments when `--force` was passed
 - Use the py launcher, if available, to select Python version with the `--python` option
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ dependencies = [
   "importlib-metadata>=3.3.0; python_version < '3.8'",
   "packaging>=20.0",
   "platformdirs>=2.1.0",
-  "userpath>=1.6.0",
+  "userpath>=1.9.0",
 ]
 dynamic = ["version"]
 


uranusjr:
Why do we need to raise the lower bound?

chrysle:
I guess because of ofek/userpath#48 and #996.

safl:
Hmm, have you considered that userpath v1.9.0 breaks automation use-cases?
Specifically I have this issue: https://github.com/ofek/userpath/issues/50 with it.

uranusjr:
I don’t feel adding an upper cap would be a good idea since it would require a new release when say 1.10 come out. You can add that upper limit yourself if it suits your use case, but someone else can’t remove one if that doesn’t work for them.

safl:
> I don’t feel adding an upper cap would be a good idea since it would require a new release when say 1.10 come out. You can add that upper limit yourself if it suits your use case, but someone else can’t remove one if that doesn’t work for them.

- An upper-bound would prevent a functional regression like the one that is happening right now
  - A new release of ``pipx`` would not be required when ``userpath`` is released, unless it provides functionality needed by ``pipx``. What could have happened here is that someone might have said "oh wait, bumping ``userpath`` to v1.9.0 introduces a regression in the behavior of ``pipx ensurepath``, let's resolve that before bumping". Or, if there were tests for non-login shells, then they would catch it. Rather than the functional regression seeping into ``pipx``.
  - I guess it is a trade-off between "conveniently" getting third-party code bumped automatically with the risk of ``pipx`` sporadically breaking / altering behavior (for the same version of ``pipx``) **vs** the chore of verifying changes to third-party code with the upside of stable / non-altering behavior (for the same version of ``pipx``)

- I cannot control the version of dependencies when installing ``pipx`` via the system package manager
  - With the introduction of https://peps.python.org/pep-0668/ then ``pipx`` is a convenient way to manage command-line tools implemented in Python. However, with the regression introduced by ``userpath`` v1.9.0 then it no longer applies for non-login shells, so basically if you run your commands via SSH-exec (``ssh user@host "my_command"``).
  - Installing ``pipx`` via the system package manager is a really convenient way to bootstrap the availability of Python-based command-line tools and with ``pipx ensurepath`` even more so
  - Thus, for example on Debian, when https://packages.debian.org/bookworm/python3-userpath is upgraded to v1.9.0, then using ``pipx ensurepath`` won't be usable for non-login shells any longer, I won't be able to fix that.
  - As I understand the motivation behind ``userpath`` dropping non-login shells, then it was not because something was "not working" but rather that it was "not nice that there where duplicate entries in the ``PATH`` variable. I might be wrong, but I just weigh functional regressions higher than nice-to-haves.

dukecat0:
> Why do we need to raise the lower bound?

Originally I thought this would fix #978 and #1007, but seems like now it can lead to more issues, so I am closing this PR.

PR #1027, opened 2023-07-31
Added asreview and mu-repo to programs to try

george-gca:

## Summary of changes
Added [asreview](https://github.com/asreview/asreview) and [mu-repo](https://github.com/fabioz/mu-repo) to programs to try


uranusjr:
I’m deciding to close this since the additional apps do not seem to add to how pipx works. I understand they are useful tools, but the list is not for showcasing different applications, but using selected applications to showcase pipx. The new additions do not help with that.

PR #1029, opened 2023-08-01
Pass --no-input

### LINKED ISSUE #219: pipx install doesn't prompt for credentials for private vcs url

Darsstar:

## Summary of changes
Pass `--no-input` to Pip.

Fixes #219 

This is a breaking change for keyring users. But was it ever properly supported anyway? As someone who needs the `artifacts-keyring` package to authenticate with our private repo I'm going with the answer being `no, you will have to apply hacks if this is important to you`. Failing fast is better then hanging forever because the user can't see the prompt asking for a username and password.

https://github.com/pypa/pip/pull/11698 which is released in Pip 23.1 added a section to the documentation how Pipx users who rely on the keyring library can set things up so keyring will be used again: https://pip.pypa.io/en/stable/topics/authentication/#using-keyring-as-a-command-line-application.

## Test plan
It should not have any impact when keyring is not required to authenticate with the repository.

Tested by running the test suite. 
```
pipx run --index https://pypi.org/simple/ cowsay no input
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,5 +30,5 @@ repos:
   rev: v1.1.1
   hooks:
   - id: mypy
-    args: ['--ignore-missing-imports', '--strict-equality','--no-implicit-optional']
+    args: ['--warn-unused-ignores', '--strict-equality','--no-implicit-optional']
     exclude: 'testdata/test_package_specifier/local_extras/setup.py'
--- a/src/pipx/venv.py
+++ b/src/pipx/venv.py
@@ -240,11 +240,15 @@ class Venv:
         ):
             # do not use -q with `pip install` so subprocess_post_check_pip_errors
             #   has more information to analyze in case of failure.
-            cmd = (
-                [str(self.python_path), "-m", "pip", "--no-input", "install"]
-                + pip_args
-                + [package_or_url]
-            )
+            cmd = [
+                str(self.python_path),
+                "-m",
+                "pip",
+                "--no-input",
+                "install",
+                *pip_args,
+                package_or_url,
+            ]
             # no logging because any errors will be specially logged by
             #   subprocess_post_check_handle_pip_error()
             pip_process = run_subprocess(cmd, log_stdout=False, log_stderr=False)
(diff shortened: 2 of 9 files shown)


.pre-commit-config.yaml
```diff
@@ -27,8 +27,8 @@ repos:
 #   cannot use --warn-unused-ignores because it conflicts with
 #     --ignore-missing-imports
 - repo: https://github.com/pre-commit/mirrors-mypy
-  rev: v1.1.1
+  rev: v0.930
```
uranusjr, line 30: Why is this downgraded?

Darsstar: Github actions runs nox, 0.930 is the version listed in `noxfile.py`. I am of the opinion they should be consistent.

Want me to upgrade the version nox uses instead?

Darsstar: Done

src/pipx/venv.py
```diff
@@ -301,7 +301,11 @@ def install_package_no_deps(self, package_or_url: str, pip_args: List[str]) -> s
             f"determining package name from {package_or_url!r}", self.do_animation
         ):
             old_package_set = self.list_installed_packages()
-            cmd = ["install"] + ["--no-dependencies"] + pip_args + [package_or_url]
+            cmd = (
+                ["--no-input", "install", "--no-dependencies"]
+                + pip_args
+                + [package_or_url]
+            )
```
uranusjr, line 308:
```suggestion
            cmd = [
                "--no-input",
                "install",
                "--no-dependencies",
                *pip_args,
                package_or_url,
            ]
```

This is growing too complicated

Darsstar: I left some in the form of `cmd = [<snip>] + cmd` alone instead of changing it to `cmd = [<snip>, *cmd]`.

uranusjr (approved):

Issue #1025, opened 2023-07-27
Please fix functional regression for non-login shells


safl:
**Functional Regression**

* Command-line tools installed via ``pipx`` and setup with ``pipx ensurepath`` can no longer be executed with **SSH**
  * Example / running: ``ssh user@hostname "my_command"`` will give you the error ``command not found``
  * Similarly when running commands via [Paramiko](https://github.com/paramiko/paramiko) ``exec_command()``
  * Generally anything using the **SSH** exec-channel, as it is intended for automated / non-interactive execution, and thus **does not** and  **should not** consume environment-setup for login-shell

**Describe the bug**

- ``pipx ensurepath`` no longer sets up  ``PATH`` for non-login shells
  - This functional regression is introduced as the implementation is delegated to the ``userpath`` module, the ``userpath`` module changed behavior and dropped support for non-login shells in v1.9.0 see https://github.com/ofek/userpath/issues/50 for a request / discussion on changing this
  - This functional regression is introduced by bumping the required ``userpath`` module to v1.9.0
  - The bumping of ``userpath`` happens "automatically" since the pyproject.toml has ``userpath>=1.6.0``, see PR https://github.com/pypa/pipx/pull/1021 for a discussion on changing this

**Usecase**

With the introduction of https://peps.python.org/pep-0668/ on recent Linux distributions, then ``pipx`` is the recommended choice for managing the virtual environment for Python-based command-line tools and making them conveniently available in the user ``PATH`` via ``pipx ensurepath``. However, with the functional regression described here, then it will be less of a viable solution.

As a user of ``pipx`` for this usecase then I hope that you will re-instate the default behavior of ``pipx ensurepath`` to what is was before this functional regression was introduced. Fortunately, then **currently** only ``pipx`` installed via ``pip`` is  affected as the update to ``userpath`` has not hit the system packages **YET**.
Thus, if this is fixed now, then this regression can be prevented for installations of ``pipx`` via system packages.

And thanks for making ``pipx`` and ``userpath``!

PR #1034, opened 2023-08-07
fix(install): set v1.8.0 as upper-bound on userpath

safl:
This fixes the functional regression introduced by userpath v1.9.0. For details see issue #1025.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ dependencies = [
   "importlib-metadata>=3.3.0; python_version < '3.8'",
   "packaging>=20.0",
   "platformdirs>=2.1.0",
-  "userpath>=1.6.0",
+  "userpath>=1.6.0, <=1.9.1, !=1.9.0",
 ]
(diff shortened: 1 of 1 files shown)


dukecat0:
I guess we don't need this anymore since userpath v1.9.1 has been released.

uranusjr:
Or add a `!=1.9.0` instead?

safl:
I have bumped the version number, yet still keeping an upper-bound.
Without an upper-bound, then ``pipx`` is open for regressions from ``userpath``.

uranusjr:
I don’t think it is a good idea to add an upper bound. If the regression reasoning is followed, we should add upper bounds to *all* dependencies.

safl:
> I don’t think it is a good idea to add an upper bound. If the regression reasoning is followed, we should add upper bounds to _all_ dependencies.

Yes. From my perspective then adding upper bounds on _all_ dependencies is preferable to arbitrary functional regression.

PR #1037, opened 2023-08-11
Add `--requirement` option for `pipx inject`

### LINKED ISSUE #934: Support for `pipx inject <package> -r <requirements-file>`

I'd like to be able to inject packages into a package environment from a standard pip requirements file and have them registered individually as "injected" dependencies, e.g. for `list --include-injected` and `upgrade(-all) --include-injected`.

Alternatives listed below all have deficiencies in some way.

This is a different take on https://github.com/pypa/pipx/issues/359 and https://github.com/pypa/pipx/issues/357 (etc).

As mentioned by @funkyfuture in https://github.com/pypa/pipx/issues/359#issuecomment-593279235.

```
pipx inject <package> -r <requirements-file>
```

The standard solution below is OK, but doesn't register the additional packages as "injected". Hence they're invisible to commands that support `--include-injected`.

```
pipx install <package>
pipx runpip <package> install -r <requirements-file>
```

The following solution by @Sispheor and @hanfried doesn't work on Windows, may miss edge-cases, and is generally clumsy.

```
cat <requirements-file> | sed -e 's/#.*//' | xargs pipx inject <package>
```

dukecat0:

## Summary of changes
Add `--requirement` option for `pipx inject`.
Close #934 

## Test plan

Tested by running
```
pipx inject black -r requirements.txt
```


uranusjr:
Do we want to allow `pipx inject foo -r bar.txt`? If not maybe we can use `add_mutually_exclusive_group` (not sure if it can be used grouping _arguments_ and flags together).

dukecat0:
> Do we want to allow `pipx inject foo -r bar.txt`?

Did you mean something like `pipx inject black foo -r bar.txt`?

uranusjr:
Yes, sorry

dukecat0:
Probably no then, since I guess most users will choose to edit their requirements files directly.

dukecat0:
> If not maybe we can use `add_mutually_exclusive_group` (not sure if it can be used grouping arguments and flags together).

Seems like it's impossible to do that currently. I've added a check and it will raise an exception if `dependencies` and `--requirement` are passed at the same time.

uranusjr:
Probably need to mention this exclusiveness in the flag help string. Otherwise looks good to me.

src/pipx/commands/inject.py
```diff
@@ -84,6 +84,7 @@ def inject(
     package_name: Optional[str],
     package_specs: List[str],
     pip_args: List[str],
+    requirement_files: List[str],
```
uranusjr, line 87: This should be keyword-only (after the `*`) instead

src/pipx/commands/inject.py
```diff
@@ -95,8 +96,25 @@ def inject(
...
+
+    if package_specs and requirement_files:
+        raise PipxError(
+            "Dependencies and --requirement/-r option cannot be passed at the same time"
+        )
+
+    if requirement_files:
+        packages_list = []
+        for file in requirement_files:
+            with open(Path(file), "r") as f:
```
uranusjr, line 112: Instead of reading the files, should we just forward the path to pip to let it parse instead? Requirements files are technically a pip-specific format and parsing them ourselves seems like a big future source of code bloat down the road. Instead we should just redirect all the responsibilities to pip.

dukecat0: But the question is that how should we determine package names?

uranusjr: Why do we need the package names? This is `inject` not `install`.

dukecat0: Since we need to forward those package names to `pipx_metadata`.

uranusjr: Would it be viable to do a `pip freeze` before and after installation and record the diff?

uranusjr: Oh, or we can ask pip to provide an installation report! We can simply record packages with the `requested` flag. https://pip.pypa.io/en/stable/reference/installation-report/

dukecat0: Haven't heard about this before. I will check it out and see if it works.

dukecat0: I just took a look at the `pipx_metadata` and turns out `pip_args` can only be passed once currently. They will be forwarded to each `pip install` command during the injection. Perhaps we will need to make some changes to the metadata format if we want to pass the requirements file to pip directly, since there might be different options provided for each package.

dukecat0: > Oh, or we can ask pip to provide an installation report! We can simply record packages with the requested flag. **[pip.pypa.io/en/stable/reference/installation-report](https://pip.pypa.io/en/stable/reference/installation-report/)**

But it looks like the installation report will also report the dependencies of a package, which are not specified in requirements.txt. Though we can exclude them from pipx_metadata by checking whether they are in the `requires_dist` field, I personally think this is too complicated. 

Here's my own solution that should solve this issue:
- Parse the requirements file line by line
- Use `split()` to split those arguments out
- Check if the argument starts with `-` or `--`, which means they are options and should be added to `pip_args`
- Have the installation as usual
- Repeat the above steps till the last line

Any thoughts?

uranusjr: > But it looks like the installation report will also report the dependencies of a package, which are not specified in requirements.txt. Though we can exclude them from pipx_metadata by checking whether they are in the `requires_dist` field, I personally think this is too complicated

There’s a field `requested` for this, we can simply filter out entries with this set to true (meaning it’s from user input i.e. the requirements file).

dukecat0: Oops, didn't notice that, thanks for pointing it out!

gaborbernat (changes requested):
Merge conflicts need fixed.

gaborbernat:

jamesmyatt:
I've made a new PR #1252 for this, based partially on this one.

PR #1042, opened 2023-08-18
Speed up tests execution

chrysle:

## Summary of changes

Added the `pytest-xdist` plugin to the testing session to distribute the tests across multiple CPUs, if no positional arguments are given. This _significantly_ speeds up the tests execution.

## Test plan

Tested by running
```
nox
```


dukecat0:
IIRC I've tried this before, but this can cause different errors and always fail the tests.

gaborbernat:
> IIRC I've tried this before, but this can cause different errors and always fail the tests.

Yeah 👍

PR #1046, opened 2023-08-24
Add fades comparison to docs

### LINKED ISSUE #1045: Comparison to fades would be useful
The [comparison to other tools](https://pypa.github.io/pipx/comparisons/) in pipx's docs is really useful IMO.
Another tool people might want to compare pipx to is [fades](https://github.com/PyAr/fades), which is more focused on the idea of specifying all of one's dependencies in a single Python script than on running whole packages, but it has some overlap with pipx in both regards: AFAICT, pipx also supports specifying dependencies in a single file since https://github.com/pypa/pipx/issues/913 and fades supports running single files directly [from remote resources](https://github.com/PyAr/fades#how-to-execute-the-script-with-fades) (but not packages, again AFAICT).

It would be great if there was another section on the comparison page comparing pipx to fades.

Draw this comparison on some other PyPA package dedicated to this topic if it exists?

smheidrich:
  * I don't think that's needed here

## Summary of changes
Adds comparison to [fades](https://github.com/PyAr/fades) to the docs.

Fixes #1045.

## Test plan
Look at docs.

### DIFF
--- a/docs/comparisons.md
+++ b/docs/comparisons.md
@@ -90,7 +90,7 @@ None.
 
 [fades](https://github.com/PyAr/fades) is a tool to run **individual** Python scripts inside automatically provisioned virtualenvs with their dependencies installed.
 
-* Both [fades](https://github.com/PyAr/fades#how-to-mark-the-dependencies-to-be-installed) and [pipx](https://github.com/pypa/pipx/pull/916) allow specifying a script's dependencies in specially formatted comments, but the exact syntax differs (see also [PEP 722](https://peps.python.org/pep-0722/) which seeks to standardize such syntax).
+* Both [fades](https://github.com/PyAr/fades#how-to-mark-the-dependencies-to-be-installed) and [pipx](examples.md#pipx-run-examples) allow specifying a script's dependencies in specially formatted comments, but the exact syntax differs (see also [PEP 722](https://peps.python.org/pep-0722/) which seeks to standardize such syntax).
 * Both tools automatically set up reusable virtualenvs containing the necessary dependencies.
 * Both can download Python scripts/packages to execute from remote resources.
 * fades can only run individual script files while pipx can also run packages.
--- a/docs/comparisons.md
+++ b/docs/comparisons.md
@@ -85,3 +85,15 @@ pip-run poetry -- -m poetry --help
 
 Example interaction:
 None.
+
+## pipx vs fades
+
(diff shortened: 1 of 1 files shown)


dukecat0 (approved):

docs/comparisons.md
```diff
@@ -85,3 +85,15 @@ pip-run poetry -- -m poetry --help
 
 Example interaction:
 None.
+
+## pipx vs fades
+
+[fades](https://github.com/PyAr/fades) is a tool to run **individual** Python scripts inside automatically provisioned virtualenvs with their dependencies installed.
+
+* Both [fades](https://github.com/PyAr/fades#how-to-mark-the-dependencies-to-be-installed) and [pipx](https://github.com/pypa/pipx/pull/916) allow specifying a script's dependencies in specially formatted comments, but the exact syntax differs (see also [PEP 722](https://peps.python.org/pep-0722/) which seeks to standardize such syntax).
```
smheidrich, line 93: I linked the PR here instead of the docs because I didn't notice that pipx already has documentation on this which just isn't published yet because there hasn't been a release since. Will update.

smheidrich: Updated

docs/comparisons.md
```diff
@@ -85,3 +85,15 @@ pip-run poetry -- -m poetry --help
 
 Example interaction:
 None.
+
+## pipx vs fades
+
+[fades](https://github.com/PyAr/fades) is a tool to run **individual** Python scripts inside automatically provisioned virtualenvs with their dependencies installed.
+
+* Both [fades](https://github.com/PyAr/fades#how-to-mark-the-dependencies-to-be-installed) and [pipx](examples.md#pipx-run-examples) allow specifying a script's dependencies in specially formatted comments, but the exact syntax differs (see also [PEP 722](https://peps.python.org/pep-0722/) which seeks to standardize such syntax).
```
uranusjr, line 93:
```suggestion
* Both [fades](https://github.com/PyAr/fades#how-to-mark-the-dependencies-to-be-installed) and [pipx run](examples.md#pipx-run-examples) allow specifying a script's dependencies in specially formatted comments, but the exact syntax differs (see also [PEP 722](https://peps.python.org/pep-0722/) which seeks to standardize such syntax).
```

dukecat0:
Thanks!

Issue #857, opened 2022-06-23
Handling manpages


CharString:
**How would this feature be useful?**
Some packages, like `ipython` and `ranger-fm` come with man pages. It would be nice to provide these, such that `man ipython` will work.

**Describe the solution you'd like**

Much like we ask users to ensure ~/.local/bin is on their PATH and we manage symlinks there. We could manage symlinks to man pages in a location and ask them to add a MANDATORY_MANPATH or MANPATH_MAP line to their .manpath file.

**Describe alternatives you've considered**

I tried globbing ~/.local/pipx/venvs/*/share/man in my man config. But didn't succeed. Adding a line for each installed package does work as expected, so that's why I think automating the process would be ideal.

I'm willing and able to implement this. I just wanted to open discussion first instead of immediately coming with an elaborate PR.

brettviren:
The lack of accessible man pages was my first stumble when trying `pipx` for the first time just now.  In my case I'm testing with the `glances` package. And, yes, I just listened to the TPTM podcast. :smile: 

Well, it's a long walk for a short command but I did this:

```
❯ source .local/pipx/venvs/glances/bin/activate.fish
❯ man glances
❯ deactivate
```

In addition to finding and symlinking man pages and other types of files, perhaps the above dance could be wrapped into a general purpose `exec` subcommand to `pipx`.  Eg to enable something like this hypothetical command:
```
pipx exec glances man glances
```

uranusjr:
Unfortunately there is not a standard way to find man pages from Python packages (or any non-system package manager, as far as I am aware). This would require a standard, and a big push to make packages adhere to it.

alicederyn:
That sounds like a great standard to have.

Quote from related issue https://github.com/pypa/packaging-problems/issues/72

> This is hard is because pip does not focus on trying to be able to correctly handle installing packages on a system level.

Given that pipx *does* focus on this, perhaps this is the right project to be building up a grassroots standard? Two options immediately spring to mind:

1. Grab any shipped data file with the correct extension and path
2. Add a custom pipx section to the package metadata

I'm sure both of these have significant downsides that will be obvious to people who know packaging better than me (e.g. 1 seems like it might have false positives, 2 relies on the metadata file being extensible which I can't see any indication about either way from the PEP)

uranusjr:
[...]

PR #1047, opened 2023-08-24
Support for exposing manual pages

peterkuma:
This is a pull request implementing issue #857 (Handling manpages). It implements a similar function as symlinking of apps in the local bin directory, but for unix manual pages.

- Can install and uninstall a simple package with manual pages, with our without `--include-deps`.
- Passes tests-3.11.

The changes can be tested with Python packages which install manual pages in `share/man/man<i>`. For example sympy or yt-dlp.

## Summary of changes

This pull request modifies several commands and some other modules. It closely mirrors existing code for app symlinking but for manual pages (where it makes sense). It adds a new environment variable PIPX_MAN_DIR, in analogy to PIPX_BIN_DIR. Some functions implementing app symlinking are generalized to support manual page symlinking as well.

On Linux and macOS it exposes manual pages in `$HOME/.local/share/man/man<i>`.

On Windows it exposes manual pages in `C:\Users\<user>\.local\share\man`, even though Windows does not have any manual page reader in its default installation. I am not sure if it makes sense to expose the manual pages on this OS at all.

## Test plan

- Tested on Devuan GNU/Linux 5 with the default system Python (3.11.2-1+b1)
- Tested on macOS 13.5.1 (x64) with the default system Python (3.9.6)
- Tested on Windows 11 (x64) with the official Python distribution (3.11.5)

Tested by running
```
$ pipx install sympy
  installed package sympy 1.12, installed using Python 3.11.2
    - isympy
  These manual pages are now globally available
    - man1/isympy.1
done! ✨ 🌟 ✨
$ man isympy
$ pipx uninstall sympy

$ pipx install yt-dlp
  installed package yt-dlp 2023.9.24, installed using Python 3.11.2
    - yt-dlp
  These manual pages are now globally available
    - man1/yt-dlp.1
done! ✨ 🌟 ✨
$ man yt-dlp
$ pipx uninstall yt-dlp
```


src/pipx/commands/common.py
```diff
@@ -222,6 +261,20 @@ def get_venv_summary(
...
+    for i in range(1, 10):
+        man_section = "man%d" % i
+        exposed_man_paths += get_exposed_man_paths_for_package(
+            venv.man_path / man_section,
+            constants.LOCAL_MAN_DIR / man_section,
+            man_pages,
+        )
+    exposed_man_pages = sorted(
+        str(Path(p.parent.name) / p.name) for p in exposed_man_paths
+    )
```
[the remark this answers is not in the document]
peterkuma: It gets slightly comlicated because `get_exposed_man_paths_for_package` is called from multiple places, among which is `_get_venv_resource_paths`. `_get_venv_resource_paths` is currently called for each `man<i>` directory separately. The reason is because it checks if a symlink can be created in the directory (`can_symlink(local_resource_dir)`), which shouldn't be done for the main man directory (`~/.local/share/man`), but for each `man<i>` subdirectory. `_get_venv_resource_paths` then calls `get_exposed_man_paths_for_package` indirectly through `_get_package_man_paths`. If the man_section loop is located in `get_exposed_man_paths_for_package`, there will have to be another man_section loop in `_get_venv_resource_paths` for the `can_symlink` checks. It is of course possible, but there isn't a neat and simple way of getting rid of the loops.

Another option how to make it slightly nicer would be to add a constant such as `MAN_DIRS = ["man%d" % i for i in range(1, 10)]` in `constants.py` and then loop over the list in other places.

src/pipx/venv_inspect.py
```diff
@@ -99,19 +107,31 @@ def get_apps(dist: metadata.Distribution, bin_path: Path) -> List[str]:
...
+
+
+def get_man_pages(dist: metadata.Distribution, man_path: Path) -> List[str]:
+    man_pages = []
+    for i in range(1, 10):
+        man_section = "man%d" % i
+        names = get_resources("man", dist, man_path / man_section)
+        for name in names:
+            man_pages += [str(Path(man_section) / name)]
+    return man_pages
```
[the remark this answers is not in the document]
peterkuma: OK, I have changed that in the latest commit (49b13fb40c9057b5f67aa15d6901d8a7b22b0399). Also I have created a new constant `MAN_SECTIONS` and changed the loops to iterate over `MAN_SECTIONS` instead of the more complicated way with `range(1, 10)`.

chrysle:
This looks good to me. Could you fix all the tests and mark this as ready for review?

peterkuma:
tests-3.11 passes now, but ubuntu-latest 3.7, 3.8 and windows-latest 3.11 fail. I will fix them soon.

peterkuma:
Now it passes all of the tests in the GitHub workflow.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - Add `pipx install --preinstall` to support preinstalling build requirements
 - Pass `--no-input` to pip when output is not piped to parent stdout
 - Fix program name in generated manual page
+- Expose package manual pages
```
uranusjr, line 18:
```suggestion
- Expose manual pages included in an application installed with `pipx install`
```

Also we’d want to add some documentation on where a package should put the man files for pipx to find. Python does not actually have a standard on this, and current packages mostly do this by convention, so we need to describe the logic a bit for package maintainers to follow.

peterkuma: I have added some documentation in bd6c65e.

The convention comes from the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) and the [Filesystem Hierarchy Standard](https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#usrsharemanManualPages).

uranusjr:
Looks good in general. Some tests on the main feature would be nice.

src/pipx/main.py
```diff
@@ -68,6 +68,7 @@ def prog_name() -> str:
 
     Virtual Environment location is {str(constants.PIPX_LOCAL_VENVS)}.
     Symlinks to apps are placed in {str(constants.LOCAL_BIN_DIR)}.
+    Symlinks to manual are placed in {str(constants.LOCAL_MAN_DIR)}.
```
chrysle, line 71:
```suggestion
    Symlinks to manual pages are placed in {str(constants.LOCAL_MAN_DIR)}.
```

src/pipx/main.py
```diff
@@ -90,7 +92,10 @@ def prog_name() -> str:
     The install command is the preferred way to globally install apps
     from python packages on your system. It creates an isolated virtual
     environment for the package, then ensures the package's apps are
-    accessible on your $PATH.
+    accessible on your $PATH, and package's manual pages installed in
+    share/man/man[1-9] can be viewed with man on operating system where
+    it is available and the path in the environment variable `PIPX_MAN_DIR`
+    (default: {constants.DEFAULT_PIPX_MAN_DIR}) is in the man search path.
```
chrysle, line 98:
```suggestion
    accessible on your $PATH. Package's manual pages installed in
    share/man/man[1-9] can be viewed with man on an operating system where
    it is available and the path in the environment variable `PIPX_MAN_DIR`
    (default: {constants.DEFAULT_PIPX_MAN_DIR}) is in the man search path ($MANPATH).
```

docs/how-pipx-works.md
```diff
@@ -46,6 +48,16 @@ setup(
 
 In this case `main_func` and `some_func` would be available to pipx after installing the above example package.
 
+To install manual pages, which can be viewed with the `man` command on operating systems which have this command, include a `data_files` section in your `setup.py` file.
```
chrysle, line 51:
```suggestion
To install manual pages, which can be viewed with the `man` command on operating systems which have this command, include a [`data_files` section](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#data-files) in your `setup.py` file.
```

docs/how-pipx-works.md
```diff
@@ -8,7 +8,9 @@ When installing a package and its binaries on linux (`pipx install package`) pip
 - create a Virtual Environment in `~/.local/share/pipx/venvs/PACKAGE` that uses the shared pip mentioned above but otherwise is isolated (pipx uses a [.pth file]( https://docs.python.org/3/library/site.html) to do this)
 - install the desired package in the Virtual Environment
 - expose binaries at `~/.local/bin` that point to new binaries in `~/.local/share/pipx/venvs/PACKAGE/bin` (such as `~/.local/bin/black` -> `~/.local/share/pipx/venvs/black/bin/black`)
-- As long as `~/.local/bin/` is on your PATH, you can now invoke the new binaries globally
+- expose manual pages at `~/.local/share/man/man[1-9]` that point to new manual pages in `~/.local/pipx/venvs/PACKAGE/share/man/man[1-9]`
+- as long as `~/.local/bin/` is on your PATH, you can now invoke the new binaries globally
+- on operating systems which have the man command, as long as `~/.local/share/man` is a recognized search path of man, you can now view the new manual pages globally
```
chrysle, line 13:
```suggestion
- on operating systems which have the `man` command, as long as `~/.local/share/man` is a recognized search path of man, you can now view the new manual pages globally
```

chrysle (commented):
A few wording suggestions.

peterkuma:
This pull request  introduces a few new attributes in the metadata file (`man_pages`, `man_pages_of_dependencies`, `man_paths` and `man_paths_of_dependencies`). Does the `pipx_metadata_version` need to be increased from 0.2 to 0.3?

cs01:
> @uranusjr
> 
> > Looks good in general. Some tests on the main feature would be nice.
> 
> The package pycowsay is used quite extensively in the tests. It would be helpful with the testing if pycowsay also installs a manual page. Do you think it would make sense to ask the package maintainer if they would be willing to add a manual page to the package? I see that the maintainer (@cs01) is also a developer of pipx.

I made pycowsay specifically as a simple demo package for pipx. I am up for adding to it, or giving pipx maintainers access to merge pull requests to pycowsay.

peterkuma:
> This pull request introduces a few new attributes in the metadata file (`man_pages`, `man_pages_of_dependencies`, `man_paths` and `man_paths_of_dependencies`). Does the `pipx_metadata_version` need to be increased from 0.2 to 0.3?

I have increased the version number to 0.3.

Other updates:

- The tests now use pycowsay 0.0.0.2.
- The tests now correctly set `LOCAL_MAN_DIR`, so that manual pages are installed within the virtual environment and not `~/.local`.
- Package metadata are now correctly tested for manual page related attributes for packages for which it is applicable.
- Test package information now contains information about manual pages for all packages which install manual pages.
- The environment variable `PIPX_MAN_DIR` is now tested.
- `test_install_all_packages` now tests if manual pages are installed according to the package information.

What remains to be done is updating a few more tests to test for manual pages and related stuff.

The standard suites of tests pass on all of the OSes (as run on GitHub Actions). Some of the install `--all-packages` tests fail, but that seems to be related to the fact that the package dependencies have been updated and produce a different set of apps than assumed in the test package information.

dukecat0:
Hi @peterkuma, when I was testing the change locally, it seemed like the man page was not handled correctly during the uninstall process:
```
$ pipx install sympy
...
$ pipx uninstall sympy
...
$ pipx install sympy
⚠️  Note: isympy was already on your PATH at /home/codespace/.local/bin/isympy
⚠️  File exists at /home/codespace/.local/share/man/man1/isympy.1 and points to /home/codespace/.local/share/man/man1/isympy.1, not
    /usr/local/py-utils/venvs/sympy/share/man/man1/isympy.1. Not modifying.
  installed package sympy 1.12, installed using Python 3.10.8
    - isympy
  These manual pages are now globally available
    - man1/isympy.1 (symlink missing or pointing to unexpected location)
done! ✨ 🌟 ✨
```

peterkuma:
@dukecat0 

What environment is this on (OS, or anything else specific)? The commands you posted work fine for me on Debian.

peterkuma:
@uranusjr  @chrysle  How do you deal with package specifications in `tests/package_info.py`? Installation of many of the packages was failing even in the main branch, which I suppose is due to changes in some of the dependencies. I updated the affected package specifications, but then I realised that `testdata/tests_packages/*.txt` would also need to be updated, and that seems like it would be quite complicated for so many combinations of OSes and Python versions. Alternatively I could keep the same package versions and just update the lists of apps and app paths. Shall I be concerned about the "slow" tests failing?

dukecat0:
> @dukecat0
> 
> What environment is this on (OS, or anything else specific)? The commands you posted work fine for me on Debian.

It's Github codespace, and Github has set the environment variable `PIPX_HOME` by default. I guess this error will occur for all users who have set `PIPX_HOME` to other locations

I just made another test on macOS without setting `PIPX_HOME`, and everything works well.

dukecat0:
> @uranusjr @chrysle How do you deal with package specifications in `tests/package_info.py`? Installation of many of the packages was failing even in the main branch, which I suppose is due to changes in some of the dependencies. I updated the affected package specifications, but then I realised that `testdata/tests_packages/*.txt` would also need to be updated, and that seems like it would be quite complicated for so many combinations of OSes and Python versions. Alternatively I could keep the same package versions and just update the lists of apps and app paths. Shall I be concerned about the "slow" tests failing?

We use a Github Action to generate those lists. You can trigger it from https://github.com/peterkuma/pipx/actions/workflows/create_tests_package_lists.yml

chrysle:
I find it easier to just use the `nox` workflow `create_test_package_list`.

peterkuma:
> > @dukecat0
> > What environment is this on (OS, or anything else specific)? The commands you posted work fine for me on Debian.
> 
> It's Github codespace, and Github has set the environment variable `PIPX_HOME` by default. I guess this error will occur for all users who have set `PIPX_HOME` to other locations
> 
> I just made another test on macOS without setting `PIPX_HOME`, and everything works well.

The reason for this appears to be that sympy is already installed in the default GitHub Codespaces image, i.e. the files `.local/bin/isympy` and `.local/share/man/man1/isympy.1` already exist in the system. Using a custom `PIPX_HOME` works fine on my local Debian.

dukecat0:
> The reason for this appears to be that sympy is already installed in the default GitHub Codespaces image, i.e. the files `.local/bin/isympy` and `.local/share/man/man1/isympy.1` already exist in the system. Using a custom PIPX_HOME works fine on my local Debian.

Thanks for the investigation. I didn't notice this before. 😅

gaborbernat (changes requested):
Needs merge conflicts fixed.

gaborbernat (approved):

PR #1049, opened 2023-08-26
Fix program name in generated manual page

jbrobst:

## Summary of changes

243c631ceafc089e2485cd97655cb44d4690232c caused `python scripts/generate_man.py` (e.g. via `nox -s build_man`) to produce a manual page with `generate_man.py` as the program name. Fix this by changing `sys.argv[0]` to `pipx` before creating the parser.

## Test plan

AFAICT `scripts/generate_man.py` is not currently tested.


dukecat0:
Thanks!

PR #1071, opened 2023-10-02
Manually parse URL without depending on packaging

### LINKED ISSUE #1070: Local path is treated as version due to packaging feature change
**Describe the bug**
Today I reinstall pipx and noticed that when I was installing a local package I was constantly getting `⚠️  Ignoring --editable install option. pipx disallows it for anything but a local path, to avoid having to create a new src/ directory.`

**How to reproduce**

```bash
# Install latest pipx

cd ~/dev/my-python-project
pipx install --editable --force .
# Gets message about --editable being ignore
```

**Expected behavior**
`--editable` should not be ignore.

**Investigation**
After some debugging the issue seems to be coming from this commit from packaging https://github.com/pypa/packaging/pull/684 where they removed URL validation. Because there is no longer an `InvalidRequirement` being thrown, in the following block the else branch gets executed which erroneously sets `valid_url` as '.'

https://github.com/pypa/pipx/blob/main/src/pipx/package_specifier.py#L80-L87

Because valid_url is set to something, valid_local_path is no longer being set and therefore pipx thinks `.` is not a local package and therefore ignores the --editable flag.
https://github.com/pypa/pipx/blob/main/src/pipx/package_specifier.py#L89-L95

`packaging` released this change on 23.2 which was releases Oct1. Reverting packaging to 23.1 fixes the issue.

uranusjr:
Fix #1070.

New packaging removed URL validation, so we implement our own. To accomodate the new packaging validation (or lack thereof), the valid URL rule is relaxed a bit; unknown schemes are now allowed (and will error later when we try to fetch them).

I didn’t add an entry to the change log since this shouldn’t affect the functionality. If anything does change, it should be considered a bug.

### DIFF
--- a/src/pipx/package_specifier.py
+++ b/src/pipx/package_specifier.py
@@ -8,6 +8,7 @@
 
 import logging
 import re
+import urllib.parse
 from pathlib import Path
 from typing import List, NamedTuple, Optional, Set, Tuple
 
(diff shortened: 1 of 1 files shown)


dukecat0 (approved):

uranusjr:
I’ll merge this into main so this can be more easily tested (with Homebrew for example), but anyone with this issue is very welcomed to test code in main and give feedback please!

PR #1075, opened 2023-10-05
Exclude known incompatible userpath 1.9.0

### LINKED ISSUE #1025: Please fix functional regression for non-login shells

uranusjr:
Close #1034. Close #1025.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ dependencies = [
   "importlib-metadata>=3.3.0; python_version < '3.8'",
   "packaging>=20.0",
   "platformdirs>=2.1.0",
-  "userpath>=1.6.0",
+  "userpath>=1.6.0,!=1.9.0",
 ]
 dynamic = ["version"]
 


deronnax:
a comment next to the exclusion to explain why a version is excluded would have been a great help 🙃 (I had to use a round of Github blame to come here).

PR #1078, opened 2023-10-08
Drop `wheel` and `setuptools` from shared libs

### LINKED ISSUE #951: `pipx run wheel` / `install wheel` doesn't work

### LINKED ISSUE #990: drop `setuptools` and `wheel` from shared libs
this would fix https://github.com/pypa/pipx/issues/951 and cause pip 23.1 to use the pep517 codepath for all packages

since 23.1 pip doesn't need wheel or setuptools anymore https://pip.pypa.io/en/stable/news/#v23-1

> When the wheel package is not installed, pip now uses the default build backend instead of setup.py install and setup.py develop for project without pyproject.toml. ([#8559](https://github.com/pypa/pip/issues/8559))

setuptools was also removed from CPython 3.12's ensurepip: https://docs.python.org/3.12/whatsnew/3.12.html#removed
> Remove the bundled setuptools wheel from [ensurepip](https://docs.python.org/3.12/library/ensurepip.html#module-ensurepip), and stop installing setuptools in environments created by [venv](https://docs.python.org/3.12/library/venv.html#module-venv).

chrysle:
And enable running and installing the remaining shared library `pip` using `pipx run`. I have transferred the hack from #970 per https://github.com/pypa/pipx/issues/951#issuecomment-1564513524 for use with `pip`.

 * Close #970 
 * Close #951 
 * Close #990

## Test plan

Tested by running
```
pipx run wheel
pipx run pip
pipx install pip
```


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## dev
 
+- Drop `setuptools` and `wheel` from the shared libraries
```
uranusjr, line 3: It may be a good idea to describe briefly what this means to end users.

chrysle: Sure, done. Have you already considered adopting `towncrier` for `pipx`? This would also simplify categorizing the changes.

gaborbernat (changes requested):
CI is failing.

src/pipx/shared_libs.py
```diff
@@ -107,8 +107,6 @@ def upgrade(
                         *_pip_args,
                         "--upgrade",
                         "pip",
```
webknjaz, line 109: This should set the minimum version of pip that doesn't need setuptools/wheel:
```suggestion
                         "pip >= 23.1",
```

src/pipx/commands/run.py
```diff
@@ -227,6 +226,12 @@ def _download_and_run(
             package_or_url, python, pip_args=pip_args, verbose=verbose
         )
 
+    override_shared = False
+
+    if package_name == "pip":
+        override_shared = True
```
webknjaz, line 232: Simplify?
```suggestion
    override_shared = package_name == "pip"
```

src/pipx/commands/install.py
```diff
@@ -58,7 +58,12 @@ def install(
             return EXIT_CODE_INSTALL_VENV_EXISTS
 
     try:
-        venv.create_venv(venv_args, pip_args)
+        # Enable installing shared library `pip` with `pipx`
+        override_shared = False
+
+        if package_name == "pip":
+            override_shared = True
```
webknjaz, line 65:
```suggestion
        override_shared = package_name == "pip"
```

webknjaz (commented):
Got a few small suggestions..

chrysle:
Thank you!

gaborbernat:
We need to fix the CI first 😎

dukecat0:
> We need to fix the CI first 😎

It should be fixed in #1117.

dukecat0:
@chrysle Please fix linting.

chrysle:
Done!

gaborbernat (approved):

dukecat0:
Thanks!

jaraco:
This change did not have the intended effect. Instead, it had the effect of leaving an old setuptools lingering in the shared libs, because that's what `venv` does:

```
root@1b9c9a77eeb4:/# python3.10 -m venv .venv
root@1b9c9a77eeb4:/# ls .venv/lib/python3.10/site-packages/
_distutils_hack  distutils-precedence.pth  pip  pip-22.0.2.dist-info  pkg_resources  setuptools  setuptools-59.6.0.dist-info
```

See #1539 for why that's worse than upgrading it.

Issue #629, opened 2021-02-13
Change VenvInspectInformation.distributions to be type `Iterable[metadata.Distribution]`


itsayellow:
**How would this feature be useful?**

Change `VenvInspectInformation.distributions` in `venv_inspect.py` to be type `Iterable[metadata.Distribution]`.  (Or whatever type `importlib.metadata.distributions()` returns.)

So we don't have to cast it to a `list()` which is unnecessary.

aanghelidi:
Hello @dukecat0 , I would like to be assigned to this one, can I work on it?

dukecat0:
Hello @aanghelidi, please feel free to work on it!

aanghelidi:
Hello @dukecat0 , here is my attempt to solve this issue, let me know what you think about it.

PR #1087, opened 2023-10-15
Relax VenvInspectInformation attribute typing

aanghelidi:
## Description

This PR aims to remove the explicit call to `list()` made in `venv_inspect.py` by using the type that `importlib.metadata.distributions()` returns (an `Iterator` from `itertools.chain.from_iterable`, [see](https://github.com/python/cpython/blob/3.11/Lib/importlib/metadata/__init__.py#L581)) from wrapping `Distribution.discover(**kwargs)`.

However, the code logic would be flawed if we used an `Iterator` (in this case skipping the explicit `list()` call) because we could only iterate on it once and need to do it multiple times in `_dfs_package_apps` to retrieve the dependencies for a given package correctly. 

Therefore, an `Iterable` was used as a more generic type hint for more flexibility, and `tuple` was used instead of `list` since we are not modifying `VenvInspectInformation.distributions`.

## Summary of changes

* Import Iterable from typing
* Replace all the corresponding type hints
* Swap the explicit list() call to a tuple() call

## Test plan
Tested by running
```
nox -s lint
nox -s tests-3.11
```
Resolves: #629

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## dev
 
+- Use `Iterable[metadata.Distribution]` instead of `List[metadata.Distribution]` in `venv_inspect.py` (#629)
 - Make usage message in `pipx run` show `package_or_url`, so extra will be printed out as well
 - Add `--force-reinstall` to pip arguments when `--force` was passed
 - Use the py launcher, if available, to select Python version with the `--python` option
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,5 @@
 ## dev
 
-- Use `Iterable[metadata.Distribution]` instead of `List[metadata.Distribution]` in `venv_inspect.py` (#629)
 - Make usage message in `pipx run` show `package_or_url`, so extra will be printed out as well
 - Add `--force-reinstall` to pip arguments when `--force` was passed
 - Use the py launcher, if available, to select Python version with the `--python` option
--- a/src/pipx/venv_inspect.py
+++ b/src/pipx/venv_inspect.py
@@ -1,9 +1,14 @@
 import json
 import logging
+import sys
 import textwrap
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## dev
 
+- Use `Iterable[metadata.Distribution]` instead of `List[metadata.Distribution]` in `venv_inspect.py` (#629)
```
dukecat0, line 3: We don't need a changelog here since this change has no effect on end users.

aanghelidi: Thank you for the suggestion @dukecat0 , the changelog entry has been removed in the last commit :+1:

uranusjr:
I would prefer Collection here, since Iterable sort of hints you are only expected to interate through this at most once (which is not the case).

aanghelidi:
Thank you for your feedback @uranusjr :+1:  The last commit reflects this change.

uranusjr (approved):

dukecat0:
Looks like some tests are failed with this error: `TypeError: 'ABCMeta' object is not subscriptable`

aanghelidi:
As described [here](https://docs.python.org/3/library/typing.html#typing.Collection) `collections.abc.Collection` can only be subscripted since `3.9`, hence this error. The last commit should fix this issue. Let me know what you think.

uranusjr:
Doing a `sys.version_info` check just for this seems a bit redundant. I’d just use the typing entry for now, or add a `from __future__ import annotations` to make the annotation not resolve at runtime instead.

aanghelidi:
I added `typing.Collection` as I'm less familiar with `from __future__ import annotations`.

uranusjr (approved):

dukecat0:
Thanks!

PR #1099, opened 2023-11-03
tests: fix regex to include build numbers

henryiii:
This fixes the regex in the case where a wheel has a build number (as black did on macOS).

* I have added an entry to `docs/changelog.md` (tests only)

## Test plan

Tested by running
```
nox -s create_test_package_list-3.7
```

Side note: why is the macOS version in the filename? Could it be removed? This is making a lot of different files for different macOS versions, and is why this wasn't in in CI yet, but I hit it, since I was making new ones since mac23 didn't exist.

### DIFF
--- a/scripts/list_test_packages.py
+++ b/scripts/list_test_packages.py
@@ -130,7 +130,7 @@ def create_test_packages_list(
     all_packages = []
     for downloaded_filename in downloaded_list:
         wheel_re = re.search(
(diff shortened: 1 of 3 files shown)


dukecat0:
> why is the macOS version in the filename? 

Probably it's because of this: https://github.com/pypa/pipx/pull/686#issuecomment-841719975

gaborbernat (approved):

PR #1100, opened 2023-11-03
feat: support PEP 723 run requirements

henryiii:

## Summary of changes

Moved the unreleased feature from the rejected PEP 722 syntax to the (provisionally accepted a few moments ago in https://github.com/python/peps/pull/3505) PEP 723 syntax.

## Test plan

Tested by running
```
nox -s tests-3.11
```


pradyunsg:
:wave: I would request that y'all hold off on merging this until there's agreement on what exactly the provisional status for the PEP means, in the context of user-facing tooling implementing the described functionality.

I've [asked about what tooling is supposed to do right now](https://discuss.python.org/t/pep-722-723-decision/36763/94?u=pradyunsg) in the acceptance thread for the PEP, because what they're supposed to do seems to be ambigous/unclear to me, and I think we should make the call explicitly to implement something that we might rip out later.

src/pipx/commands/run.py
```diff
@@ -319,41 +326,45 @@ def _http_get_request(url: str) -> str:
         raise PipxError(str(e)) from e
 
 
+PEP723 = re.compile(
+    r"(?m)^# /// (?P<type>[a-zA-Z0-9-]+)$\s(?P<content>(^#(| .*)$\s)+)^# ///$"
+)
```
uranusjr, line 331: A comment on the regex pattern (and/or maybe using `re.VERBOSE`) would be a good idea

src/pipx/commands/run.py
```diff
@@ -319,41 +326,45 @@ def _http_get_request(url: str) -> str:
...
-    else:
-        # No "Requirements:" line in the file
+    """
+    Supports PEP 723.
+    """
+
+    name = "pyproject"
+
+    # Windows is currently getting un-normalized line endings, so normalize
+    content = content.replace("\r\n", "\n")
```
uranusjr, line 342: Is this needed? From what I can tell not normalising would simply produce some harmless empty lines that do not match anyway.

src/pipx/commands/run.py
henryiii: I'd say comment as long as we match the PEP's suggested regex.

henryiii:
> explicitly to implement something that we might rip out later.

To be clear, PEP 722 (rejected) is already implemented in main, and this is just moving that unreleased code to PEP 723 (provisionally accepted). IMO, if this exists at all, it better in the provisionally accepted form rather than the rejected form.

src/pipx/commands/run.py
henryiii: (The CI was breaking on Windows otherwise ;) )

gaborbernat:
> To be clear, PEP 722 (rejected) is already implemented in main, and this is just moving that unreleased code to PEP 723 (provisionally accepted). IMO, if this exists at all, it better in the provisionally accepted form rather than the rejected form.

Merged this because I agree here more with @henryiii than the concerns raised by Pradyun.

robinshen36:
@algora-pbc attempt

Been looking for something like this to work on.

Recently finished up something pretty similar.

My approach would be to start by understanding the inject, then iterate from there.

Around $170-$290 seems fair to me, but open to whatever works.

Let me know if this sounds good!

PR #1102, opened 2023-11-04
Allow installing multiple applications at once

### LINKED ISSUE #971: Allow installing multiple applications with `install` command
Instead of:
```
$pip = @("internetarchive", "git+https://github.com/arecarn/dploy.git", "tubeup")
foreach ($p in $pip) {pipx install $p}
```
I will be able to do just `pipx install internetarchive git+https://github.com/arecarn/dploy.git tubeup`

Allow installing multiple apps with `pipx install`

Using foreach or something like this.

sam-sw:

## Summary of changes
As a response to issue #971, the `commands.install` function is modified to take lists of package names/specs as arguments, and iterate through the lists, installing each application separately. The associated CLI functions and help strings are updated to reflect this. The `commands.reinstall` function is also modified to ensure it complies with the new argument requirements of `commands.install`.  Two tests have been added (replicating `test_install_easy_packages` and `test_install_tricky_packages`, but doing multiple packages at once).

## Test plan

Tested by manually running
```
pipx install pycowsay black
```
and also with the new unit tests.

Resolves #971


sam-sw:
This is my take on responding to #971, happy to receive comments if a different approach may work better.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 - Fix program name in generated manual page
 - Print all environment variables in `pipx environment`
 - Return an error message when directory can't be added to PATH successfully
+- Add ability to install multiple packages at once with `pipx install one two three ...`
```
chrysle, line 20:
```suggestion
- Add ability to install multiple packages at once
```

chrysle (commented):
This is an approach for creating a separate virtual environment for each of the packages.
Should an option be added to install all into the same one?

Also, I'd think one should also be able to remove and upgrade multiple packages. But that's out of the scope of this PR.

sam-sw:
I agree with your comment about removing and upgrading multiple packages.

For installing multiple packages to the same virtual environment, would the current approach be to install one, and then inject the others, using the `--include-apps` option to ensure all applications are on the `PATH`? This is definitely more involved for users than a separate option on the install command, but then again the main reason for using pipx is to separate applications into unique virtual environments.

Personally I would lean away from having the extra install option, to keep things simple.

uranusjr:
> Should an option be added to install all into the same one?

This is already avaiable with `--preinstall`. It makes more sense to have `--postinstall` if needed.

gaborbernat:

sam-sw:
Apologies for the delay, merged latest changes from the main branch now and fixed any conflicts etc.

gaborbernat:
Seems you still need to fix some tests. 🤔

sam-sw:
I've just re-ran `nox` locally and tests seem to be passing, except Python 3.9. But that also fails on the main branch as well?

gaborbernat:
No failure on main https://github.com/pypa/pipx/actions/runs/7219455002

gaborbernat:
Push a new commit to trigger a new CI run.

sam-sw:
Looks like 3.8 and 3.9 fail because the `strict=True` argument in `zip` isn't implemented for those versions. Using `zip` this way was to satisfy the pre-commit check for code linting, what would you recommend?

src/pipx/commands/install.py
```diff
@@ -33,73 +33,86 @@ def install(
...
+        package_names = [
+            package_name_from_spec(package_spec, python, pip_args=pip_args, verbose=verbose)
+            for package_spec in package_specs
+        ]
 
-    try:
-        exists = venv_dir.exists() and bool(next(venv_dir.iterdir()))
-    except StopIteration:
-        exists = False
+    for package_name, package_spec in zip(package_names, package_specs, strict=True):
```
chrysle, line 47: Probably telling the linter not to bother about a missing `strict` argument until we have dropped these Python versions is enough here?

Don't know what error code this generates, but you could write a `# noqa: EXYZ`.

sam-sw: Thanks for your help! Just removing the `strict=True` and re-running the linter without any cached nox environments etc. seems to now pass, without needing to use any `# noqa` for some reason?

Either way, linting and tests (including on 3.8 etc.) seem to now pass locally.

src/pipx/commands/install.py
```diff
@@ -33,73 +33,86 @@ def install(
...
-        package_name = package_name_from_spec(package_spec, python, pip_args=pip_args, verbose=verbose)
-    if venv_dir is None:
-        venv_container = VenvContainer(constants.PIPX_LOCAL_VENVS)
-        venv_dir = venv_container.get_venv_dir(f"{package_name}{suffix}")
+    if package_names is None:
+        package_names = [
+            package_name_from_spec(package_spec, python, pip_args=pip_args, verbose=verbose)
+            for package_spec in package_specs
+        ]
+    elif len(package_names) != len(package_specs):
```
chrysle, line 41: This seems to hold exactly the same logic as the condition above. Could you merge the two?

tests/test_install.py
```diff
@@ -39,6 +39,26 @@ def install_package(capsys, pipx_temp_env, caplog, package, package_name=""):
     assert "WARNING" not in caplog.text
 
 
+def install_multiple_packages(capsys, pipx_temp_env, caplog, packages, package_names=()):
```
chrysle, line 42: Let's not copy too much, and create a unified `install_packages()` function.

sam-sw: Done this, but left the existing tests that use `parametrize` to run the install commands separately, so that there's still some level of testing installing single and multiple packages at once.

chrysle (approved):
LGTM!

gaborbernat (approved):

Issue #1091, opened 2023-10-20
pipx crashes when a file named logging.py is in the current directory


jamescooke:
**Describe the bug**

When a file called `logging.py` exists in the directory that `pipx` is invoked from, `pipx` crashes.

**How to reproduce**

On my Ubuntu 20.04 LTS machine:

```sh
cd $(mktemp -d)
pipx --version
```

```
1.2.0
```

Create a `logging.py` file:

```sh
cat > logging.py
print('hello world')
^d
```

Run pipx:

```sh
pipx --version
```

"hello world" appears, then a crash:

```
hello world
Traceback (most recent call last):
  File "/home/james/.local/bin/pipx", line 5, in <module>
    from pipx.main import cli
  File "/home/james/.local/lib/python3.8/site-packages/pipx/main.py", line 7, in <module>
    import logging.config
ModuleNotFoundError: No module named 'logging.config'; 'logging' is not a package
Error in sys.excepthook:
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/apport_python_hook.py", line 72, in apport_excepthook
    from apport.fileutils import likely_packaged, get_recent_crashes
  File "/usr/lib/python3/dist-packages/apport/__init__.py", line 5, in <module>
    from apport.report import Report
  File "/usr/lib/python3/dist-packages/apport/report.py", line 32, in <module>
    import apport.fileutils
  File "/usr/lib/python3/dist-packages/apport/fileutils.py", line 12, in <module>
    import os, glob, subprocess, os.path, time, pwd, sys, requests_unixsocket
  File "/usr/lib/python3/dist-packages/requests_unixsocket/__init__.py", line 1, in <module>
    import requests
  File "/usr/lib/python3/dist-packages/requests/__init__.py", line 43, in <module>
    import urllib3
  File "/usr/lib/python3/dist-packages/urllib3/__init__.py", line 7, in <module>
    from .connectionpool import HTTPConnectionPool, HTTPSConnectionPool, connection_from_url
  File "/usr/lib/python3/dist-packages/urllib3/connectionpool.py", line 29, in <module>
    from .connection import (
  File "/usr/lib/python3/dist-packages/urllib3/connection.py", line 41, in <module>
    from .util.ssl_ import (
  File "/usr/lib/python3/dist-packages/urllib3/util/__init__.py", line 20, in <module>
    from .retry import Retry
  File "/usr/lib/python3/dist-packages/urllib3/util/retry.py", line 20, in <module>
[...]

PR #1115, opened 2023-12-01
Run venv Creation in Isolated Mode (#1091)

ehlewis:

## Summary of changes
Addressing #1091 
Created venv using [isolated mode](https://docs.python.org/3/using/cmdline.html#cmdoption-I) by passing the -I flag in order to resolve a bug where python files in the cwd could break the venv creation.
Also a slight security improvement since by ignoring the python files in the cwd which could potentially override files on the sys.path

## Test plan

Manually tested resolution of bug by creating a file in cwd 
`echo "blah blah" > logging.py`
then running the current release of pipx
`pipx install frogmouth`
we get the expected error
```  
File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/venv/__init__.py", line 7, in <module>
  import logging
  File "/Users/ehlewis/test/logging.py", line 1
    blah blah
         ^^^^
SyntaxError: invalid syntax

'/Library/Frameworks/Python.framework/Versions/3.12/bin/python3.12 -m venv --without-pip
/Users/ehlewis/.local/pipx/venvs/frogmouth' failed
```
running the same command with the modified `create_venv` shows a successful install and the installed package functions as expected
```
python3 src/pipx install frogmouth                                                 
  installed package frogmouth 0.9.2, installed using Python 3.12.0
    - frogmouth
done! ✨ 🌟 ✨
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Fix program name in generated manual page
 - Print all environment variables in `pipx environment`
 - Return an error message when directory can't be added to PATH successfully
+- Run venv creation in isolated mode with -I
 
 ## 1.2.1
 
--- a/src/pipx/venv.py
+++ b/src/pipx/venv.py
@@ -2,8 +2,6 @@ import json
 import logging
 import re
 import time
-import os
-import uuid
 from pathlib import Path
 from subprocess import CompletedProcess
 from typing import Dict, Generator, List, NoReturn, Optional, Set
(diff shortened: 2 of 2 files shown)


uranusjr:
The isolated mode turns off a lot more than just `sys.path`. I’m not strictly against this (it’s fundamentally a good idea), but we need to communicate the change very carefully with users, including a major release and visible warnings to help migration.

Issue #1122, opened 2023-12-02
Pipx 1.3.0 adds `--force-reinstall` in a wrong place when `--editable` installation is used


potiuk:
The `pipx 1.3.0` released ~1 hour ago started to fail our CI builds in Airflow 

Example here: https://github.com/apache/airflow/actions/runs/7071524725/job/19249611301?pr=36026

We are running:

```
python -m pipx install --editable ./dev/breeze/ --force
```

```
    ERROR: --force-reinstall is not a valid editable requirement. It should either be a path to a local project or a VCS URL (beginning with bzr+http, bzr+https, bzr+ssh, bzr+sftp, bzr+ftp, bzr+lp, bzr+file, git+http, git+https, git+ssh, git+git, git+file, hg+file, hg+http, hg+https, hg+ssh, hg+static-http, svn+ssh, svn+http, svn+https, svn+svn, svn+file).
⚠️  Not removing existing venv
    /home/runner/.local/pipx/venvs/apache-airflow-breeze because it was not
    created in this session
Error installing apache-airflow-breeze from spec
'/home/runner/actions-runner/_work/airflow/airflow/dev/breeze'.
```

It looks like the resuling `pip` installation is wrong after this change in the changelog of `pipx 1.3.0`:

> Add --force-reinstall to pip arguments when --force was passed

Likely the resulting command is:

```
pip install --editable --force-reinstall ./dev/breeze/
```

This is wrong becasue `--editable` expects URL/PATH right after it.

**Describe the bug**

**How to reproduce**

**Expected behavior**

PR #1124, opened 2023-12-02
Fix combining of `--editable` and `--force` flag

potiuk:

## Summary of changes

The 1.3.0 version of `pipx` (with #933) introduced a but where the `--force-reinstall` flag caused failure when also the `--editable` flag was used - because it was adding the `--force-reinstall` flag after `--editable` one (and `--editable` flag expects url/path to the Python package to install to follow it.

The change fixes it by adding the `--force-reinstall` flag at the beginning rather than at the end of arguments to avoid this kind of problem also in case other flags might have similar problem.

Fixes: #1122

## Test plan

Tested by running
```
pytest tests/test_install.py::test_force_install_changes_editable
```

Screenshot 2023-12-02 at 20 11 27

### DIFF
--- a/tests/test_install.py
+++ b/tests/test_install.py
@@ -13,6 +13,8 @@ from pipx import constants
 
 TEST_DATA_PATH = "./testdata/test_package_specifier"
 
+SOURCE_DIR = Path(__file__).parent.parent.resolve()
+
 
 def test_help_text(monkeypatch, capsys):
     mock_exit = mock.Mock(side_effect=ValueError("raised in test to exit early"))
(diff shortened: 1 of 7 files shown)


tests/test_install.py
```diff
@@ -13,6 +13,8 @@
 
 TEST_DATA_PATH = "./testdata/test_package_specifier"
 
+SOURCE_DIR = Path(__file__).parent.parent.resolve()
```
gaborbernat, line 16: Remove this and used the `root` fixture instead.

gaborbernat (changes requested):

tests/test_install.py
potiuk: Ah nice. TIL

potiuk: Fixed.

gaborbernat:
No yanking but https://github.com/pypa/pipx/actions/runs/7072334702

potiuk:
👍 🎉 🎉 🎉 🎉 🎉 👍

potiuk:
And released ! :)

PR #1127, opened 2023-12-02
Run Subprocesses in Destination dir

ehlewis:

## Summary of changes
Addressing #1091 
Set venv creation and install_package subprocesses to run inside the destination directory (self.root) in order to resolve a bug where python files in the cwd could break the venv creation.
Also a slight security improvement since by ignoring the python files in the cwd which could potentially override files on the sys.path

## Test plan

Manually tested resolution of bug by creating a file in cwd 
`echo "blah blah" > logging.py`
then running the current release of pipx
`pipx install frogmouth`
we get the expected error
```  
File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/venv/__init__.py", line 7, in <module>
  import logging
  File "/Users/ehlewis/test/logging.py", line 1
    blah blah
         ^^^^
SyntaxError: invalid syntax

'/Library/Frameworks/Python.framework/Versions/3.12/bin/python3.12 -m venv --without-pip
/Users/ehlewis/.local/pipx/venvs/frogmouth' failed
```
running the same command with the modified `create_venv` shows a successful install and the installed package functions as expected
```
python3 src/pipx install frogmouth                                                 
  installed package frogmouth 0.9.2, installed using Python 3.12.0
    - frogmouth
done! ✨ 🌟 ✨
```

Repo tests were also passed successfully
https://github.com/ehlewis/pipx/actions/runs/7072389139

### DIFF
--- a/src/pipx/util.py
+++ b/src/pipx/util.py
@@ -161,6 +161,7 @@ def run_subprocess(
     log_cmd_str: Optional[str] = None,
     log_stdout: bool = True,
     log_stderr: bool = True,
+    run_dir: Optional[str] = None,
 ) -> "subprocess.CompletedProcess[str]":
     """Run arbitrary command as subprocess, capturing stderr and stout"""
     env = dict(os.environ)
@@ -169,6 +170,8 @@ def run_subprocess(
     if log_cmd_str is None:
         log_cmd_str = " ".join(str(c) for c in cmd)
     logger.info(f"running {log_cmd_str}")
+    if run_dir:
+        os.makedirs(run_dir, exist_ok=True)
     # windows cannot take Path objects, only strings
     cmd_str_list = [str(c) for c in cmd]
     completed_process = subprocess.run(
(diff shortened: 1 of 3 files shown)


src/pipx/util.py
```diff
@@ -161,6 +161,7 @@ def run_subprocess(
     log_cmd_str: Optional[str] = None,
     log_stdout: bool = True,
     log_stderr: bool = True,
+    run_dir: Optional[str] = None,
```
webknjaz, line 164: Using `Optional` is considered confusing. There's even linters for making it `Union[$thing, None]`, which better represents what it means. With modern Pythons, it's more convenient to do this:
```suggestion
    run_dir: str | None = None,
```
Though, for supporting older versions, adding `from __future__ import annotations` is needed.

gaborbernat (changes requested):
Test please.

gaborbernat:

PR #1130, opened 2023-12-03
Pull version from VCS to simplify release and build dynamic docs via mkdocs plugin

gaborbernat:
This will simplify the release process to just creating a GitHub release, and also adds man page and zipapp testing to the core CI (not just release time operation). Furthermore, extends the type checker to consider type information from dependencies. Finally, also switches to trusted publish for PyPI.

image


chrysle (commented):
I think the branch name is incorrect, it should be `pull/1130/merge`. Could try using `{{ github.ref }}` in the workflow definition instead?

.github/workflows/tests.yml
```diff
@@ -41,12 +42,26 @@ jobs:
       - name: Install nox
         run: python -m pip install nox
       - name: Execute Tests
-        run: nox --non-interactive --session tests-${{ matrix.python-version }}
+        run: nox --error-on-missing-interpreters --non-interactive --session tests-${{ matrix.python-version }}
+  man:
+    name: Build mang page
```
dukecat0, line 47:
```suggestion
    name: Build man page
```

Issue #464, opened 2020-08-10
Check for latest version before running with `pipx run`


anthem-bmk:
**How would this feature be useful?**

When using `pipx run someapp`, a version of `someapp` is cached for 14 days. There could be new releases within that timeframe that could be useful to have.

**Describe the solution you'd like**

I think my ideal solution would be to just automatically check for a newer version and upgrade if it exists. If you're already on the latest version, use the cache.

The next best solution (which would avoid any backwards compatibility issues) would be `pipx run --latest someapp`, that would work the same way but require an explicit flag.

A third solution, that might work well with the second, is having some sort of config you can set (e.g. `~/.pipxrc`) that lets you say you always want the `--latest` flag.

**Describe alternatives you've considered**

The only real workaround I've found is to run the app a second time without cache, then a third time with the cache. This also only works if you know the version has changed and want to force it to change. For example:

```
pipx run someapp --version # v0.1.0
# some new update comes out
pipx run --no-cache someapp --version # v0.1.1
pipx run someapp --version # v0.1.1
```

Some downsides:
* If the app has a lot of dependencies, running `--no-cache` every time will be pretty inefficient.
* If you don't always run with `--no-cache`, you need to know somehow if there's a version update
* If you want to update once with `--no-cache`, it will delete the cache the first time and download the latest version, then the second time it'll download it again but actually cache it.

cs01:
I like this idea. 14 days is kind of a long time. I tried to land on a good balance between checking for updates too often (and slowing things down), and not long enough.

My main concern with the `--latest` flag is that it might take a long time to determine the version since things will have to be downloaded/cloned. @itsayellow you have done some work in this space. Is there a way to determine the latest version of a package quickly for something like this?

itsayellow:
The easy way to do this is to parse `pip search <package_name>`.  We'd then have to parse the output for the `INSTALLED` version and `(latest)` version.  This would take as long as a `pip search` takes, which from an anecdotal test just now on my machine took 0.6 seconds.

Another way would be to do a query of pypi directly using their simple API.  I don't know if this is any better than using pip, and it would be a bunch more work.

I couldn't figure out a way with pip (or otherwise) to just directly check a package's current version other than the above, but I may be missing something.

anthem-bmk:
[...]

PR #1138, opened 2023-12-06
Modification to run.py to check for and upgrade to latest app version when using pipx run

akau01:
- [x ] I have added an entry to `docs/changelog.md`

## Summary of changes
This pull request addresses issue #464 to check for and upgrade to the latest app version when using pipx run. pipx would check for new versions no more than once within a given time period. Right now, the time period is 24 hours. Checking for new versions involves a query of PyPI directly using their simple API to check for a specific package’s latest version. If the latest version is more recent than the currently installed version, the code automatically upgrades the package on your system. Right now, the feature can only handle apps accessible through the simple API.

## Test plan

A basic test was performed by installing an older version of cowsay and running the app. After running pipx run, the program upgraded the app to the latest version (6.1) before running cowsay.

Tested by running

```
pipx install cowsay==5.0
python3 src/pipx run cowsay -t mooo
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## dev
 
+- Check for latest version and upgrade to latest version if it exists when running pipx run
+
 ## 1.3.3
 
 - [docs] Make the logo more visible in dark mode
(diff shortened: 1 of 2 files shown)


chrysle:
I'm not sure this should be accomplished by parsing the HTML content? There was another suggestion in https://github.com/pypa/pipx/issues/464#issuecomment-675206384.

gaborbernat:

PR #1145, opened 2023-12-13
Global install

vit-zikmund:

## Summary of changes
While `pipx` aims not to clutter the global/user python environments, the current instructions do exactly that. How about we stay clean on that front, too?

These instructions use a widely available `python3-venv` package to create a virtual environment for `pipx`, so even it stays out of the still pretty vulnerable `--user` python environment.

## Test plan

Tested by running those suggested commands :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## dev
 
 - [docs] Add Scoop installation instructions
+- [docs] Change Ubuntu <=22.04 installation instructions to also use a global installation (with own virtual environment)
 
 ## 1.3.3
 
--- a/README.md
+++ b/README.md
@@ -51,12 +51,15 @@ pipx ensurepath
 
 - Ubuntu 22.04 or below
 
-```
-python3 -m pip install --user pipx
-python3 -m pipx ensurepath
+```sh
+sudo apt install python3-venv
+sudo python3 -m venv --upgrade-deps /opt/pipx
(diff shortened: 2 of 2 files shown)


chrysle:
This seems unnecessarily complex. I think we should remove these instructions altogether. We'll be able to suggest installing the `pipx` package via Apt with Ubuntu 24.04.

vit-zikmund:
Well, that's already a few lines above, but most people won't be at Ubuntu 24.04 for at least a year or so. This is just an offer ;)

maxwell-k:
I believe that pipx 1.0.0 is packaged for Ubuntu 22.04 in "universe", see https://packages.ubuntu.com/jammy/pipx

So on Ubuntu 22.04 I  follow the instructions above in the section labelled "Ubuntu 23.04 or above"

I haven't worked with Ubuntu 20.04 recently but I think there is a version of pipx available: https://packages.ubuntu.com/focal/pipx

vit-zikmund:
Ah, interesting, there's really enough default coverage, if there's at least some `pipx` at the last supported LTS Ubuntu 20.04:
```sh
docker run --rm -it ubuntu:20.04 bash
apt update && apt install pipx
pipx --version
# 0.12.3.1
```
As the version is a tad old, I suppose the users should still be informed, that likely the latest and freshest release comes from `pip`, but then it's open for any power user to do whatever they please and we don't need to shove this admittedly complicated setup to the general public.
Thanks for the info. Closing the issue now.

PR #1146, opened 2023-12-13
Be more user-friendly by pointing to PyPA specifications instead of PEPs

jeanas:
## Summary of changes

Link to the inline script metadata specification on packaging.python.org instead of PEP 723.


chrysle (approved):
Looks good!

Issue #1150, opened 2023-12-15
--python python3.11 results in an error


maxwell-k:
**Describe the bug**

- The output of `--help` explains that `pipx run` supports  `--python python3.11`.
- My testing shows that this results in an error
- I tested on 1.3.3 and the latest version from main
- A workaround is to use `--python 3.11`
- *Edit* I tested with the `python-launcher` package installed on Fedora 38

**How to reproduce**

Command to reproduce this bug:

    pipx run --python python3.11 pycowsay moo

Error message:

    Traceback (most recent call last):
      File "/home/maxwell-k/.local/bin/pipx", line 8, in <module>
        sys.exit(cli())
                 ^^^^^
      File "/home/maxwell-k/.local/pipx/venvs/pipx/lib64/python3.11/site-packages/pipx/main.py", line 885, in cli
        return run_pipx_command(parsed_pipx_args)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/maxwell-k/.local/pipx/venvs/pipx/lib64/python3.11/site-packages/pipx/main.py", line 192, in run_pipx_command
        py_launcher_python = find_py_launcher_python(args.python)
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/maxwell-k/.local/pipx/venvs/pipx/lib64/python3.11/site-packages/pipx/interpreter.py", line 33, in find_py_launcher_python
        py = subprocess.run(
             ^^^^^^^^^^^^^^^
      File "/usr/lib64/python3.11/subprocess.py", line 571, in run
        raise CalledProcessError(retcode, process.args,
    subprocess.CalledProcessError: Command '['/usr/bin/py', '-python3.11', '-c', 'import sys; print(sys.executable)']' returned non-zero exit status 2.

Command to show `pipx` version:

    pipx --version

Output:

    1.3.3

Also tested with:

    1.3.4.dev8+gb66335c

**Expected behavior**

Command to show help:

    pipx run --help

Excerpt from help:

```
  --python PYTHON       Python to run with. Possible values can be the executable
                        name (python3.11), the version to pass to py launcher (3.11),
                        or the full path to the executable. Requires Python 3.8 or
                        above.
```

I therefore expect the two commands below to be equivalent:

    pipx run --python python3.11 pycowsay moo
    pipx run --python 3.11 pycowsay moo

They should both display a mooing cow:

      ---
    < moo >
      ---
[...]

PR #1151, opened 2023-12-15
Remove executable name option from help

maxwell-k:
This is only one approach to addressing #1150 and may not be preferred.

## Summary of changes

This is a change to the help text for install and run to remove the description of passing an executable name. After this change the help does not suggest invocations including `--python python3.11` which lead to #1150 . Fixes https://github.com/pypa/pipx/issues/1150

## Test plan

Tested by running:

```
.nox/develop-3-12/bin/python src/pipx install --help
.nox/develop-3-12/bin/python src/pipx run --help
```

Example excerpt from output:

```
  --python PYTHON       Python to run with. Possible values can be the version to pass to py launcher
                        (3.11) or the full path to the executable. Requires Python 3.8 or above.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## dev
 
+- Fix `--help` by removing the option to pass an executable name to `--python`
 - [docs] Add Scoop installation instructions
 
 ## 1.3.3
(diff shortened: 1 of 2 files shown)


maxwell-k:
After some testing without python-launcher installed I am not sure this is the right fix.

Issue #1152, opened 2023-12-15
pipx 1.3 breaks `run <app>` if `app` is also a directory in your cwd


robinvandernoord:
In pipx 1.2.x, you could do `pipx run py4web version` if there was a directory called py4web in your cwd. Since 1.3 up to the latest patch, this leads to an error:

```
Traceback (most recent call last):
  File "/home/ubuntu/.local/bin/pipx", line 8, in <module>
    sys.exit(cli())
  File "/home/ubuntu/.local/lib/python3.10/site-packages/pipx/main.py", line 885, in cli
    return run_pipx_command(parsed_pipx_args)
  File "/home/ubuntu/.local/lib/python3.10/site-packages/pipx/main.py", line 197, in run_pipx_command
    commands.run(
  File "/home/ubuntu/.local/lib/python3.10/site-packages/pipx/commands/run.py", line 195, in run
    content = None if spec is not None else maybe_script_content(app, is_path)
  File "/home/ubuntu/.local/lib/python3.10/site-packages/pipx/commands/run.py", line 52, in maybe_script_content
    return app_path.read_text(encoding="utf-8")
  File "/usr/lib/python3.10/pathlib.py", line 1134, in read_text
    with self.open(mode='r', encoding=encoding, errors=errors) as f:
  File "/usr/lib/python3.10/pathlib.py", line 1119, in open
    return self._accessor.open(self, mode, buffering, encoding, errors,
IsADirectoryError: [Errno 21] Is a directory: 'py4web'
```

**How to reproduce**

Create a directory with the same name as a Python package you would use with pipx. Then try to run `pipx run <name>`.

**Expected behavior**

I expected the pipx installed app to run like in 1.2.x

dukecat0:
Probably we should only treat the app name as local script when it contains:
- `.py` extension
- A shebang line at the beginning of script

...Or providing a new option called `--ignore-local`/`-i`? (Since we have `--path` which forces pipx to find the script locally.)

robinvandernoord:
I would prefer the first solution since it's more backwards compatible, if there was an option for it I have to change a lot of docker build scripts and other tooling that use pipx.

PR #1155, opened 2023-12-18
Verify if script name provided is a file before running it

dukecat0:

## Summary of changes
`pipx run`: Verify whether the script name provided is a file before running it
Closes https://github.com/pypa/pipx/issues/1152

## Test plan

Tested by running

```
nox -s tests-3.12
```


chrysle:
Didn't you want to detect whether the file is a valid script?

src/pipx/commands/run.py
```diff
@@ -48,7 +48,7 @@ def maybe_script_content(app: str, is_path: bool) -> Optional[str]:
 
     # Look for a local file first.
     app_path = Path(app)
-    if app_path.exists():
+    if app_path.exists() and app_path.is_file():
```
uranusjr, line 51:
```suggestion
    if app_path.is_file():
```

I believe `is_file` implies `exists`.

https://docs.python.org/3/library/pathlib.html#pathlib.Path.is_file

> False is also returned if the path doesn’t exist […].

dukecat0:
> Didn't you want to detect whether the file is a valid script?

I think a file that has no `.py` extension or shebang can be a valid script as well. (At least the Python interpreter doesn't care about it.) That's why I personally believe that checking whether it is a file is enough.

gaborbernat (changes requested):
Test please 👍

dukecat0:
Not sure if we need a test for this...? I personally think this is a very minor bug and doesn't require a test.

gaborbernat:
> Not sure if we need a test for this...? I personally think this is a very minor bug and doesn't require a test.

I disagree.

gaborbernat (changes requested):
Tests please.

dukecat0:
> Tests please.

Added

gaborbernat (approved):

PR #1163, opened 2023-12-22
Allow skipping maintenance tasks for list command

### LINKED ISSUE #1081: Option for the `list` command not to go out to the internet

When using pipx while the system has no internet connection, `pipx list` takes an awful long time because before printing teh list of packages it tries to upgrade the shared libraries.

The main purpose of the `list` command is to list what's there, so if I know I have no internet I should be able to say "_Just give me the list. I don't care about doing other maintenance things._" because these maintenance things won't work in any case.

An option like `--no-maintenance` or `--dont-update` for the `list` command.

Gitznik:
- [X] I have added an entry to `docs/changelog.md`

## Summary of changes
Add a `--skip-maintenance` flag to the list command. This sets a new state on the `_SharedLibs` class, that is used to skip the shared libs upgrade step by setting `needs_upgrade` to False. This change allows the user to skip the functionality that reaches out to `pip`, which makes executing the command a lot faster when offline.
resolves #1081 

## Test plan

Introduced a new test case `test_skip_maintenance`. It first adds some packages to introduce shared libs. It then verifies that a normal `list` command sets the expected state on the `shared_libs` object. Then it verifies that running `list` with `--skip-maintenance` results in state on the `shared_libs` object that indicates that no upgrade was performed.

Tested by running `nox -s tests-3.12`

```
pipx list --skip-maintenance
```


src/pipx/commands/list_packages.py
```diff
@@ -89,12 +89,17 @@ def list_packages(
...
     short_format: bool,
+    skip_maintenance: bool,
 ) -> ExitCode:
     """Returns pipx exit code."""
     venv_dirs: Collection[Path] = sorted(venv_container.iter_venv_dirs())
     if not venv_dirs:
         print(f"nothing has been installed with pipx {sleep}", file=sys.stderr)
 
+    if skip_maintenance:
+        shared_libs.shared_libs.skip_maintenance = True
```
chrysle, line 100: Maybe rename this to `skip_upgrade`... I don't think it's a problem to be more concise internally.

chrysle (commented):
Concerning the test, probably really collect the times after running the CLI, and compare the durations.

Gitznik:
> Concerning the test, probably really collect the times after running the CLI, and compare the durations.

I did this manually:
`python3 src/pipx list --skip-maintenance  0.42s user 0.14s system 87% cpu 0.638 total`
vs with maintenance
`python3 src/pipx list  0.72s user 0.21s system 10% cpu 8.623 total`
when running offline and being in a state that would require an upgrade.

Do you have a suggestion on how to automate this with a test? My only suggestion would be to modify `pipx_temp_env` to allow creating an unresponsive local pypi server, but that feels like a lot of custom logic for one test.

uranusjr (approved):

gaborbernat (approved):

gaborbernat (changes requested):
Needs merge conflicts fixed and rebased.

Gitznik:
> Needs merge conflicts fixed and rebased.

Done

dukecat0 (dismissed):

PR #1176, opened 2023-12-29
Don't show debug messages when verbose isn't passed

### LINKED ISSUE #1175: Getting debug-like log messages in console after upgrade to v1.4.0
**Describe the bug**

After upgrading to v1.4.0, I suddenly get the following console messages when running `pipx` commands:

```sh
$ pipx list
pipx version is 1.4.0
Default python interpreter is '/opt/homebrew/opt/python@3.12/libexec/bin/python'
running <checking pip's availability>
Time since last upgrade of shared libs, in seconds: 2674. Upgrade will be run by pipx if greater than 2592000.
running <checking pip's availability>
running <checking pip's availability>
running <checking pip's availability>
running <checking pip's availability>
running <checking pip's availability>
running <checking pip's availability>
running <checking pip's availability>
running <checking pip's availability>
...
```

The excessive log messages will go away when adding one level of `-q` flag to `pipx`.

On a first glance, the message `running <checking pip's availability>` is logged with `logger.info()`. Not sure, if that's really meant to be a info and not debug message though. But that log message call has not been changed for a long time. Maybe there’s a side effect with one of the recent log level changes that went into v1.4.

**How to reproduce**

Run `pipx list` without `-q` or `-v` flags.

**Expected behavior**

Show console log messages as it was pre v1.4.

**Environment**

```sh
$ pipx environment
pipx version is 1.4.0
Default python interpreter is '/opt/homebrew/opt/python@3.12/libexec/bin/python'

PIPX_HOME=
PIPX_BIN_DIR=
PIPX_MAN_DIR=
PIPX_SHARED_LIBS=
PIPX_DEFAULT_PYTHON=
USE_EMOJI=

PIPX_HOME=/Users/tspng/.local/pipx
PIPX_BIN_DIR=/Users/tspng/.local/bin
PIPX_MAN_DIR=/Users/tspng/.local/share/man
PIPX_SHARED_LIBS=/Users/tspng/.local/pipx/shared
PIPX_LOCAL_VENVS=/Users/tspng/.local/pipx/venvs
PIPX_LOG_DIR=/Users/tspng/.local/pipx/logs
PIPX_TRASH_DIR=/Users/tspng/.local/pipx/.trash
PIPX_VENV_CACHEDIR=/Users/tspng/.local/pipx/.cache
USE_EMOJI=true
```

dukecat0:

## Summary of changes
Don't show debug messages when verbose isn't passed. This is implemented by changing all `logging.info(...)` to `logging.debug(...)`.
Closes #1175.

## Test plan

Tested by running

```
```

### DIFF
--- a/src/pipx/commands/common.py
+++ b/src/pipx/commands/common.py
@@ -129,7 +129,7 @@ def _symlink_package_resource(
         mkdir(symlink_path.parent)
 
     if force:
-        logger.info(f"Force is true. Removing {str(symlink_path)}.")
+        logger.debug(f"Force is true. Removing {str(symlink_path)}.")
         try:
             symlink_path.unlink()
         except FileNotFoundError:
@@ -141,7 +141,7 @@ def _symlink_package_resource(
     is_symlink = symlink_path.is_symlink()
     if exists:
         if symlink_path.samefile(path):
-            logger.info(f"Same path {str(symlink_path)} and {str(path)}")
+            logger.debug(f"Same path {str(symlink_path)} and {str(path)}")
         else:
             logger.warning(
                 pipx_wrap(
@@ -155,7 +155,7 @@ def _symlink_package_resource(
             )
         return
     if is_symlink and not exists:
-        logger.info(f"Removing existing symlink {str(symlink_path)} since it " "pointed non-existent location")
+        logger.debug(f"Removing existing symlink {str(symlink_path)} since it pointed to non-existent location")
         symlink_path.unlink()
 
     if executable:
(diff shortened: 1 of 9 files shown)


jaklan:
@dukecat0 what's wrong with that solution? These logs seem to be very typical `debug` messages, which are usually displayed when providing `--verbose / -v` flag.

dukecat0:
@jaklan I think the underlying issue isn't here, but it's related to the logic here: https://github.com/pypa/pipx/blob/7274ee47ba3ca0315e3ae92e5e1d26c66241a394/src/pipx/main.py#L794-L796

jaklan:
@dukecat0 it depends on the approach. The current logic makes `INFO` the default level, which is not bad per se - it's bad because at the moment we make typical `DEBUG` logs the `INFO` ones, so I believe that PR has a lot of sense (conceptually, as I haven't analysed the individual changes).

Generally I like the approach used in e.g. [`dbt`](https://docs.getdbt.com/reference/global-configs/logs), imho it's very reasonable:
- `INFO` level by default
- `--debug` / `--verbose` flag to change the level to `DEBUG`
- `--quiet` flag to change the level to `ERROR`
- `--log-level` flag to select any level you wish

dukecat0:
The current solution is here: https://github.com/pypa/pipx/issues/1175#issuecomment-1872320056
I just confirmed locally and it should make everything go back to its original state.

jaklan:
@dukecat0 if we want to treat `WARNING` logs as the user-facing ones, not the `INFO` ones, then sure - would work as well. That's what I meant by `depends on the approach` 😄

PR #1178, opened 2023-12-30
Set default logging level to WARNING

### LINKED ISSUE #1175: Getting debug-like log messages in console after upgrade to v1.4.0

### LINKED ISSUE #1177: Running `pipx` without any arguments errors on `1.4.0`
**Describe the bug**

Running `pipx` without any args in `1.3.3` prints the help:

```
❯ pipx
usage: pipx [-h] [--version] {install,uninject,inject,upgrade,upgrade-all,uninstall,uninstall-all,reinstall,reinstall-all,list,run,runpip,ensurepath,environment,completions} ...

Install and execute apps from Python packages.

Binaries can either be installed globally into isolated Virtual Environments
or run directly in a temporary Virtual Environment.
```

in `1.4.0` it errors with:
```
❯ pipx
  File "/opt/homebrew/bin/pipx", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "/opt/homebrew/Cellar/pipx/1.4.0/libexec/lib/python3.12/site-packages/pipx/main.py", line 905, in cli
    setup(parsed_pipx_args)
  File "/opt/homebrew/Cellar/pipx/1.4.0/libexec/lib/python3.12/site-packages/pipx/main.py", line 843, in setup
    verbose = args.verbose - args.quiet
              ^^^^^^^^^^^^
AttributeError: 'Namespace' object has no attribute 'verbose'
```
**How to reproduce**

Run `pipx` against latest version with no args. Installed with homebrew. Python 3.12.

**Expected behavior**

Should print the same as `pipx --help`, which works on `1.4.0`

dukecat0:

## Summary of changes
Set the default logging level to WARNING, so debug log messages won't be shown without passing additional flags such as `--verbose`.

Closes #1175
Closes #1177 

## Test plan

Tested by running

```
nox -s tests-3.12
```

### DIFF
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -840,7 +840,7 @@ def setup(args: argparse.Namespace) -> None:
         print_version()
         sys.exit(0)
 
-    verbose = getattr(args, 'verbose', 0) - getattr(args, 'quiet', 0)
+    verbose = getattr(args, "verbose", 0) - getattr(args, "quiet", 0)
 
     setup_logging(verbose)
 
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -699,15 +699,6 @@ def get_command_parser() -> argparse.ArgumentParser:
 
     completer_venvs = InstalledVenvsCompleter(venv_container)
 
-    parser = argparse.ArgumentParser(
-        prog=prog_name(),
-        formatter_class=LineWrapRawTextHelpFormatter,
-        description=PIPX_DESCRIPTION,
-    )
-    parser.man_short_description = PIPX_DESCRIPTION.splitlines()[1]  # type: ignore
-
-    subparsers = parser.add_subparsers(dest="command", description="Get help for commands with pipx COMMAND --help")
-
     shared_parser = argparse.ArgumentParser(add_help=False)
 
     shared_parser.add_argument(
@@ -722,6 +713,16 @@ def get_command_parser() -> argparse.ArgumentParser:
     )
 
     shared_parser.add_argument("--verbose", "-v", action="count", default=0, help=("Give more output."))
+
+    parser = argparse.ArgumentParser(
+        prog=prog_name(),
+        formatter_class=LineWrapRawTextHelpFormatter,
+        description=PIPX_DESCRIPTION,
+        parents=[shared_parser],
+    )
+    parser.man_short_description = PIPX_DESCRIPTION.splitlines()[1]  # type: ignore
+
+    subparsers = parser.add_subparsers(dest="command", description="Get help for commands with pipx COMMAND --help")
 
     _add_install(subparsers, shared_parser)
     _add_uninject(subparsers, completer_venvs.use, shared_parser)
@@ -840,7 +841,7 @@ def setup(args: argparse.Namespace) -> None:
         print_version()
         sys.exit(0)
 
-    verbose = getattr(args, "verbose", 0) - getattr(args, "quiet", 0)
+    verbose = args.verbose - args.quiet
 
     setup_logging(verbose)
 
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -786,12 +786,12 @@ def setup_log_file() -> Path:
         return _setup_log_file(platformdirs.user_log_path("pipx"))
 
 
-def setup_logging(verbose: bool) -> None:
+def setup_logging(verbose: int) -> None:
     pipx_str = bold(green("pipx >")) if sys.stdout.isatty() else "pipx >"
     pipx.constants.pipx_log_file = setup_log_file()
 
     # Determine logging level
-    level_number = max(0, 2 - verbose) * 10
+    level_number = max(0, logging.WARNING - 10 * verbose)
 
     level = logging.getLevelName(level_number)
 
@@ -840,7 +840,7 @@ def setup(args: argparse.Namespace) -> None:
         print_version()
         sys.exit(0)
 
-    verbose = args.verbose - args.quiet
+    verbose = getattr(args, 'verbose', 0) - getattr(args, 'quiet', 0)
 
     setup_logging(verbose)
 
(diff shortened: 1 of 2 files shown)


src/pipx/main.py
```diff
@@ -840,7 +840,7 @@ def setup(args: argparse.Namespace) -> None:
         print_version()
         sys.exit(0)
 
-    verbose = args.verbose - args.quiet
+    verbose = getattr(args, "verbose", 0) - getattr(args, "quiet", 0)
```
chrysle, line 843: Or we could set argparses' [`default`](https://docs.python.org/3/library/argparse.html#default) keyword argument?

dukecat0: It's set to 0 already: https://github.com/pypa/pipx/blob/98635d27d93aff9f74bfca660ae81115d3d37d86/src/pipx/main.py#L724

chrysle: Then I don't understand why this change is needed.

dukecat0: See #1177

chrysle: Yes, but `argparse` is supposed to fill in that argument with the default value, isn't it?

I think I've found the culprit: We use a shared parser for defining the verbose and quiet arguments, but don't apply that as a parent parser for the base parser. Thus, the following patch should work:

```patch
diff --git a/src/pipx/main.py b/src/pipx/main.py
index 11e01e5..7dc5d75 100644
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -720,12 +720,6 @@ def get_command_parser() -> argparse.ArgumentParser:
 
     completer_venvs = InstalledVenvsCompleter(venv_container)
 
-    parser = argparse.ArgumentParser(
-        prog=prog_name(),
-        formatter_class=LineWrapRawTextHelpFormatter,
-        description=PIPX_DESCRIPTION,
-    )
-    parser.man_short_description = PIPX_DESCRIPTION.splitlines()[1]  # type: ignore
 
     shared_parser = argparse.ArgumentParser(add_help=False)
 
@@ -744,6 +738,14 @@ def get_command_parser() -> argparse.ArgumentParser:
         "--verbose", "-v", action="count", default=0, help=("Give more output.")
     )
 
+    parser = argparse.ArgumentParser(
+        prog=prog_name(),
+        formatter_class=LineWrapRawTextHelpFormatter,
+        description=PIPX_DESCRIPTION,
+        parents=[shared_parser],
+    )
+    parser.man_short_description = PIPX_DESCRIPTION.splitlines()[1]  # type: ignore
+
     subparsers = parser.add_subparsers(
         dest="command", description="Get help for commands with pipx COMMAND --help"
     )
```

dukecat0: Done

chrysle (approved):

dukecat0:
Probably we need to update this line to `ERROR and CRITICAL logging levels.`?
https://github.com/pypa/pipx/blob/81c2fa509768715de24a3ba10d46596de980052f/src/pipx/main.py#L720

chrysle:
Yes, looks like it.

mayeut:
The `"-q"` added in #1159 shall also be removed:
https://github.com/pypa/pipx/blob/81c2fa509768715de24a3ba10d46596de980052f/tests/test_list.py#L83

dukecat0:
We can change its behavior instead of removing it. I remember the original usage of this flag was hiding all console outputs, not just the debug log messages.

mayeut:
> We can change its behavior instead of removing it. I remember the original usage of this flag was hiding all console outputs, not just the debug log messages.

It might be so but:
IMHO, it should have raised a red flag in review (but can be easily missed).
Removing the flag in that test allows to see that the default log level is the correct one (for now, it seems it's the only test for that - although probably accidental).
Given this PR reverts the default log level to WARNING, I think we should also revert this change that was only introduced because of the default log level change.

If there are more changes to be made as you suggest, then it would be for another PR but for now, I think changing back this test is the best option.

dukecat0:
I've opened #1181 to revert it now.

PR #1180, opened 2023-12-31
fix: use modern PEP 723 syntax, fix mistake in name

henryiii:

## Summary of changes

PEP 723 has been updated to use a `scripts` block and no `run` section. This updates to the new syntax, ~~but still runs with the old syntax, producing a warning.~~ This can be upgraded to an error (but with a nice message) later (or now, if you prefer, as it's still undocumented).

Edit: the old form used the wrong name (requirements instead of dependencies) so I went ahead and made it an error now. I can still do the soft deprecation instead though if that’s preferred.

## Test plan

Tested by running

```
nox -s tests-3.11
```

There were 18 failures before this PR, and 18 after, so hoped that was good! :)


src/pipx/commands/run.py
```diff
@@ -348,7 +361,8 @@ def _get_requirements_from_script(content: str) -> Optional[List[str]]:
     pyproject = tomllib.loads(content)
 
     requirements = []
-    for requirement in pyproject.get("run", {}).get("requirements", []):
+    req_table = pyproject if script else pyproject.get("run", {})
+    for requirement in req_table.get("requirements", []):
```
branchvincent, line 365: isn't the table called `dependencies`?
```suggestion
    for requirement in req_table.get("dependencies", []):
```

henryiii: Oh, wow. That was an issue before too. Okay, I think since the implementation was wrong we can go ahead and make it a hard (but nicely descriptive) error if the old form is encountered?

dukecat0 (approved):

dukecat0:
Please add a changelog entry.

flying-sheep:
@henryiii would it be possible to make the tracebacks friendlier? Currently they show

```pytb
  File "<string>", line 11, in <module>
ModuleNotFoundError: No module named 'requests'
```

Would be nice if that read “File "actual-file-name.py", line 11, in __main__`

henryiii:
> would it be possible to make the tracebacks friendlier?

I can look into that.

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 ## dev
 
 - Allow skipping maintenance tasks during list command
-
```
dukecat0, line 4: It should be located in `dev` section.

dukecat0 (approved):

CHANGELOG.md
dukecat0: Oh wrong button, should be request change 😅

dukecat0: It seems like the changelog is very messy due to merge conflicts. I will open a new PR to tidy it up.

PR #1183, opened 2024-01-02
Remove "Running pipx From Source Tree" section

chrysle:

## Summary of changes

Running `pipx` from source isn't possible any more since #1130 (removal of the `pipx.version` module). Therefore I removed
the corresponding section in `CONTRIBUTING.md`.

## Test plan

Tested by running

```
```


dukecat0:
Seems like it can still be run correctly:
```
$ python src/pipx --version
1.4.1.dev5+g98dad34
```

chrysle:
Good idea, done. I also removed the `enabled` argument from the search plugin configuration in `mkdocs.yml`. It somehow broke the MkDocs build, and is set to `true` by default anyway.

dukecat0 (approved):

PR #1184, opened 2024-01-02
Bump changelog manually

dukecat0:
Looks like the workflow "Bump changelog on release" can't be run successfully due to branch protection.


chrysle (approved):
Thanks, I think I'll have a fix for the GH Action soon.

PR #1190, opened 2024-01-05
Update run.py - do not append .exe in _download_and_run()

Spitfire1900:

## Test plan

Tested by running

```
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## dev
 
 - Allow skipping maintenance tasks during list command
- 
+
 ## 1.4.1
 
 - Set default logging level to WARNING, so debug log messages won't be shown without passing additional flags such as `--verbose`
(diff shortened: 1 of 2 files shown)


chrysle:
Could you probably add logic to detect whether the application has a _.exe_ file ending or not?

henryiii:
Assuming `.exe` is in `%PATHEXT%` (which it should always be!), then you don't need the extension to call it or to find it with [shutil.which](https://docs.python.org/3/library/shutil.html#shutil.which). If it's used other places, milage will vary, but I know those two are fine.

Spitfire1900:
Not having '.exe' in %PATHEXT% on Windows would likely break your system, I don't think it's even appropriate to append .exe to the command call, unless we're intentionally trying to avoid overlap with a .bat somewhere.

Spitfire1900:
FYI I'm still confused why I'm getting this, if this is reproducible on others' systems it would be a BLOCKER bug.

I will attempt to reproduce this on another Windows system before removing the draft status on this PR.

gaborbernat:

Issue #1187, opened 2024-01-04
PEP 723 support is faulty: uses wrong TOML field name


flying-sheep:
**Describe the bug**
#1100 implemented [PEP 723](https://peps.python.org/pep-0723/) wrong:

https://github.com/pypa/pipx/blob/5b18ddf5b2871749a1d9e90938da407c4ba3f7a4/src/pipx/commands/run.py#L351

this should be `"dependencies"`, not `"requirements"`.

**How to reproduce**

Create a PEP 723 compliant script:

```python
#!/usr/bin/env -S pipx run
# /// pyproject
# run.dependencies = ["httpx"]
# ///
import httpx
```

and run it using `pipx run ./script.py`. You’ll see

```pytb
Traceback (most recent call last):
  File "<string>", line 11, in <module>
ModuleNotFoundError: No module named 'requests'
```

**Expected behavior**

`pipx` should run it in a temporary environment.

Also the traceback should contain the file name, not `<string>`

dukecat0:
Fixed in #1180.

flying-sheep:
Ah, indeed. I searched the issues, but apparently not the PRs

dukecat0:
> Also the traceback should contain the file name, not <string>

However, I think this issue still exists.

maxwell-k:
The example in the issue report is unfortunately also out of date the most recent change to PEP 723 was to change:

```diff
-# /// pyproject
+# /// script
```

I don't want to write an example and risk getting it wrong myself, instead I will point to the example in the PEP
https://peps.python.org/pep-0723/#how-to-teach-this

For the background on this change please see https://discuss.python.org/t/how-would-you-like-to-declare-runtime-dependencies-and-python-requirements-for-pep-723/40418/70

henryiii:
@maxwell-k this is also fixed in #1180.

PR #1191, opened 2024-01-05
fix: nicer tracebacks in run mode on scripts

henryiii:
This is one way to improve the tracebacks as reported in #1187 and mentioned in #1180.

## Summary of changes

This is the simplest way I initially see to get filenames. Instead of passing around the contents of the script as a string, scripts are passed around as Paths, which allows the run command to run the path rather than passing the string on the command line.

## Test plan

If you start with this script:

```python
import httpx
```

You get this:

```console
$ pipx run ./script.py
  File "<string>", line 1, in <module>
    import httpx
ModuleNotFoundError: No module named 'httpx'
```

After this PR, you get this:

```console
$ pipx run ./script.py
  File "/Users/henryschreiner/git/software/pipx/script.py", line 1, in <module>
    import httpx
ModuleNotFoundError: No module named 'httpx'
```

### DIFF
--- a/src/pipx/commands/run.py
+++ b/src/pipx/commands/run.py
@@ -8,7 +8,7 @@ import urllib.parse
 import urllib.request
 from pathlib import Path
 from shutil import which
-from typing import List, NoReturn, Optional
+from typing import List, NoReturn, Optional, Union
 
 from packaging.requirements import InvalidRequirement, Requirement
 
@@ -42,14 +42,14 @@ Available executable scripts:
     {app_lines}"""
 
 
-def maybe_script_content(app: str, is_path: bool) -> Optional[str]:
+def maybe_script_content(app: str, is_path: bool) -> Optional[Union[str, Path]]:
     # If the app is a script, return its content.
     # Return None if it should be treated as a package name.
 
     # Look for a local file first.
     app_path = Path(app)
     if app_path.is_file():
-        return app_path.read_text(encoding="utf-8")
+        return app_path
     elif is_path:
         raise PipxError(f"The specified path {app} does not exist")
 
@@ -71,7 +71,7 @@ def maybe_script_content(app: str, is_path: bool) -> Optional[str]:
 
 
 def run_script(
-    content: str,
+    content: Union[str, Path],
     app_args: List[str],
     python: str,
     pip_args: List[str],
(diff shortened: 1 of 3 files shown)


src/pipx/commands/run.py
```diff
@@ -99,7 +102,10 @@ def run_script(
             venv = Venv(venv_dir, python=python, verbose=verbose)
             venv.create_venv(venv_args, pip_args)
             venv.install_unmanaged_packages(requirements, pip_args)
-        exec_app([venv.python_path, "-c", content, *app_args])
+        if isinstance(content, Path):
+            exec_app([venv.python_path, content, *app_args])
+        else:
+            exec_app([venv.python_path, "-c", content, *app_args])
```
flying-sheep, line 108:
```suggestion
        python_path = venv.python_path

    if isinstance(content, Path):
        exec_app([python_path, content, *app_args])
    else:
        exec_app([python_path, "-c", content, *app_args])
```

flying-sheep (changes requested):
Looks lovely! I think there could be some slight deduplication, otherwise wonderful!

src/pipx/commands/run.py
```diff
@@ -81,7 +81,7 @@ def run_script(
 ) -> NoReturn:
     requirements = _get_requirements_from_script(content)
     if requirements is None:
-        exec_app([python, "-c", content, *app_args])
+        python_path = python
```
henryiii, line 84:
```suggestion
        python_path = Path(python)
```

chrysle (dismissed):
Looks good!

chrysle:
I think we should really enable merge queues, this constant branch updating is annoying.

rsalmaso:
Happy for this patch!

Running script instead of string solve another problem: run the script via shebang `#!/usr/bin/env -S pipx run` and get the correct filename `sys.argv[0]`.

dukecat0 (changes requested):
Test please, otherwise LGTM.

henryiii:
What should this do if a URI is passed, by the way? Currently that still says "<string>". That could be a future improvement, but something to think about.

dukecat0:
> What should this do if a URI is passed, by the way? Currently that still says "<string>". That could be a future improvement, > but something to think about.

Perhaps we can have an additional check here?
https://github.com/pypa/pipx/blob/5563ae68aecda0b36e59979bcbd8e81a96ec3fbd/src/pipx/commands/run.py#L57

henryiii:
A check wouldn't help, we are already doing the "right" think (since when getting a URI it needs to be downloaded, possibly from a remote source), it's just then a question of if there's a way to get it to show something nicer than "<string>". This was fixable since loading a local file can just be directly run by the `python` command. I would assume there's a way to customize running on arbitrary input, since things like IPython do it (though at in least some cases, they write out temporary files).

gaborbernat (changes requested):
Let's rebase and add changelog.

dukecat0 (approved):

PR #1168, opened 2023-12-24
Pass real paths when running subprocesses

dechamps:
## Summary of changes

Fixes #1164 by making sure real paths are passed to executables that are not subject to path redirection.

## Test plan

```
winget install python --source=msstore
pip install .
python -m pipx ensurepath
pipx install pycowsay
pycowsay
```

The above fails without this PR with the error shown in #1164, but succeeds with this PR.

I [tried](https://github.com/dechamps/pipx/commit/8cb2b83884d736c28923bb261e550dbb425ac77e) to add a GitHub Actions job that does the above so that any future regressions will be caught, but unfortunately installing Microsoft Store apps in a GitHub Actions workflow does not appear to be easy (see e.g. Cyberboss/install-winget#1).

PR #1196, opened 2024-01-08
Fix python path resolution on linux

### LINKED ISSUE #1195: Failing to create venv when passing --python version on linux
**Describe the bug**

It looks like since https://github.com/pypa/pipx/pull/1168 pipx is no longer able to create a virtual environment when passing `--python` (at least on linux and with pyenv managed versions). It still works fine if a venv already exists.

From my tests I assume it's caused by how `os.path.realpath` creates the file path to the python executable provided. I've only tested this with python versions being managed by pyenv.

```
which python3.9
-> /home/robert/.pyenv/shims/python3.9 # This is fine

python3.9 -c "import sys; print(sys.executable)"
-> /home/robert/.pyenv/versions/3.9.18/bin/python3.9 # This is fine

python3.9 -c "import os;print(os.path.realpath('python3.9'));"
-> /home/robert/private/code/pipx/python3.9 # This is wrong!
```
**How to reproduce**
`pipx run --python=python3.11 pycowsay moo`

```
(develop-3-12) ➜  pipx git:(main) ✗ pipx run --python=python3.9 --verbose pycowsay moo
pipx >(setup:853): pipx version is 0.1.dev661+g7d82745.d20240106
pipx >(setup:854): Default python interpreter is '/home/robert/private/code/pipx/.nox/develop-3-12/bin/python'
pipx >(run_package:156): venv location is /home/robert/.local/pipx/.cache/7c9b64e6b834c48
pipx >(package_name_from_spec:370): Determined package name: pycowsay
pipx >(package_name_from_spec:371): Package name determined in 0.0s
pipx >(run_subprocess:168): running python3.9 -m venv --without-pip /home/robert/.local/pipx/.cache/7c9b64e6b834c48
  File "/home/robert/private/code/pipx/.nox/develop-3-12/bin/pipx", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "/home/robert/private/code/pipx/src/pipx/main.py", line 914, in cli
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/robert/private/code/pipx/src/pipx/main.py", line 197, in run_pipx_command
    commands.run(
  File "/home/robert/private/code/pipx/src/pipx/commands/run.py", line 200, in run
    run_package(
  File "/home/robert/private/code/pipx/src/pipx/commands/run.py", line 157, in run_package
    _download_and_run(
  File "/home/robert/private/code/pipx/src/pipx/commands/run.py", line 234, in _download_and_run
    venv.create_venv(venv_args, pip_args, override_shared)
  File "/home/robert/private/code/pipx/src/pipx/venv.py", line 160, in create_venv
    venv_process = run_subprocess(cmd + venv_args + [str(self.root)])
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/robert/private/code/pipx/src/pipx/util.py", line 179, in run_subprocess
    completed_process = subprocess.run(
                        ^^^^^^^^^^^^^^^
  File "/home/robert/.pyenv/versions/3.12.1/lib/python3.12/subprocess.py", line 548, in run
    with Popen(*popenargs, **kwargs) as process:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/robert/.pyenv/versions/3.12.1/lib/python3.12/subprocess.py", line 1026, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/home/robert/.pyenv/versions/3.12.1/lib/python3.12/subprocess.py", line 1950, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: '/home/robert/private/code/pipx/python3.9'
```
**Expected behavior**

Creating a venv and then printing pycowsay.

Gitznik:
- [X] I have added an entry to `docs/changelog.md`

## Summary of changes

Only use the logic introduced in https://github.com/pypa/pipx/pull/1168 to fix a path resolution issue on windows when pipx is running on a Windows machine. 
This fixes #1195 

## Test plan

When no venv for python3.9 exists, but python is on the PATH (in my case via pyenv):
```
pipx run --python=python3.9 pycowsay moo
```
Before this change leads to the error described in #1195 , and works after this change.

Tested by running

```
pipx run --python=python3.9 pycowsay moo
```

### DIFF
--- a/src/pipx/util.py
+++ b/src/pipx/util.py
@@ -174,7 +174,7 @@ def run_subprocess(
     # See https://github.com/pypa/pipx/issues/1164
     # Conversely, if the binary is a symlink, then we should NOT use the real path, as Python expects to receive the
     # symlink in argv[0] so that it can locate the venv.
-    if not os.path.islink(cmd_str_list[0]):
+    if not os.path.islink(cmd_str_list[0]) and WINDOWS:
         cmd_str_list[0] = os.path.realpath(cmd_str_list[0])
     completed_process = subprocess.run(
         cmd_str_list,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Update `pipx run` on scripts using `/// script` and no `run` table following the updated version of PEP 723 (#1180)
 - Avoid repeated exception logging in a few rare cases (#1192)
 - Include `tomli` into `pipx.pyz` (zipapp) so that it can be executed with Python 3.10 or earlier (#1142)
+- Fix resolving the python executable path on linux
 
 ## 1.4.1
 
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -1,6 +1,6 @@
 import sys
 
-import pytest
+import pytest  # type: ignore
 
 from pipx import util
 from pipx.util import run_subprocess
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -1,0 +1,26 @@
+import sys
+
+import pytest
+
+from pipx import util
+from pipx.util import run_subprocess
+
+
+@pytest.mark.skipif(sys.platform.startswith("win"), reason="Path resolution skip if not on windows")
+def test_executable_path_resolution_unix():
+    cmd = ["python99.99", "-c", "import sys;"]
+    try:
+        run_subprocess(cmd)
+    except FileNotFoundError as e:
+        assert "No such file or directory: 'python99.99'" in str(e)
+
+
+@pytest.mark.skipif(sys.platform.startswith("win"), reason="Path resolution skip if not on windows")
+def test_executable_path_resolution_fails_on_unix_if_not_skipped(monkeypatch: pytest.MonkeyPatch):
+    cmd = ["python99.99", "-c", "import sys;"]
+    monkeypatch.setattr(util, "WINDOWS", True)
+    monkeypatch.chdir("./tests")
+    try:
+        run_subprocess(cmd)
+    except FileNotFoundError as e:
+        assert "tests/python99.99'" in str(e)


robinbowes:
Would love to see this merged b& released!

gaborbernat (changes requested):
Test please and merge conflict.

Gitznik:
> Test please and merge conflict.

I added some tests. I could not come up with a good way of reproducing this exact issue in a test suite without requiring multiple python versions to be installed on the runner, which I think would not be great for local execution. Let me know if you have other ideas @gaborbernat. 

BTW we could set `os.path.realpath` to fail if it can't resolve the path (at the moment it does not fail, but instead creates the invalid path on unix based systems), but I was unsure if that would break a lot of things on windows.

src/pipx/util.py
```diff
@@ -174,7 +174,7 @@ def run_subprocess(
     # See https://github.com/pypa/pipx/issues/1164
     # Conversely, if the binary is a symlink, then we should NOT use the real path, as Python expects to receive the
     # symlink in argv[0] so that it can locate the venv.
-    if not os.path.islink(cmd_str_list[0]):
+    if not os.path.islink(cmd_str_list[0]) and WINDOWS:
```
ewjoachim, line 177: The file uses a mix of `pathlib.Path` and `os.path`. pathlib supports a islink equivalent (`Path.is_symlink`) and a `realpath` (`Path.resolve`). I think it might make sense to use those rather than os.path for the 2 lines, but I can't really give a compelling reason, so it's ok to completely ignore this too.

Also, before I saw this PR, I was thinking about dealing with it by checking if `cmd_str_list[0]` is absolute or not, but your solution `and WINDOWS` also makes sense. I wanted to suggest the alternative because I don't know yet if any is better, but other reviewers may have ideas on the matter. I do think the `and WINDOWS` approach is at least very pragmatic and unlikely to break further so 👍 

Finally, I think the line would be more readable as:
```suggestion
    if WINDOWS and not os.path.islink(cmd_str_list[0]):
```
because then when you read the `is_link` condition, you already know you're in the windows and...

I wonder if `islink` can actually return True under windows, so is it possible that this actually is just:
```suggestion
    if WINDOWS:
```
?

I'll try to inspect the cpython source. I don't know enough Windows to tell :D

ewjoachim: CPython under Windows [has an implementation](https://github.com/python/cpython/blob/86ebd5c3fa9ac0fba3b651f1d4abfca79614af5f/Modules/posixmodule.c#L5321) for islink

Though it's unclear if it makes sense to check for symlink if we know we're under windows, as I'm unsure venvs use symlinks under windows.

ewjoachim: The original fix author, @dechamps might have an opinion?

dechamps: > I'm unsure whether venvs use symlinks under windows

They don't unless you explicitly tell `venv` to use symlinks, see [this doc](https://docs.python.org/3/library/venv.html#creating-virtual-environments) and especially the note "While symlinks are supported on Windows, they are not recommended". As long as it's pipx that's creating the venv, we can assume that it's not using symlinks on Windows as pipx does not pass `--symlinks` to `venv`.

I think this change makes sense, and I agree it can probably be simplified to just `if WINDOWS:`. This is probably what I should have written from the start - looks like I was a bit too obsessed with making this somewhat platform-agnostic and ended up breaking things 😢

ewjoachim: Don't beat yourself up too hard, that stuff is too complex for any sane person to navigate with ease. You fixed your case, and wrote some test, nothing more may really be expected from anyone. Sometimes, luck factors in, too. And thanks for chiming in on such a short notice <3

Gitznik: Thank you for the review @ewjoachim and thanks for chiming in @dechamps :) 

Will simplify the if statement and clean up the commend above of it after work :+1:

ewjoachim: (Just to be extremely explicit: given the conditions, I believe it's not worth switching the single `realpath` to `pathlib`, I withdraw my suggestion regarding that part)

gaborbernat (approved):

ewjoachim:
@gaborbernat thanks for merging ! In [this](https://github.com/pypa/pipx/pull/1196#discussion_r1450016331) message, @Gitznik mentionned that they were expecting to change one line, but I think you merged before they had an opportunity to do that (well, the tool was broken, so it's a good decision whatsoever).

Do we want to make a follow-up PR with the suggested simplification, or drop it altogether?

PR #1192, opened 2024-01-05
chore: modernize toml

henryiii:
- [x] I have added an entry to `docs/changelog.md`

## Summary of changes

This is a minor modernization of the pyproject.toml, removing deprecated options. One small change, B904 is restored, and the two failures were fixed - the exceptions are already printed/logged, so "from None" is correct (probably a Python 2 holdover).

Based on <https://learn.scientific-python.org/development/guides/repo-review/?repo=pypa%2Fpipx&branch=main>.

## Test plan

Tested by running

```
pre-commit run -a
```

Issue #1142, opened 2023-12-12
Executing zipapp with Python 3.10 or earlier fails on importing 'tomli'


sgryjp:
**Describe the bug**

Executing the zipapp of pipx with Python 3.10 or earlier fails on trying to import `tomli` which is not included in `pipx.pyz`.

**How to reproduce**

Executing `pipx.pyz` of version 1.3.2 using Python 3.10 or earlier will always reproduce the error. Below is an example output:

```sh
mac:~/Downloads% python3.11 pipx.pyz --version
1.3.3
mac:~/Downloads% python3.10 pipx.pyz --version
Traceback (most recent call last):
  File "/Users/sgryjp/Downloads/pipx.pyz/_bootstrap/__init__.py", line 76, in import_string
  File "/Users/sgryjp/.shiv/pipx.pyz_0fe43a768b3b979cecbbd4b3678299380e71ea146670c0f28fd6199751c78f96/site-packages/pipx/main.py", line 23, in <module>
    from pipx import commands, constants
  File "/Users/sgryjp/.shiv/pipx.pyz_0fe43a768b3b979cecbbd4b3678299380e71ea146670c0f28fd6199751c78f96/site-packages/pipx/commands/__init__.py", line 7, in <module>
    from pipx.commands.run import run
  File "/Users/sgryjp/.shiv/pipx.pyz_0fe43a768b3b979cecbbd4b3678299380e71ea146670c0f28fd6199751c78f96/site-packages/pipx/commands/run.py", line 30, in <module>
    import tomli as tomllib
ModuleNotFoundError: No module named 'tomli'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/sgryjp/Downloads/pipx.pyz/_bootstrap/__init__.py", line 84, in import_string
AttributeError: module 'pipx' has no attribute 'main'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/homebrew/Cellar/python@3.10/3.10.13_1/Frameworks/Python.framework/Versions/3.10/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/opt/homebrew/Cellar/python@3.10/3.10.13_1/Frameworks/Python.framework/Versions/3.10/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/Users/sgryjp/Downloads/pipx.pyz/__main__.py", line 3, in <module>
[...]

Issue #1089, opened 2023-10-16
ENH: Better handling of pipx install error when requested python version not available


corneliusroemer:
**How would this feature be useful?**
Following the example from [here](https://pypa.github.io/pipx/examples/]:
```bash
pipx install --python python3.7 pycowsay
```
I get a rather unwieldy exception that would be better wrapped into something more end user friendly:

```txt
1 ❯ pipx install --python python3.7 pycowsay
Traceback (most recent call last):
  File "/opt/homebrew/bin/pipx", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "/opt/homebrew/Cellar/pipx/1.2.0/libexec/lib/python3.12/site-packages/pipx/main.py", line 863, in cli
    return run_pipx_command(parsed_pipx_args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/pipx/1.2.0/libexec/lib/python3.12/site-packages/pipx/main.py", line 214, in run_pipx_command
    return commands.install(
           ^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/pipx/1.2.0/libexec/lib/python3.12/site-packages/pipx/commands/install.py", line 59, in install
    venv.create_venv(venv_args, pip_args)
  File "/opt/homebrew/Cellar/pipx/1.2.0/libexec/lib/python3.12/site-packages/pipx/venv.py", line 162, in create_venv
    venv_process = run_subprocess(cmd + venv_args + [str(self.root)])
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/pipx/1.2.0/libexec/lib/python3.12/site-packages/pipx/util.py", line 176, in run_subprocess
    completed_process = subprocess.run(
                        ^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.0/Frameworks/Python.framework/Versions/3.12/lib/python3.12/subprocess.py", line 548, in run
    with Popen(*popenargs, **kwargs) as process:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.0/Frameworks/Python.framework/Versions/3.12/lib/python3.12/subprocess.py", line 1026, in __init__
[...]

PR #1201, opened 2024-01-11
Introduce towncrier

### LINKED ISSUE #1161: Manage changelog with `towncrier`?

I think it's time to consider adopting [`towncrier`](https://towncrier.readthedocs.io/en/stable/markdown.html) for keeping the changelog. This would have several advantages, including:

- no merge conflicts due to changelog
- more easily readable, ordered changelog
- ability to describe changes in more detail
- easy changelog creation for contributors and management for developers 
via `towncrier` CLI

Configuring `towncrier` for this project. I'd be willing to work on that.

I'm not against the current way of managing the changelog.

Gitznik:

## Summary of changes

As suggested in https://github.com/pypa/pipx/issues/1161, this introduces towncrier for managing the changelog. @chrysle I hope you're fine with me doing some work on this. I really wanted to stop having conflicts in the changelog ;)

I've updated the changelog for all releases since 1.0, excluding linking the issues.

I've updated the current open changelog entries to be towncrier news fragements.

I've pretty much stuck to the default configuration suggested by towncrier. Please let me know if you have any other preferences.

Closes #1161.

The current open changelog items would be added to the changelog like this:
` towncrier build --version 1.5.0`

### Added

- Allow skipping maintenance tasks during list command ([#1081](https://github.com/pypa/pipx/issues/1081))
- Include  into  (zipapp) so that it can be executed with Python 3.10 or earlier ([#1142](https://github.com/pypa/pipx/issues/1142))
- Introduce towncrier for managing the changelog ([#1161](https://github.com/pypa/pipx/issues/1161))

### Changed

- Update  on scripts using  and no  table following the updated version of PEP 723 ([#1180](https://github.com/pypa/pipx/issues/1180))
- Update `pipx run` on scripts using `/// script` and no `run` table following the updated version of PEP 723 ([#1180](https://github.com/pypa/pipx/issues/1180))

### Fixed

- Raise more user friendly error when provided `--python` version is not found ([#1089](https://github.com/pypa/pipx/issues/1089))
- `pipx run`: Verify whether the script name provided is a file before running it ([#1152](https://github.com/pypa/pipx/issues/1152))
- Avoid repeated exception logging in a few rare cases ([#1192](https://github.com/pypa/pipx/issues/1192))


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,6 +1,6 @@
 <!-- add an 'x' in the brackets below -->
 
-- [ ] I have added an entry to `docs/changelog.md`
+- [ ] I have added an entry to `changelog.d`
```
chrysle, line 3:
```suggestion
- [ ] I have added a news fragment under `changelog.d/` (if the patch affects the end users)
```
Since many have stumbled over this before.

Gitznik:
Just realized this needs a bit more work like updating the CONTRIBUTING.md, so I'll mark this as draft for now.

changelog.d/1192.fixed.md
chrysle, line 1: Probably we should use the issue number in the filename whenever possible?

changelog.d/1180.changed.md
chrysle, line 1: This looks like a duplicate?

changelog.d/1142.added.md
```diff
@@ -0,0 +1 @@
+Include  into  (zipapp) so that it can be executed with Python 3.10 or earlier
```
chrysle, line 1:
```suggestion
Include `tomli` into `pipx.pyz` (zipapp) so that it can be executed with Python 3.10 or earlier
```

Looks like there was some trouble with interpreting backticks?

chrysle (commented):
Thank you for working on this! I had already started this review so I'm submitting regardless of the draft status.

Yes, `CONTRIBUTING.md` will need to be updated. Also, we'll have to refactor our changelog CI process.

pyproject.toml
```diff
@@ -74,6 +74,45 @@ lint.mccabe.max-complexity = 15
...
+
+[[tool.towncrier.type]]
+directory = "changed"
+name = "Changed"
+showcontent = true
+
+[[tool.towncrier.type]]
+directory = "fixed"
+name = "Fixed"
+showcontent = true
```
gaborbernat, line 114: Let's make all this inline tables to keep it more compact.

changelog.d/1192.fixed.md
Gitznik: Absolutely, This PR did not have an issue though I think?

chrysle: Right, my mistake.

.github/PULL_REQUEST_TEMPLATE.md
Gitznik: This would throw off the pre-commit hook, as it checks if a news fragment has been created in the branch. In your opinion - should I get rid of the pre-commit hook, or should I create a new towncrier type for internal changes that will not be added to the changelog? I err on the side of not having the pre-commit hook, as we have this checkmark in the PR template and creating a news fragment just for it to be discarded seems unecessary.

pyproject.toml
Gitznik: I've actually moved to the default fragment types (towncrier has the following default fragment types: feature, bugfix, doc, removal, and misc.), so we don't have to do any configuration at all.
It has the added benefit that this way the towncrier CLI reports which fragment types are available and a short description what they should be used for:
```
-> towncrier create --help

Usage: towncrier create [OPTIONS] FILENAME

  Create a new news fragment.

  Create a new news fragment called FILENAME or pass the full path for a file.
  Towncrier has a few standard types of news fragments, signified by the file
  extension.

  These are:
  * .feature - a new feature
  * .bugfix - a bug fix
  * .doc - a documentation improvement,
  * .removal - a deprecation or removal of public API,
  * .misc - a ticket has been closed, but it is not of interest to users.
```

CHANGELOG.md
```diff
@@ -1,134 +1,185 @@
...
+- Delete directories directly instead of spawning rmdir on Windows
+
+### Fixed
+
+- "Failed to delete" error when using Microsoft Store Python
+- "No pyvenv.cfg file" error when using Microsoft Store Python (#1164)
+
+## [1.3.3](https://github.com/pypa/pipx/tree/1.3.3)
+
+### Fixed
```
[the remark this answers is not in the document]
Gitznik: Switched this to docs

changelog.d/1142.added.md
Gitznik: Yeah that's something to be careful of for sure. Will drop all the fragments once the changelog bump is merged and I can rebase.

.github/PULL_REQUEST_TEMPLATE.md
chrysle: > should I get rid of the pre-commit hook

Yes, I think so. There can also be very minor changes for which creating changelog would be overshooting.

> I err on the side of not having the pre-commit hook, as we have this checkmark in the PR template and creating a news fragment just for it to be discarded seems unecessary.

Well, that checkmark implies, and you could also document it in the contributing guide, that a news fragment is not always needed.

.pre-commit-config.yaml
```diff
@@ -28,3 +28,7 @@ repos:
     - "packaging>=20"
     - "platformdirs>=2.1"
     - "tomli; python_version < '3.11'"
+- repo: https://github.com/twisted/towncrier
```
gaborbernat, line 31: Yeah, let's remove it.

gaborbernat (changes requested):
Also needs https://github.com/pypa/pipx/blob/main/.github/workflows/tests.yml#L134 updated.

Gitznik:
I have mostly copied the `Submitting changes` section from [twine](https://github.com/pypa/twine/blob/main/docs/contributing.rst#submitting-changes). Let me know what you think.

Gitznik:
> Also needs https://github.com/pypa/pipx/blob/main/.github/workflows/tests.yml#L134 updated.

I hope my change should be all that is necessary. Most other repos in pypa I have checked that use towncrier have a manual step to do the changelog build. Not sure if there's a reason for it.

gaborbernat:
> I hope my change should be all that is necessary. Most other repos in pypa I have checked that use towncrier have a manual step to do the changelog build. Not sure if there's a reason for it.

This project wants to avoid that. PyPA is not a unified project set, we each do what works best for us.

gaborbernat:
You have some merge conflicts to resolve first 👍

gaborbernat (dismissed):

gaborbernat (changes requested):
Can you add a changelog entry so we can see it showing up here https://pipx--1201.org.readthedocs.build/1201/changelog

Also I don't want to see this:

```
This project uses [towncrier](https://towncrier.readthedocs.io/) and the changes for the upcoming release can be found in https://github.com/twisted/my-project/tree/main/changelog.d/.
```
on the changelog entry. 

Instead, we should change the doc generation to include it in-line. That way we can see how the changelog will look post release as part of the PR.


See virtualenv doing this https://github.com/pypa/virtualenv/blob/main/docs/conf.py#L54

CONTRIBUTING.md
```diff
@@ -3,6 +3,42 @@ Thanks for your interest in contributing to pipx!
...
+4. Submit a pull request to the `main` branch on GitHub, referencing an
+   open issue.
+
+### Changelog entries
+
+The `CHANGELOG.md` file is built by
+[towncrier](https://pypi.org/project/towncrier/) from files in the
+`changelog.d/` directory. To add an entry, create a file in that directory
+named `{number}.{type}.md`, where `{number}` is the issue number,
+and `{type}` is `feature`, `bugfix`, `doc`, `removal`, or `misc`.
```
chrysle, line 21:
```suggestion
and `{type}` is one of `feature`, `bugfix`, `doc`, `removal`, or `misc`.
```

CONTRIBUTING.md
```diff
@@ -3,6 +3,42 @@ Thanks for your interest in contributing to pipx!
...
+2. Make a branch off of `main` and commit your changes to it.
+3. Add a changelog entry.
+4. Submit a pull request to the `main` branch on GitHub, referencing an
+   open issue.
+
+### Changelog entries
+
+The `CHANGELOG.md` file is built by
+[towncrier](https://pypi.org/project/towncrier/) from files in the
+`changelog.d/` directory. To add an entry, create a file in that directory
```
chrysle, line 19:
```suggestion
[towncrier](https://pypi.org/project/towncrier/) from news fragments in the
`changelog.d/` directory. To add an entry, create a news fragment in that directory
```

.github/workflows/tests.yml
```diff
@@ -139,7 +139,7 @@ jobs:
       - name: Create pull request branch
         run: git switch -c "bump-changelog-for-${RELEASE_VERSION}"
       - name: Update changelog
-        run: echo -e "## dev\n\n## $RELEASE_VERSION\n$(tail -n +2 CHANGELOG.md)" > CHANGELOG.md
+        run: pipx run towncrier build --version $RELEASE_VERSION
```
chrysle, line 142:
```suggestion
        run: pipx run towncrier build --yes --version $RELEASE_VERSION
```

noxfile.py
```diff
@@ -112,6 +112,7 @@ def build_docs(session: nox.Session) -> None:
     site_dir = session.posargs or ["site/"]
     session.install(*DOC_DEPENDENCIES, ".")
     session.env["PIPX__DOC_DEFAULT_PYTHON"] = "typically the python used to execute pipx"
+    session.run("towncrier", "build", "--version", "Next", "--keep")
```
chrysle, line 115:
```suggestion
    session.run("towncrier", "build", "--draft", "--version", "Next", "--keep")
```

Why should we modify the changelog?

Gitznik: If we don't,  the latest news fragments would not show up in the docs built, would they?

chrysle: Right, but the changelog is not the most important part of the documentation, so wouldn't it be enough to write to standard output? The risk of accidentally committing you mentioned would be gone with that.

Or we might add a `git restore` afterwards?

Gitznik: @gaborbernat what's your opinion on this?

`git restore` seems dangerous to me in case someone actually does a fix in the changelog file. That would make them lose their changes.

Alternatively there's a mkdocs plugin to embed a markdown file in another. Using this, we could store the new changelog entries in a separate file with the towncrier build,  that is not checked into source control. Then we'd just embed it into the changelog. That's pretty much what `virtualenv` does in their sphinx configuration.

gaborbernat: Yeah the later is the path ahead.

Gitznik: There's a couple of options for mkdocs plugins that allow this. I went for https://github.com/fralau/mkdocs-macros-plugin, because it has the most stars and comes with a lot of nice functionality that might be helpful in the future. It also had some of the easiest implementation for including the draft file.

Gitznik: Where are the "live" docs served from? Are they also created with the build_docs nox command? If so, do I somehow need to check in the build step whether we're running on main or on a feature branch? Or do we want to add the upcoming news fragments to the `prod` changelog under a special section like `Upcoming` or `Planned for next Release`?

chrysle: > Where are the "live" docs served from? Are they also created with the build_docs nox command?

No, there's a separate session for that:

https://github.com/pypa/pipx/blob/4dd1b6b84cbc2dc14e958ff2223c7f3388982e98/noxfile.py#L119

> Or do we want to add the upcoming news fragments to the `prod` changelog under a special section like `Upcoming` or `Planned for next Release`?

Sounds reasonable.

Gitznik: I've adapted it to show the upcoming changes separately at the top with a expressive header, both in dev and prod.

One thing to note is that the `_draft_changelog.md` file is copied into the sites directory. So while it's not linked anywhere, it is technically available if one enters it's URL directly. IMO that's fine. let me know if you disagree. Then I'll add deleting it from there to the nox command.

gaborbernat: We should delete it.

Gitznik: > No, there's a separate session for that:
> 
> https://github.com/pypa/pipx/blob/4dd1b6b84cbc2dc14e958ff2223c7f3388982e98/noxfile.py#L119

@chrysle I'm quite confused by this. I can't see `watch_docs` be used anywhere in the repository. Can you give me a bit more detail on how it is used to serve our `prod` docs?

chrysle: > Can you give me a bit more detail on how it is used to serve our `prod` docs?

There isn't much to say about it. It builds the directory tree at it current state and serves it, rebuilding whenever a change is detected. So I think your modifications are right.

Gitznik: > There isn't much to say about it. It builds the directory tree at it current state and serves it, rebuilding whenever a change is detected. So I think your modifications are right.

I assume we have a misunderstanding here, but correct me if I'm wrong.

- `build_docs` are used for serving the docs on `readthedocs`, including our production docs at https://pipx.pypa.io/
- `watch_docs` is for live previewing the docs locally during development.

Is that accurate?

If so - for `build_docs` I now remove the `_draft_changelog` from the site directory after building. For `watch_docs` that's not really easy to do because it builds and then serves immediately in a blocking way. No chance for me to delete the file in between. But if that is only used for dev - I think it does not matter. Might even be useful for debugging the behavior on dev.

chrysle: > Is that accurate?

Yes.

> But if that is only used for dev - I think it does not matter. Might even be useful for debugging the behavior on dev.

I also think so.

gaborbernat (approved):

Gitznik:
Thanks @gaborbernat @chrysle with your thorough support and suggestions for getting this out there :) Much appreciated!

gaborbernat:
Thanks for making it happen 👍

chrysle:
> Thanks for making it happen 👍

I'm going along with that!

gaborbernat:
This did fail in the latest release, see https://github.com/pypa/pipx/actions/runs/7544497267/job/20539252092 @Gitznik can you take a look please?

chrysle:
Probably running `towncrier` with `pipx` is a problem?

Gitznik:
Yeah I'm testing installing it first instead of running directly. That fixed the problem for me locally.

PR #1204, opened 2024-01-12
Add macos shebang workaround description

### LINKED ISSUE #1198: $PIPX_HOME on macOS since release 1.3.0 breaks awscli
It appears that since release 1.3.0, when installing pipx on macOS Ventura, it chooses to set `$PIPX_HOME` to `~/Library/Application Support/pipx`, which creates a problem for scripts like `awscli` where the directory forms part of the shebang, because shebangs are not allowed to have spaces in it.

As recent as release 1.2.1, $PIPX_HOME on macOS Ventura is set to be `~/.local/pipx` which doesn't cause problems.

**How to reproduce**

I was using Python 3.8 installed using pyenv.
```
pip install pipx==1.3.0
pipx environment 
```
It prints the below:
Screenshot 2024-01-08 at 18 43 56

**Expected behavior**

`$PIPX_HOME` should be `/Users/test3/.local/pipx`

Gitznik:

## Summary of changes

Add docs for shebang path issues on macOS according to https://github.com/pypa/pipx/issues/1198

closes #1198

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -96,14 +96,15 @@ Reference:
 
 ## macOS issues
 
-If you want to use a Pipx installed package in a shebang (a common example is the AWS cli),
-you will likely not be able to, because the binary will be under `~/Library/Application Support/pipx`.
+If you want to use a Pipx-installed package in a shebang (a common example is the AWS CLI),
+you will likely not be able to, because the binary will be stored under `~/Library/Application Support/pipx/`.
 The space in the path is not supported in a shebang. A simple solution is symlinking
 `~/Library/Application Support/pipx` to `~/Library/ApplicationSupport/pipx`, and using that as the
 path in the shebang instead.
 
 ```
-ln -s $HOME/Library/Application Support/pipx $HOME/Library/ApplicationSupport/pipx
+mkdir $HOME/Library/ApplicationSupport
+ln -s $HOME/Library/Application\ Support/pipx $HOME/Library/ApplicationSupport/pipx
 ```
 
 ## Does it work to install your package with `pip`?
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -94,6 +94,18 @@ sudo apt install python3-venv python3-pip
 Reference:
 [Python Packaging User Guide: Installing pip/setuptools/wheel with Linux Package Managers](https://packaging.python.org/guides/installing-using-linux-tools)
 
+## macOS issues
+
+If you want to use a Pipx installed package in a shebang (a common example is the AWS cli),
+you will likely not be able to, because the binary will be under `~/Library/Application Support/pipx`.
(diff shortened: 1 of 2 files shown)


gaborbernat (approved):

docs/troubleshooting.md
```diff
@@ -94,6 +94,18 @@ sudo apt install python3-venv python3-pip
...
+## macOS issues
+
+If you want to use a Pipx installed package in a shebang (a common example is the AWS cli),
+you will likely not be able to, because the binary will be under `~/Library/Application Support/pipx`.
+The space in the path is not supported in a shebang. A simple solution is symlinking
+`~/Library/Application Support/pipx` to `~/Library/ApplicationSupport/pipx`, and using that as the
+path in the shebang instead.
+
+```
+ln -s $HOME/Library/Application Support/pipx $HOME/Library/ApplicationSupport/pipx
```
dukecat0, line 106:
```suggestion
mkdir $HOME/Library/ApplicationSupport
ln -s $HOME/Library/Application\ Support/pipx $HOME/Library/ApplicationSupport/pipx
```

dukecat0: The original command doesn't work properly.

dukecat0 (changes requested):

docs/troubleshooting.md
```diff
@@ -94,6 +94,18 @@ sudo apt install python3-venv python3-pip
 Reference:
 [Python Packaging User Guide: Installing pip/setuptools/wheel with Linux Package Managers](https://packaging.python.org/guides/installing-using-linux-tools)
 
+## macOS issues
+
+If you want to use a Pipx installed package in a shebang (a common example is the AWS cli),
+you will likely not be able to, because the binary will be under `~/Library/Application Support/pipx`.
```
chrysle, line 100:
```suggestion
If you want to use a Pipx-installed package in a shebang (a common example is the AWS CLI),
you will likely not be able to, because the binary will be stored under `~/Library/Application Support/pipx/`.
```

docs/troubleshooting.md
Gitznik: Oh damn, I should have tried it in a mac VM. Thanks for verifying!

dukecat0 (approved):

jaraco:
As pointed out in [this comment](https://sourceforge.net/p/docutils/bugs/479/#1403), the issue isn't specific to macOS. It affects all Unix systems because [Unix is hostile to spaces in filenames](https://blog.jaraco.com/unix-doesnt-support-spaces-in-filenames/). It just _happens_ to affect pipx users on macOS because pipx has opted to use the recommended platform dirs for mac that include spaces.

Also, as pointed out in the docutils comment, this documented workaround is slightly terrible. It requires mutating the install on every package install or update. It also requires polluting the user's `~/Library`, which affects tab completion and is just ugly.

PR #1211, opened 2024-01-16
docs/comparisons: Add pae/pactivate

0cjs:

(Does not affect users of the program; documentation update only.)

## Summary of changes

Add comparison with a somewhat similar tool, [pae/pactivate](https://github.com/cynic-net/pactivate).

## Test plan

Tested by regenerating the documentation with `nox -s build_docs` and reading the new output in `site/comparisons/index.html`.


docs/comparisons.md
```diff
@@ -111,3 +111,19 @@ virtualenvs with their dependencies installed.
 - fades can only run individual script files while pipx can also run packages.
 
 Example interaction: None.
+
+## pipx vs pae/pactivate
+
+_pae_ is a Bash command-line function distributed with [pactivate](https://github.com/cynic-net/pactivate) that uses pactivate to create non-ephermeral environments that can be used for both command-line applications and libraries.
+
+- pae and pipx both create isolated environments without having to specify (and remember) a directory in which to store them.
+- pae allows you to use any Python interpreter available on your system, so you can use arbitrary versions of Python.
```
gaborbernat, line 120: Same for pipx.

docs/comparisons.md
```diff
@@ -111,3 +111,19 @@ virtualenvs with their dependencies installed.
...
+- pae does not require you to have anything other than a Python interpreter available; it downloads and bootstraps `pip` and `virtualenv`, always using the latest releases of both.
```
gaborbernat, line 121: Same for pipx, assuming your using the zipapp packaging.

docs/comparisons.md
```diff
@@ -111,3 +111,19 @@ virtualenvs with their dependencies installed.
...
+- pae and pipx both make it simple to run commands in the isolated environments, but pae does not add these commands to your PATH by default. You must use `pae COMMAND` or `pae ENV COMMAND` to run them, or `pae -a ENV` to activate the environment.
+- pae allows you to activate and deactivate an isolated environment so you you can use libraries within it.
```
Gitznik, line 123:
```suggestion
- pae allows you to activate and deactivate an isolated environment so you can use libraries within it.
```

docs/comparisons.md
```diff
@@ -111,3 +111,19 @@ virtualenvs with their dependencies installed.
 - fades can only run individual script files while pipx can also run packages.
 
 Example interaction: None.
+
+## pipx vs pae/pactivate
+
+_pae_ is a Bash command-line function distributed with [pactivate](https://github.com/cynic-net/pactivate) that uses pactivate to create non-ephermeral environments that can be used for both command-line applications and libraries.
```
Gitznik, line 117:
```suggestion
_pae_ is a Bash command-line function distributed with [pactivate](https://github.com/cynic-net/pactivate) that uses pactivate to create non-ephemeral environments that can be used for both command-line applications and libraries.
```

docs/comparisons.md
```diff
@@ -111,3 +111,19 @@ virtualenvs with their dependencies installed.
...
+
+- pae and pipx both create isolated environments without having to specify (and remember) a directory in which to store them.
+- pae allows you to use any Python interpreter available on your system, so you can use arbitrary versions of Python.
+- pae does not require you to have anything other than a Python interpreter available; it downloads and bootstraps `pip` and `virtualenv`, always using the latest releases of both.
+- pae and pipx both make it simple to run commands in the isolated environments, but pae does not add these commands to your PATH by default. You must use `pae COMMAND` or `pae ENV COMMAND` to run them, or `pae -a ENV` to activate the environment.
+- pae allows you to activate and deactivate an isolated environment so you you can use libraries within it.
+- With pae, packages must be manually installed in an environment with `pae ENV pip install ...`.
+- pae does not create temporary environments; they must be manually created with `pae -c` and removed with `pae --rm` when no longer needed.
+- pae is significantly slower to create virtual environments due to the bootstrap process.
```
Gitznik, line 126: Is there a benchmark you can link here?

Gitznik (changes requested):
Thanks for contributing to the docs :)

docs/comparisons.md
0cjs: Actually, it's not quite the same. `pipx` can still fail in more cases on OS-supplied "minimal" Python interpreter installations. For example, on my Debian 12 system, `python3 ~/Downloads/pipx.pyz  install grip` fails with:

```
The virtual environment was not created successfully because ensurepip is not
available.  On Debian/Ubuntu systems, you need to install the python3-venv
package using the following command.

    apt install python3.11-venv

You may need to use sudo with that command.  After installing the python3-venv
package, recreate your virtual environment.
```

(pae downloads and installs the latest `virtualenv` package, and so does not rely on the Python installation providing that, as mine doesn't above. I'd not realised that you're using the venv subset of virtualenv instead; I've added a note about that to the comparison.)

That's not to say that pae is perfect in this regard; it runs into a similar issue on Debian 10 and below, where I print a similar message when the OS is lacking `python3-distutils`.

docs/comparisons.md
0cjs: I've not done any formal benchmarking, but I did some quick comparisons. The very first time you use pipx to install something it takes about the same time as pae. But after you've done your first pipx install and `.local/share/pipx/shared` is set up, pipx time is the same as just the `pip install packagename` time, about 8 seconds faster on my slowish desktop machine.

But I'm open to your suggestions about whether this note should even be included. I personally feel it's a fairly notable difference, but then again I'm tending to create/delete virtual environments a lot with pae; other pae users that simply install more programs into existing virtual environments wouldn't have such a difference.

0cjs:
Note that I'm putting up separate commits on the dev branch for each separate conversation here; I believe this should make reviewing easier. I'll of course squash these all into one commit when we're ready to merge, unless you guys think that this history here should be kept.

docs/comparisons.md
Gitznik: If not based on measurements, can you quickly outline the technical reason why pipx is faster in this case compared to pae?

0cjs:
I've just added a new bullet point near the top that I think more precisely explains a core difference in approach between pipx and pae.  I realise this whole thing may be getting a bit long; I'm open to removing whatever you think isn't too important. (I could also add a section in the pae/pactivate documentation with a detailed comparison with pipx, and we could just summarize here and link to that, if that sounds like a better idea.)

0cjs:
By the way, I did play with the `pipx inejct` command to inject other packages into a virtual environment, but it didn't create links in my `~/.local/bin/` for the CLI programs in it. And actually even programs for dependencies don't get links. (I do believe that both these decisions are correct for pipx.) Is there any way with pipx to run an arbitrary program from a specific virtual environment other than giving the full path to it, e.g., `~/.local/share/pipx/venvs/grip/bin/flask --version`?

chrysle:
> (I do believe that both these decisions are correct for pipx.) 

Yes. I asked for a corresponding option for `pipx install`, but did not receive an approval. 

> Is there any way with pipx to run an arbitrary program from a specific virtual environment other than giving the full path to it, e.g., `~/.local/share/pipx/venvs/grip/bin/flask --version`?

Unfortunately, no. You could use a virtual environment manager and configure it to use the `pipx` venv path. And you can always use `pipx run`, of course.

docs/comparisons.md
0cjs: `pipx install` is usually faster (with the notable exception of the first run) because once `.local/share/pipx/shared/` is set up it uses (as far as I can tell) the existing pip and venv there to set up a new virtual environment. `pae -c`, by contrast, always bootstraps with a newly downloaded version of pip and virtualenv before creating the actual virtual environment.

I'm not going to update this PR with this yet, because the deeper I get into examining the design decisions of pipx and comparing them with my own for pae, the more interesting things get, and the more I feel the scope of a write-up on this is exceeding what should be on the comparison page. So my tentative plan at the moment is to do a much more detailed comparison write-up for the pae documentation, and then reduce my proposed entry `docs/comparison.md` to a summary of that, with a link to the full version. (I'm open to other ideas about this, of course.)

0cjs:
Ok, I've created a [detailed comparison](https://github.com/cynic-net/pactivate/blob/main/doc/vs-pipx.md) in the pactivate documentation, and slimmed down the one here to cover just the essentials and link to the details. Thoughts?

(As usual, this is a separate commit on the end; a squash will bring the whole set down to just what's on the end.)

gaborbernat (approved):

Gitznik (approved):

1cjs:
Thanks!

PR #1213, opened 2024-01-16
Fix towncrier build

Gitznik:

## Summary of changes

This fixes the error in the pipeline as reported in https://github.com/pypa/pipx/pull/1201#issuecomment-1894146048

@gaborbernat 

running `pipx install towncrier` and then running `towncrier build` fixed it for me locally, but not in the pipeline.

## Test plan

See https://github.com/Gitznik/pipx/actions/runs/7545283319/job/20540558260

Tested by running

```
```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -134,6 +134,11 @@ jobs:
     steps:
       - name: Checkout ${{ github.ref }}
         uses: actions/checkout@v4
+      - name: Set up Python ${{ env.default-python }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ env.default-python }}
+          cache: "pip"
       - name: Extract release tag
         run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
       - name: Create pull request branch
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -139,7 +139,9 @@ jobs:
       - name: Create pull request branch
         run: git switch -c "bump-changelog-for-${RELEASE_VERSION}"
       - name: Update changelog
-        run: pipx run towncrier build --yes --version $RELEASE_VERSION
+        run: |
```
gaborbernat, line 142: I would prefer this to be a nox target invocation, in case Github is down, I still want to perform this work locally.

Gitznik: Done :+1: See it running in my fork here: https://github.com/Gitznik/pipx/actions/runs/7546614725/job/20544738799

.github/workflows/tests.yml
```diff
@@ -139,7 +139,9 @@ jobs:
...
+          pip install --user towncrier
```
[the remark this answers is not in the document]
Gitznik: According to @gaborbernat's suggestion I've moved the invocation and the install into nox :+1:

gaborbernat (approved):

gaborbernat:
@Gitznik can you put in one more PR to clean up this failed batch https://github.com/pypa/pipx/tree/main/changelog.d :-D from the last release?

PR #1215, opened 2024-01-16
bump changelog to 1.4.3

Gitznik:

## Summary of changes

Bumped the changelog to 1.4.3, as the automated bump failed. @gaborbernat 

## Test plan

Tested by running

```
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,21 @@ This project uses [*towncrier*](https://towncrier.readthedocs.io/) for keeping t
 
 <!-- towncrier release notes start -->
 
+## [1.4.3](https://github.com/pypa/pipx/tree/1.4.3) - 2024-01-16
+
+
+### Bugfixes
+
+- Autofix python version for pylauncher, when version is provided prefixed with `python` ([#1150](https://github.com/pypa/pipx/issues/1150))
+- Support building pipx wheels with setuptools-scm<7, such as on FreeBSD. ([#1208](https://github.com/pypa/pipx/issues/1208))
+
+### Improved Documentation
+
+- Provide useful error messages when unresolvable python version is passed ([#1150](https://github.com/pypa/pipx/issues/1150))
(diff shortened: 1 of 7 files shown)


gaborbernat:
Seems readthedocs failed with this 🤔

Gitznik:
> Seems readthedocs failed with this 🤔

Yeah looks like it does not like it if there's no news fragment. Will add a fix.

Gitznik:
> > Seems readthedocs failed with this 🤔
> 
> Yeah looks like it does not like it if there's no news fragment. Will add a fix.

That fixed it. While looking at it I saw that the pipx inside pipx note doesn't seem to be rendered correctly. Is that expected in dev?
image

gaborbernat:
> Is that expected?

Sounds like a bug to me.

chrysle:
> While looking at it I saw that the pipx inside pipx note doesn't seem to be rendered correctly. Is that expected?

Let's use the [note admonition](https://squidfunk.github.io/mkdocs-material/reference/admonitions/#usage) instead.

gaborbernat (approved):

Gitznik:
> > While looking at it I saw that the pipx inside pipx note doesn't seem to be rendered correctly. Is that expected?
> 
> Let's use the [note admonition](https://squidfunk.github.io/mkdocs-material/reference/admonitions/#usage) instead.

It already uses that: https://github.com/pypa/pipx/blob/2cf78d2c50887e222ac34d246abaad6bebe61540/docs/installation.md?plain=1#L71

chrysle:
> It already uses that:

Oh, we're talking about different documents, then. I meant the note on the front page. The one in `installation.md` renders correctly.

gaborbernat:
Yeah we should fix that but it sounds orthogonal to this pull request.

Gitznik:
> > It already uses that:
> 
> Oh, we're talking about different documents, then. I meant the note on the front page. The one in `installation.md` renders correctly.

Shouldn't try to debug something while in a hurry. Thanks for clarifying. Will create a separate issue :+1:

PR #1216, opened 2024-01-17
Switch to note admonition in readme

Gitznik:

## Summary of changes

The current syntax does not render correctly and is not in line with the rest of the docs. I could not find any more usages of the other syntax in the docs.

See https://github.com/pypa/pipx/pull/1215 for the source of this PR.

## Test plan

Tested by running `nox -s build_docs` and checking `site/index.html`

```
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -126,19 +126,21 @@ To run unit tests in Python3.12, you can run
 nox -s tests-3.12
 ```
 
-!!! tip You can run a specific unit test by passing arguments to pytest, the test runner pipx uses:
-
-    ```
-    nox -s tests-3.8 -- -k EXPRESSION
-    ```
-
-    `EXPRESSION` can be a test name, such as
-
-    ```
-    nox -s tests-3.8 -- -k test_uninstall
-    ```
-
-    Coverage errors can usually be ignored when only running a subset of tests.
+> [!TIP]
+>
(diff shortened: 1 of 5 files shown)


dukecat0:
Seems like it only works on mkdocs, but not Github.

image

chrysle:
> Seems like it only works on mkdocs, but not Github.

Is that a problem? We don't attend a beauty contest for GitHub-hosted docs, do we? I always thought that's what a website is for.

henryiii:
Github does have a syntax for admonitions, and it doesn't match anything else. I wonder if it could be supported via a plugin, though?

* https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts
* https://github.com/orgs/community/discussions/16925

henryiii:
Ahh, found it in the third-party extensions page for Python-Markdown: https://github.com/dahlia/markdown-gfm-admonition

(From https://github.com/Python-Markdown/markdown/wiki/Third-Party-Extensions#additional-tags)

henryiii:
Screenshot 2024-01-18 at 10 57 51 AM

dukecat0 (approved):

dukecat0:
@henryiii Thanks a lot for fixing this!

Gitznik:
Great find @henryiii , thanks for taking this over 👌

PR #1224, opened 2024-01-20
Update the completion instructions for zipapp users

### LINKED ISSUE #1072: Support shell completion with pipx.pyz

Users of pipx.pyz don't get shell completion for pipx. The current instructions printed by `pipx completions` don't work because the `register-python-argcomplete` command is missing. This is a limitation compared to "full" installations of pipx made with pip install --user, apt, brew, etc.

To be specific, I mean users who move pipx.pyz to `~/.local/bin/pipx` (or similar) and mark it executable. I think it's probably impossible to support completion for users who run it explicitly as `python3 /path/to/pipx.pyz ...`, so they are out of scope.

It turns out this is already possible. It just needs to be documented in `pipx completions`. Instead of
```sh
eval "$(register-python-argcomplete pipx)"
```
your bash config file must contain:
```sh
eval "$(SHIV_CONSOLE_SCRIPT=register-python-argcomplete pipx pipx 2>/dev/null)"
```

I only tested bash. Someone please test the other shells.

Explanation: SHIV_CONSOLE_SCRIPT tells the shiv bootstrap code to use another script bundled inside pipx.pyz. The first `pipx` is the program to run. The second `pipx` is an argument to register-python-complete (telling it what to generate completions for). The `2>/dev/null` is necessary to work around linkedin/shiv#240, at least for now.

Another option could be to stop using argcomplete and rewrite the whole pipx CLI code with click. pipx already depends on click indirectly because of userpath. click's [completion support](https://click.palletsprojects.com/en/latest/shell-completion/) does not depend on other programs, so it works nicely with pyz files and doesn't need SHIV_CONSOLE_SCRIPT. However all existing users of completion would have to change their configs.

I don't use pipx completion and I don't personally care about this issue. I'm just filing it because I noticed it doesn't already exist.

dukecat0:

## Summary of changes
Update the completion instructions for zipapp users.
Closes #1072 
## Test plan

Tested by running

```
pipx install argcomplete
pipx completions
eval "$(register-python-argcomplete pipx)"
```


chrysle (approved):

chrysle (approved):

anentropic:
I have pipx installed via Homebrew and I needed to do this step too

I have had pipx installed for ages but I guess it got updated recently

after a reboot and subsequent restart of my terminal app I started getting:
```
.zshrc:29: command not found: register-python-argcomplete
```

`pipx install argcomplete` fixed it for me

Issue #1203, opened 2024-01-12
Pipx managing itself is broken on windows


guahki:
**Describe the bug**

When pipx manages itself (pipx is installed as a pipx app) on Windows, it is no longer able to uninstall itself (since #1168 was merged).

The following error message is shown:

```
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\not_of_interest\.local\bin\pipx.exe\__main__.py", line 7, in <module>
  File "C:\Users\not_of_interest\.local\pipx\venvs\pipx\Lib\site-packages\pipx\main.py", line 914, in cli
  File "C:\Users\not_of_interest\.local\pipx\venvs\pipx\Lib\site-packages\pipx\main.py", line 274, in run_pipx_command
  File "C:\Users\not_of_interest\.local\pipx\venvs\pipx\Lib\site-packages\pipx\commands\reinstall.py", line 53, in reinstall
  File "C:\Users\not_of_interest\.local\pipx\venvs\pipx\Lib\site-packages\pipx\commands\uninstall.py", line 151, in uninstall
  File "C:\Users\not_of_interest\.local\pipx\venvs\pipx\Lib\site-packages\pipx\util.py", line 57, in rmdir
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.12_3.12.496.0_x64__qbz5n2kfra8p0\Lib\shutil.py", line 808, in rmtree
    return _rmtree_unsafe(path, onexc)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.12_3.12.496.0_x64__qbz5n2kfra8p0\Lib\shutil.py", line 631, in _rmtree_unsafe
    _rmtree_unsafe(fullname, onexc)
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.12_3.12.496.0_x64__qbz5n2kfra8p0\Lib\shutil.py", line 636, in _rmtree_unsafe
    onexc(os.unlink, fullname, err)
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.12_3.12.496.0_x64__qbz5n2kfra8p0\Lib\shutil.py", line 634, in _rmtree_unsafe
    os.unlink(fullname)
[...]

PR #1231, opened 2024-01-25
Let self-managed pipx uninstall itself on windows again

guahki:

## Summary of changes

Change deletion of directories to ignore errors when ``safe_rm`` is true, such that locked binaries on windows can be moved to the trash afterwards.

## Test plan

Tested by installing pipx into pipx on windows and then running

```
pipx uninstall pipx
```

which failed before and now works


Gitznik (changes requested):
Can you verify this behavior in a test please?

guahki:
I can easily add some basic tests, trying to delete a non-existing folder and an existing one.

**But** this would not really test the "gist" of this function, the ``safe_rm`` feature, which moves files to the "trash" on windows, if they are locked (e.g. due to being a running executable). I'm not sure, if this is even testable and tbh I'm not willing to spend lot's of time finding a solution, how to test this. At least, I added a comment (as also suggested in the last paragraph of https://github.com/pypa/pipx/issues/1203#issuecomment-1913557486).

I understand, that adding new tests with each PR improves the tests and test coverage. But to request a test from me for a "bugfix" like this (obviously no test was requested, when this was broken in #1168; and most likely the test would not have found the issue, due to the facts described above) seems a little hard for me.

So: I can assist with providing "basic" tests, which prove nothing (or very little) and most likely do not catch newly introduced bugs, if you want me to. But I would see it as timewaste tbh. If you want real tests, I have to ask for assistance from somebody with the time to think about, how the locked-file scenario can be tested properly.

Gitznik:
> I understand, that adding new tests with each PR improves the tests and test coverage. But to request a test from me for a "bugfix" like this (obviously no test was requested, when this was broken in #1168; and most likely the test would not have found the issue, due to the facts described above) seems a little hard for me.

The point is not necessarily adding a test with each PR, it's adding a test with every bugfix to make sure we're not regressing. The point that no test was added in #1168 is not an argument against writing tests, it is a lesson that we should've done it. 

I appreciate this might not be easy to test. I will try and look into it a little, but as I'm not on windows it's a little cumbersome for me to play around with it. I appreciate any windows user that is willing to spend some brain power on this as well.

Should we not be able to find a way to test this, IMO we can merge it, but we should do our due diligence first trying to find a meaningful test.

guahki:
> The point is not necessarily adding a test with each PR, it's adding a test with every bugfix to make sure we're not regressing. The point that no test was added in #1168 is not an argument against writing tests, it is a lesson that we should've done it.

Yes, I get that.

> I appreciate this might not be easy to test. I will try and look into it a little, but as I'm not on windows it's a little cumbersome for me to play around with it. I appreciate any windows user that is willing to spend some brain power on this as well.

Would be very much thankful for any help here, as I unfortunately most likely will not soon have the time to develop a sensible test strategy and write sensible testcases for this one.

> Should we not be able to find a way to test this, IMO we can merge it, but we should do our due diligence first trying to find a meaningful test.

👍

Gitznik:
I tried a little bit but had a hard time reproducing the behavior in a test as well.

Manually, @guahki verified the behavior on windows, I did on Linux. I think this is ready for review.

Gitznik (approved):

guahki:
> I tried a little bit but had a hard time reproducing the behavior in a test as well.

Thanks for taking the time and trying! I think one would have to have a running executable and try to delete that. Not an easy task.

> Manually, @guahki verified the behavior on windows, I did on Linux. I think this is ready for review.

Thank you for all the moderation and help. Please let me know, if there is anything to do for me on this PR.

Issue #1205, opened 2024-01-13
``--python`` flag is broken on Windows


guahki:
**Describe the bug**

The ``--python`` flag used to work with a executable name in the PATH. E.g. ``pipx install --python python3.12.exe ...`` worked, if ``python3.12.exe`` was found on the path.

Since #1168, however, this does no longer work. See analysis below.

```
$ where python3.12.exe
C:\Users\Dominik\AppData\Local\Microsoft\WindowsApps\python3.12.exe
$ pipx install --python python3.12.exe flit
creating virtual environment...

Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\not_of_interest\Scripts\pipx.exe\__main__.py", line 7, in <module>
  File "C:\not_of_interest\Lib\site-packages\pipx\main.py", line 911, in cli
    return run_pipx_command(parsed_pipx_args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\not_of_interest\Lib\site-packages\pipx\main.py", line 212, in run_pipx_command
    return commands.install(
           ^^^^^^^^^^^^^^^^^
  File "C:\not_of_interest\Lib\site-packages\pipx\commands\install.py", line 83, in install
    venv.create_venv(venv_args, pip_args, override_shared)
  File "C:\not_of_interest\Lib\site-packages\pipx\venv.py", line 160, in create_venv
    venv_process = run_subprocess(cmd + venv_args + [str(self.root)])
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\not_of_interest\Lib\site-packages\pipx\util.py", line 179, in run_subprocess
    completed_process = subprocess.run(
                        ^^^^^^^^^^^^^^^
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.12_3.12.496.0_x64__qbz5n2kfra8p0\Lib\subprocess.py", line 548, in run
    with Popen(*popenargs, **kwargs) as process:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.12_3.12.496.0_x64__qbz5n2kfra8p0\Lib\subprocess.py", line 1026, in __init__
[...]

Issue #1164, opened 2023-12-23
pipx unable to create venv when using Python from Microsoft Store


dechamps:
On my Windows 11 23H2 (22631.2661) system, with Python 3.12.1 [installed from the Microsoft Store](https://www.microsoft.com/store/productId/9NCVDN91XZQP?ocid=pdpshare), and following the [pipx Windows installation instructions](https://pipx.pypa.io/stable/#on-windows-install-via-pip-requires-pip-190-or-later), this is what happens when I try to install anything:

```
PS C:\Users\etien> pipx install --verbose pycowsay
pipx >(setup:824): pipx version is 1.3.3
pipx >(setup:825): Default python interpreter is 'C:\Users\etien\AppData\Local\Microsoft\WindowsApps\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\python.exe'
pipx >(rmdir:55): removing directory C:\Users\etien\AppData\Local\pipx\pipx\trash
The system cannot find the path specified.
pipx >(rmdir:71): Failed to delete C:\Users\etien\AppData\Local\pipx\pipx\trash. You may need to delete it manually.
pipx >(package_name_from_spec:370): Determined package name: pycowsay
pipx >(package_name_from_spec:371): Package name determined in 0.0s
creating virtual environment...
pipx >(run_subprocess:171): running C:\Users\etien\AppData\Local\Microsoft\WindowsApps\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\python.exe -m venv --without-pip C:\Users\etien\AppData\Local\pipx\pipx\venvs\pycowsay
pipx >(run_subprocess:171): running <checking pip's availability>
pipx >(run_subprocess:171): running C:\Users\etien\AppData\Local\pipx\pipx\venvs\pycowsay\Scripts\python.exe -c import sysconfig; print(sysconfig.get_path('purelib'))
No pyvenv.cfg file
pipx >(run_subprocess:171): running C:\Users\etien\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\Local\pipx\pipx\shared\Scripts\python.exe -c import sysconfig; print(sysconfig.get_path('purelib'))
pipx >(run_subprocess:171): running C:\Users\etien\AppData\Local\pipx\pipx\venvs\pycowsay\Scripts\python.exe --version
pipx >(_parsed_package_to_package_or_url:137): cleaned package spec: pycowsay
[...]

PR #1232, opened 2024-01-25
Fix path resolution for python executables looked up in PATH on windows

guahki:

## Summary of changes

Only use the logic introduced in https://github.com/pypa/pipx/pull/1168 to fix a path resolution issue on windows when the given path already is absolute. This avoids prepending the cwd to relative paths.
This fixes https://github.com/pypa/pipx/issues/1205.

## Test plan

Tested by running the following command on a windows machine with python3.12.exe in PATH (which errored before):

```
pipx run --python=python3.12.exe pycowsay moo
```


src/pipx/util.py
```diff
@@ -174,7 +174,7 @@ def run_subprocess(
     # See https://github.com/pypa/pipx/issues/1164
     # Conversely, if the binary is a symlink, then we should NOT use the real path, as Python expects to receive the
     # symlink in argv[0] so that it can locate the venv.
-    if not os.path.islink(cmd_str_list[0]) and WINDOWS:
+    if WINDOWS and os.path.isabs(cmd_str_list[0]) and not os.path.islink(cmd_str_list[0]):
         cmd_str_list[0] = os.path.realpath(cmd_str_list[0])
```
Gitznik, line 178: I still think we should raise if `os.path.realpath` cannot successfully create the "real path", as I don't think that would ever create a meaningful path.

src/pipx/util.py
```diff
@@ -174,7 +174,7 @@ def run_subprocess(
     # See https://github.com/pypa/pipx/issues/1164
     # Conversely, if the binary is a symlink, then we should NOT use the real path, as Python expects to receive the
     # symlink in argv[0] so that it can locate the venv.
```
Gitznik, line 176: Please update the comment according to the new logic

Gitznik (changes requested):
IMO we should not merge this before someone was able to replicate this still fixes the path redirection issue.

Also please fix the failing tests and add tests covering this new logic.

dechamps:
> IMO we should not merge this before someone was able to replicate this still fixes the path redirection issue.

I just tested 02f5fe26a8943c101f288a5db2e041db8aa19a17 myself and can confirm there is no regression - #1164 is still fixed on that commit.

Gitznik:
> > IMO we should not merge this before someone was able to replicate this still fixes the path redirection issue.
> 
> I just tested [02f5fe2](https://github.com/pypa/pipx/commit/02f5fe26a8943c101f288a5db2e041db8aa19a17) myself and can confirm there is no regression - #1164 is still fixed on that commit.

Thanks for verifying!

src/pipx/util.py
```diff
@@ -174,7 +174,7 @@ def run_subprocess(
...
-    if not os.path.islink(cmd_str_list[0]) and WINDOWS:
```
Gitznik, line 177: I'm also wondering if we could move this entire logic to `pipx/interpreter:find_python_interpreter`, as to me that seems like the more logical place. 

I've played around with this on my fork, feel free to take a look: https://github.com/Gitznik/pipx/tree/parse_python_to_path

gaborbernat (dismissed):
You need to make the CI pass before we do any more review.

guahki:
I dived into the source code and had a totally new idea how/where to fix the original issue: I think it is quite logical to resolve the root path when calculating the venv paths. Small caveat is, that the shared lib paths have to be recomputed after creating the lib, as the first computation might no correctly resolve the redirected paths (I think, due to the fact, that the class is instantiated at import time and not even the pipx folder might exist at this early time).

I added comments in the interesting places.

@Gitznik what do you think about this?

@dechamps I tested this with your issue #1164 and for the tests I did, it worked. But of course I invite you to heavily test this commit and share your findings here!

Gitznik:
I think this only works on a subset of the problem. In `pipx/venv.py` line 157, we would still use the provided python version, not the one that is resolved via `get_venv_paths`. Can you test if this still works if you explicitly pass a python version that is subject to path redirection via the `--python` argument? @guahki

guahki:
> Can you test if this still works if you explicitly pass a python version that is subject to path redirection via the `--python` argument?

I tried to test this and did not manage to find any issue. Using a MS Store Python (via PATH lookup or direct path) in ``--python`` did work for me.

I tried at least

* ``pipx run --python python3.12.exe pycowsay moo``
* ``pipx run --python C:\Users\<snip>\AppData\Local\Microsoft\WindowsApps\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\python.exe pycowsay moo``

> I think this only works on a subset of the problem. In `pipx/venv.py` line 157, we would still use the provided python version, not the one that is resolved via `get_venv_paths`.

I'm not 100% sure, how this whole path redirection magic works for MS Store Apps. But as far as I understand, the line you quote is no problem, as the python executable loaded here is aware of it's path redirection magic and handles that correctly by itself, independent of us (``pipx``) giving it a helping hand.

As far as I understand (and also the hypothesis by @dechamps expressed in https://github.com/pypa/pipx/issues/1164#issuecomment-1868340120), the issue arises only when calling a non-MS-Store executable living inside the redirected path by it's "unredirected" name. Then this executable thinks, it lives in (and searches for other files relative to) the unredirected path, when in reality it should use the redirected one. And this executable (due to not being part of a MS Store bundle) does not transparently redirect the paths on OS level.

guahki:
However, as discussed in https://github.com/pypa/pipx/issues/1203#issuecomment-1924697018, using different Python versions than the one used by pipx itself can not work at all with MS Store Python and the location of pipx's venv folder inside ``AppData\Local``

Gitznik:
> As far as I understand (and also the hypothesis by @dechamps expressed in [#1164 (comment)](https://github.com/pypa/pipx/issues/1164#issuecomment-1868340120)), the issue arises only when calling a non-MS-Store executable living inside the redirected path by it's "unredirected" name. Then this executable thinks, it lives in (and searches for other files relative to) the unredirected path, when in reality it should use the redirected one. And this executable (due to not being part of a MS Store bundle) does not transparently redirect the paths on OS level.

Thanks for double checking. In that case I think this is the preferable fix as it only resolves the path exactly where the issue occurs. 

> However, as discussed in [#1203 (comment)](https://github.com/pypa/pipx/issues/1203#issuecomment-1924697018), using different Python versions than the one used by pipx itself can not work at all with MS Store Python and the location of pipx's venv folder inside `AppData\Local`

So this only fixes passing the python version that is already used anyways by default as it's the one used by pipx?

guahki:
> So this only fixes passing the python version that is already used anyways by default as it's the one used by pipx?

Yes and no.

Yes: in the "new" directory placement (version 1.3.0 and newer) it is inherently impossible to use another python version. I think I will open another issue about this. Nevertheless, I could give this version as full path as well as just ``--python python3.12.exe`` which does a PATH lookup.

And no: when a legacy (pre 1.3.0) directory is present, I could use any of my installed python versions (MS Store Python as well as non-MS Store one) by lookup in the PATH as well as giving the full path. The only thing I could not test quickly, was using the py-launcher (as I do not use it on this system), but the change here should obviously not create any issues with it (and your new ``pipx/interpreter:find_python_interpreter`` handled this quite nice).

Gitznik:
In that case I think this is a good fix. Please fix the failing test and then mark ready for review :+1: I think the thing tested there is no longer relevant because you removed the `realpath` logic from the `run_subprocess` function.

Also please add a test to `get_venv_path` that verifies the results. I don't think we can actually fully reproduce the original error there as it has too many assumptions on how python is installed on the runner, but maybe you can just create a file in a directory that is subject to path redirection and check if it resolves correctly?

dechamps:
> maybe you can just create a file in a directory that is subject to path redirection and check if it resolves correctly?

I am not aware of any way to manually set up this kind of "path redirection" mechanism on Windows outside of an MS Store app. Back when I wrote #1168 I tried to write a regression test in the form of a GitHub Actions workflow that installed Python from the MS Store, but I hit a wall in the form of Cyberboss/install-winget#1. It would sure be nice to catch MS Store Python regressions automatically, but it's not clear how that can be done.

Gitznik:
> I am not aware of any way to manually set up this kind of "path redirection" mechanism on Windows outside of an MS Store app.

So the linked example from the python docs only applies if executed via a MS store python version?

dechamps:
As far as I'm aware, yes.

guahki:
> please fix the failing CI and then this is ready for review

I fixed the CI by just deleting the two test functions, testing the path resolution in the ``run_subprocess`` function. As discussed, I do not know, what (or better how) to test these changes in a better way.

Thanks for all your help, moderation and understanding @Gitznik!

Gitznik (approved):
Looks good to me. Thanks for all the investigation @guahki

PR #1236, opened 2024-01-29
Add --pre pip flag passthrough

TheTechromancer:
As a daily user of pipx, I use the `--pre` flag enough I thought it was worthy to add to the passthroughs. Using `--pip-args` always trips me up because of the need to specify `'\--backslashes'` 🙃.

Thanks!

### DIFF
--- a/changelog.d/1236.feature.md
+++ b/changelog.d/1236.feature.md
@@ -1,0 +1,1 @@
+- Add --pre flag
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -148,6 +148,9 @@ def get_pip_args(parsed_args: Dict[str, str]) -> List[str]:
     if parsed_args.get("index_url"):
         pip_args += ["--index-url", parsed_args["index_url"]]
 
+    if parsed_args.get("pre"):
+        pip_args += ["--pre"]
+
     if parsed_args.get("pip_args"):
         pip_args += shlex.split(parsed_args.get("pip_args", ""), posix=not WINDOWS)
 
@@ -320,6 +323,7 @@ def add_pip_venv_args(parser: argparse.ArgumentParser) -> None:
         help="Give the virtual environment access to the system site-packages dir.",
     )
     parser.add_argument("--index-url", "-i", help="Base URL of Python Package Index")
+    parser.add_argument("--pre", help="Include pre-release and development versions", action="store_true",)
     parser.add_argument(
         "--editable",
         "-e",
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -323,7 +323,11 @@ def add_pip_venv_args(parser: argparse.ArgumentParser) -> None:
         help="Give the virtual environment access to the system site-packages dir.",
     )
     parser.add_argument("--index-url", "-i", help="Base URL of Python Package Index")
-    parser.add_argument("--pre", help="Include pre-release and development versions", action="store_true",)
+    parser.add_argument(
+        "--pre",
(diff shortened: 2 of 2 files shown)


chrysle:
Thanks! Could you add a test, please?

chrysle:
Also, please take a look at the contribution guidelines – We'll need some changelog.

TheTechromancer:
Changelog entry added. @chrysle do you know if there's a way in `package_info.py` to add a package that has two version, one normal and one prerelease? This is the best way I can think of to test this option.

chrysle:
> do you know if there's a way in `package_info.py` to add a package that has two versions, one normal and one prerelease?

I don't think that's intended, as we always make use of the one specified version in the tests.
Basically, if a package isn't pinned, but does have an upper cap, `pip` will select the latest beta version it can grab with `--pre`, correct?
Then you could add a parameter to this test:

https://github.com/pypa/pipx/blob/10cbc9c83d79f4bc1e5dfbd03ada9233ee533f24/tests/test_package_specifier.py#L264

gaborbernat:
FWIW, I'm -1 on this. Seems few enough users that the `pip-args` should suffice. I'd rather keep the interface not more complicated than needs to be.

TheTechromancer:
From my perspective, `pipx install --pip-args '\--pre'` --> `pipx install --pre` is cleaner and less complex.

But come to think of it, I do agree that making individual exceptions for each of these flags
seems a little silly. Pipx's purpose is to be a pip wrapper. Is there a reason why we don't simply pass through all the non-pipx arguments to pip? This would let us keep our clean interface while saving users from the dash-escaping weirdness.

gaborbernat:
I'm minus 1 on passing through unrecognized flags, It means that we will never be sure if they use their mistyped something or that's just a pip flag, that's IMHO bad UX.

TheTechromancer:
> passing through unrecognized flags

Why not just import pip's own argument parser? I'll explore this a bit when I get time.

chrysle:
> Why not just import pip's own argument parser? I'll explore this a bit when I get time.

However keep in mind it's discouraged to use `pip` as an API. Also, `pipx` has quite a few own options. This is where the parsing takes place in `pip`:

https://github.com/pypa/pip/blob/99a82b7e49a1657d18c8453c42f7eff971ce1cb5/src/pip/_internal/cli/main_parser.py#L69

TheTechromancer:
Hmm yeah it's possible that solution would be too complex and not worth the trouble. In the grand scheme of things, the `--pip-args` disconnect is a small issue. Pipx is already a great tool. But I think it could benefit from a more seamless interaction between pipx and pip.

Would a PR like this be considered?

gaborbernat:
> But I think it could benefit from a more seamless interaction between pipx and pip.

pipx doesn't offer a public API, so achieving that wouldn't be easy. The existing `--pip-args` IMHO is the best we can get. 🤔 And I'm a bit fearful of supporting every pip flag explicitly. IMHO, this downside:

> Using `--pip-args` always trips me up because of the need to specify `'\--backslashes'` 🙃.

It is acceptable considering how rarely people need to use it. We need to consider here the impact of such a change will have not just on the end users, but also how much problem it would generate for maintainers.

TheTechromancer:
@gaborbernat all good points. If there isn't an elegant solution codewise, I agree it's not for the best.

gaborbernat:
I guess we concluded that this is a won't do for now. But thanks for the contribution, hope even though the outcome is not what you wanted, you agree with the solution.

Issue #1242, opened 2024-02-04
[feature] Download/use standalone python build when chose `--python` version doesn't exist


alextremblay:
**How would this feature be useful?**

It would be awesome if we could run `pipx run --python=python3.10 cowsay hello` and have it *just work*, even if we didn't have `python3.10` installed locally

It would be awesome if we could share scripts with colleagues that start with `#!/usr/bin/env -S pipx run --python=python3.11` and have it *just work* even if said colleagues don't have python 3.11 installed

**Describe the solution you'd like**

my idea would be to modify `pipx.interpreter.find_python_interpreter` so that if it's not able to find a python binary matching what the user requested with the `--python` flag, rather than failing with an exception, the funtion would first attempt to download and unpack a python standalone build to a local cache directory (`PIPX_HOME/py` for example) and use that

**Describe alternatives you've considered**

AFAIK there is not currently a workaround / way to do this in pipx, other than to instruct all users of saind scripts / tools to manually install the correct version of python in one of the myriad ways that python can be installed on a given system

PR #1243, opened 2024-02-04
[feature] Download/use standalone python build when chose --python version doesn't exist

alextremblay:

## Summary of changes

As per #1242, this PR updates `pipx.interpreter.find_python_interpreter` to download the desired python version if `pipx` is not able to find it in the system, rather than simply failing with an exception.

## Test plan

Tested by running

```
# The following commands were successfully tested on a system which only has python 3.8 and 3.10 installed
$ pipx run --python=python3.9 cowsay -t hello
$ pipx run --python=python3.10 cowsay -t hello
$ pipx run --python=python3.11 cowsay -t hello
$ pipx run --python=python3.12 cowsay -t hello
$ pipx run --python=3.9 cowsay -t hello
$ pipx run --python=3.10 cowsay -t hello
$ pipx run --python=3.11 cowsay -t hello
$ pipx run --python=3.12 cowsay -t hello
```

The following was also tested:

`myscript.py`:
```python
#!/usr/bin/env -S pipx run --python=3.11 --path

# /// script
# dependencies = [
#   "requests",  # example of external dependency
# ]
# ///
import requests

print(requests.get("https://example.com").text)
```
```console
$ chmod +x myscript.py
$ ./myscript.py | head
<!doctype html>
<html>
<head>
    <title>Example Domain</title>

    <meta charset="utf-8" />
    <meta http-equiv="Content-type" content="text/html; charset=utf-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1" />
    <style type="text/css">
    body {
```


alextremblay:
@dukecat0 @uranusjr I've got a preliminary POC of this feature which I'm quite happy with, but before I sink any more time into this, I'm hoping to get some feedback and make sure I'm on the right track.

Is this something you would like me to keep working on? do you have any notes on my implementation?

dukecat0:
Thanks for working on this! I personally find this feature very useful. 👍 Probably it will be better if there's an environment variable that allows users to disable this behavior.

chrysle:
Looks like a helpful feature, thanks!

> Probably it will be better if there's an environment variable that allows users to disable this behavior.

I agree with that – or even a flag to toggle the behaviour.

Gitznik:
Good stuff! 

I would argue that this feature should be opt-in, rather than opt-out. I would be very confused if I accidentally mistype a python version and pipx just starts downloading it unprompted.

chrysle:
> opt-in, rather than opt-out

I agree – that's why I suggest a flag (with negative default value).

alextremblay:
Excellent idea! I’m more than happy to implement that. Not sure what to call the flag though… something succinct?

ideally also an environment variable to flip the default

maybe something like `—fetch-as-needed`? It’s a bit wordy, but I can’t think of anything more succinct that gets the idea across

Gitznik:
> ideally also an environment variable to flip the default
+1
> maybe something like `—fetch-as-needed`? It’s a bit wordy, but I can’t think of anything more succinct that gets the idea across
`fetch-missing-python`? `fetch-as-needed` could also refer to the package to install/run itself.

src/pipx/constants.py
```diff
@@ -32,6 +34,7 @@
 PIPX_SHARED_PTH = "pipx_shared.pth"
 LOCAL_BIN_DIR = Path(os.environ.get("PIPX_BIN_DIR", DEFAULT_PIPX_BIN_DIR)).resolve()
 LOCAL_MAN_DIR = Path(os.environ.get("PIPX_MAN_DIR", DEFAULT_PIPX_MAN_DIR)).resolve()
+FETCH_MISSING_PYTHON = os.environ.get("PIPX_FETCH_MISSING_PYTHON", "") == ""
```
chrysle, line 37:
```suggestion
FETCH_MISSING_PYTHON = os.environ.get("PIPX_FETCH_MISSING_PYTHON", False)
```

src/pipx/standalone_python.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+def download_python_build_standalone(python_version: str):
+    """When all other python executable resolutions have failed,
+    attempt to download and use an appropriate python build
+    from https://github.com/indygreg/python-build-standalone
+    and unpack it into the pipx shared directory."""
+
+    # python_version can be a bare version number like "3.9" or a "binary name" like python3.10
+    # let's coerce it to a bare version number
```
chrysle, line 53:
```suggestion
    # Split it to a bare version specifier
```

Didn't even know that word, `coerce`. Americans!

src/pipx/standalone_python.py
```diff
@@ -0,0 +1,182 @@
...
+        expected_checksum = urlopen(checksum_link).read().decode().rstrip("\n")
+        if checksum != expected_checksum:
+            raise PipxError(
+                f"Checksum mismatch for python {full_version} build. " f"Expected {expected_checksum}, got {checksum}."
+            )
+
+        with tarfile.open(archive_path, mode="r:gz") as tar:
+            tar.extractall(download_dir)
+
+        (download_dir / "python").rename(install_dir)
```
chrysle, line 99: I don't exactly understand what this does?

src/pipx/standalone_python.py
```diff
@@ -0,0 +1,182 @@
...
+
+def get_latest_python_releases() -> List[str]:
+    """Returns the list of python download links from the latest github release."""
+    try:
+        with urlopen(GITHUB_API_URL) as response:
+            release_data = json.load(response)
+
+    except urllib.error.URLError as e:
+        # raise
+        raise PipxError("Unable to fetch python-build-standalone release data.") from e
```
chrysle, line 133:
```suggestion
        raise PipxError(f"Unable to fetch python-build-standalone release data (from {GITHUB_API_URL}).") from e
```

src/pipx/standalone_python.py
```diff
@@ -0,0 +1,182 @@
...
+
+def resolve_python_version(requested_version: str):
+    pythons = list_pythons()
+
+    for version, version_download_link in pythons.items():
+        if version.startswith(requested_version):
+            python_version = version
+            download_link = version_download_link
+            break
+    else:
```
chrysle, line 179: This seems to be indented wrongly?

src/pipx/interpreter.py
```diff
@@ -58,6 +59,14 @@ def find_python_interpreter(python_version: str) -> str:
 
     if shutil.which(python_version):
         return python_version
+
+    if fetch_missing_python or FETCH_MISSING_PYTHON:
+        try:
+            standalone_executable = download_python_build_standalone(python_version)
+            return standalone_executable
+        except PipxError as e:
+            raise InterpreterResolutionError(source="PATH", version=python_version) from e
```
chrysle, line 68: I think we should add another source for `InterpreterResolutionError`?

chrysle:
Please add a news fragment, too. And we'll need tests.

src/pipx/standalone_python.py
```diff
@@ -0,0 +1,182 @@
...
+        if checksum != expected_checksum:
+            raise PipxError(
+                f"Checksum mismatch for python {full_version} build. " f"Expected {expected_checksum}, got {checksum}."
+            )
+
+        with tarfile.open(archive_path, mode="r:gz") as tar:
+            tar.extractall(download_dir)
+
+        (download_dir / "python").rename(install_dir)
+        shutil.rmtree(download_dir, ignore_errors=True)
```
Gitznik, line 100: Should this be in a `finally` block to make sure this function always cleans up after itself?

Gitznik (changes requested):

src/pipx/constants.py
alextremblay: yes, that is simply. ty!

src/pipx/standalone_python.py
alextremblay: hahaha I'm Canadian, but i get your meaning :) 

I've re-worded it

src/pipx/standalone_python.py
alextremblay: the contents of the extracted tarball contain a folder called python which contains the `lib` and `bin` etc directories we want, so we specifically move that subfolder into place rather than the base `download_dir` folder

i've modified that function and added code comments to clarify

src/pipx/standalone_python.py
alextremblay: good idea, I've added this suggestion

src/pipx/standalone_python.py
alextremblay: That is a for-else loop. if the for loop is exhausted without breaking (without finding a valid python version), the else clause is executed and raises an exception.

src/pipx/interpreter.py
alextremblay: i suppose so... what should we call it?

src/pipx/standalone_python.py
```diff
@@ -0,0 +1,182 @@
...
+    and unpack it into the pipx shared directory."""
+
+    # python_version can be a bare version number like "3.9" or a "binary name" like python3.10
+    # let's coerce it to a bare version number
+    python_version = re.sub(r"[c]?python", "", python_version)
+
+    install_dir = PIPX_STANDALONE_PYTHON_CACHEDIR / python_version
+
+    if install_dir.exists():
+        return str(install_dir / "bin/python3")
```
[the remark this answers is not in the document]
alextremblay: hmmm... good eye, fixed!

src/pipx/standalone_python.py
alextremblay: good idea.

I started wrapping the logic in a try-finally block, and realized that the code had gotten a bit long in the tooth, so I've gone ahead and split it up to make it clearer.

I'll start writing tests now. 

Not sure if i can write a windows test for this, as i don't have a windows machine to test with, but I'll give it a shot

src/pipx/interpreter.py
alextremblay: ok, i got it. it looks a bit verbose in code, but it should come out clear to the end user, and that's arguably more important.

check out [f241014](https://github.com/pypa/pipx/pull/1243/commits/f2410140622645c773b33b416d8e38eda6017e08) and tell me what you think

src/pipx/standalone_python.py
```diff
@@ -0,0 +1,197 @@
...
+    download_link_suffix = MACHINE_SUFFIX[system][machine]
+    # linux suffixes are nested under glibc or musl builds
+    if system == "Linux":
+        # fallback to musl if libc version is not found
+        libc_version = platform.libc_ver()[0] or "musl"
+        download_link_suffix = download_link_suffix[libc_version]
+
+    python_releases = get_or_update_index()["releases"]
+
+    available_python_links = [link for link in python_releases if link.endswith(download_link_suffix)]
```
chrysle, line 165: Let's ensure these aren't duplicated?

```suggestion
    available_python_links = (link for link in python_releases if link.endswith(download_link_suffix))
```

changelog.d/1242.feature.md
```diff
@@ -0,0 +1 @@
+Add a `--fetch-missing-python` flag to all commands that accept a `--python` flag. When combined, will automatically download a standalone copy of the requested python version if it's not already available on the user's system
```
chrysle, line 1: Let's use [semantic linefeeds](https://rhodesmill.org/brandon/2012/one-sentence-per-line/).

```suggestion
Add a `--fetch-missing-python` flag to all commands that accept a `--python` flag. 

When combined, this will automatically download a standalone copy of the requested python version if it's not already available on the user's system.
```

src/pipx/main.py
```diff
@@ -528,6 +549,13 @@ def _add_reinstall(subparsers, venv_completer: VenvCompleter, shared_parser: arg
             f"Requires Python {MINIMUM_PYTHON_VERSION} or above."
         ),
     )
+    p.add_argument(
```
chrysle, line 552: Let's create a helper function in the style of `add_include_dependencies`, instead of duplicating?

src/pipx/standalone_python.py
```diff
@@ -0,0 +1,197 @@
...
+            tar.extractall(download_dir)
+
+
+def get_or_update_index():
+    """Get or update the index of available python builds from
+    the python-build-standalone repository."""
+    index_file = PIPX_STANDALONE_PYTHON_CACHEDIR / "index.json"
+    if index_file.exists():
+        index = json.loads(index_file.read_text())
+        # update index if it's been more than a month
```
chrysle, line 126:
```suggestion
        # update index after 30 days
```

Every month has 30 days?

tests/test_interpreter.py
```diff
@@ -167,3 +186,17 @@ def test_find_python_interpreter_missing_on_path_raises(monkeypatch):
...
         assert "on your PATH" in str(e)
+
+
+@pytest.mark.skipif(sys.platform.startswith("win"), reason="Looks for python3")
+def test_fetch_missing_python(monkeypatch):
+    major = sys.version_info.major
+
+    # For this test, we want to ask pipx for a version that doesn't exist locally.
+    # Since we test on all supported versions, we can't just pick a specific version to test with.
+    minor = 11 if sys.version_info.minor == 10 else 10
```
chrysle, line 197: I'm not quite happy with that. Isn't there some other way to ensure a Python version isn't available?

tests/test_interpreter.py
```diff
@@ -167,3 +186,17 @@ def test_find_python_interpreter_missing_on_path_raises(monkeypatch):
         find_python_interpreter(interpreter)
         assert "Python Launcher" in str(e)
         assert "on your PATH" in str(e)
+
+
+@pytest.mark.skipif(sys.platform.startswith("win"), reason="Looks for python3")
+def test_fetch_missing_python(monkeypatch):
+    major = sys.version_info.major
```
chrysle, line 193: This will always be _3_?

src/pipx/interpreter.py
chrysle: Looks good! Resolving.

src/pipx/standalone_python.py
alextremblay: I don't believe turning the list comprehension into a generator comprehension buys us anything.

Assuming you intend to turn it into a set comprehension, that would work, but it would also be unnecessary, as the list is immediately turned into a dict, keyed on version number (see implementation [here](https://github.com/pypa/pipx/pull/1243/files#diff-0bebcbd675eca63cf1777a9bed07c3317f597837e80061206f6829e7e0f9a921R167-R172))

Here is the return value of `list_python`:
```python
{'3.10.13': 'https://github.com/indygreg/python-build-standalone/releases/download/20240107/cpython-3.10.13%2B20240107-x86_64_v3-unknown-linux-gnu-install_only.tar.gz',
 '3.11.7': 'https://github.com/indygreg/python-build-standalone/releases/download/20240107/cpython-3.11.7%2B20240107-x86_64_v3-unknown-linux-gnu-install_only.tar.gz',
 '3.12.1': 'https://github.com/indygreg/python-build-standalone/releases/download/20240107/cpython-3.12.1%2B20240107-x86_64_v3-unknown-linux-gnu-install_only.tar.gz',
 '3.9.18': 'https://github.com/indygreg/python-build-standalone/releases/download/20240107/cpython-3.9.18%2B20240107-x86_64_v3-unknown-linux-gnu-install_only.tar.gz'}
```

src/pipx/main.py
alextremblay: not sure what that would look like... could you provide an example implementation?

tests/test_interpreter.py
alextremblay: as far as I'm aware, yes

tests/test_interpreter.py
alextremblay: good point, it's been fixed. 

Please have a look and tell me what you think

src/pipx/main.py
chrysle: I basically suggest adding this option where needed like in this line:

https://github.com/pypa/pipx/blob/e28c7b1033eded0242d7b5d37156a8e8f94a49ba/src/pipx/main.py#L348

Only with a different underlying function.

alextremblay:
Hi All, this feature is now, to the best of my ability, complete.

I've added tests, a news entry, and adopted / implemented just about every change that's been requested.

The tests are all now passing on all platforms and python versions.

I've spent a not insignificant amount of time and energy making this feature implementation as good as it can be in the hopes that it will be accepted and merged. I am running out of time and energy to devote to this now. 

I think it's ready to review and/or merge. I'm going to step away from it now. If there are any additional changes you would like to make, feel free to make them before or after merging the feature.

Thank you

tests/test_interpreter.py
```diff
@@ -167,3 +183,34 @@ def test_find_python_interpreter_missing_on_path_raises(monkeypatch):
...
         assert "Python Launcher" in str(e)
         assert "on your PATH" in str(e)
+
+
+@pytest.mark.skipif(sys.platform.startswith("win"), reason="Looks for python3")
+def test_fetch_missing_python(monkeypatch):
+    major, minor = _derive_target_python_version()
+
+    python_path = find_python_interpreter(f"{major}.{minor}", fetch_missing_python=True)
+    assert python_path is not None
```
chrysle, line 193: I'd probably ensure the Python really isn't found (e. g. for users who simply run `pytest`):

```suggestion
   python_version = f"{major}.{minor}"
   
   with pytest.raises(InterpreterResolutionError):
       find_python_interpreter(python_version)

    python_path = find_python_interpreter(python_version, fetch_missing_python=True)
    assert python_path is not None
```

Let's add a variable `python_version` as demonstrated here, BTW, to simplify.

src/pipx/standalone_python.py
```diff
@@ -0,0 +1,197 @@
...
+        download_link_suffix = download_link_suffix[libc_version]
+
+    python_releases = get_or_update_index()["releases"]
+
+    available_python_links = [link for link in python_releases if link.endswith(download_link_suffix)]
+
+    python_versions: dict[str, str] = {}
+    for link in available_python_links:
+        match = PYTHON_VERSION_REGEX.search(link)
+        assert match is not None
```
Gitznik, line 170: I don't think we should assert here, but maybe rather log a warning and skip the link?

tests/test_interpreter.py
```diff
@@ -167,3 +183,34 @@ def test_find_python_interpreter_missing_on_path_raises(monkeypatch):
...
+    # unless the current minor version is the minimum supported version, in which case the requested
+    # minor version will be one more than the current minor version.
+    # ie if the test is running on python 3.9, the requested version will be 3.8
+    # if the test is running on python 3.8, the requested version will be 3.9
+    if minor <= minimum_minor_version:
+        minor = minimum_minor_version + 1
+    else:
+        minor -= 1
+
+    return major, minor
```
Gitznik, line 220: Does this work in local tests were any number of python versions might already be available?

Gitznik (changes requested):
I can work on my change requests myself.

tests/test_interpreter.py
alextremblay: hmmmm... perhaps if you mock-out `shutil.which` to return false either way?

tests/test_interpreter.py
chrysle: Indenting with GH code suggestions is a pain.... This wasn't really meant to be committed as-is, sorry. The other test should also be amended, and the variable can be used once more.

alextremblay: gotcha. I'll revert and commit

Gitznik:
Would be nice if there was a command to clean up the python versions, e.g. when a downloaded python version does not have a installed app that uses it anymore. We can treat that as a separate feature request though IMO.

alextremblay:
> Would be nice if there was a command to clean up the python versions, e.g. when a downloaded python version does not have a installed app that uses it anymore. We can treat that as a separate feature request though IMO.

Also a command to update one or all standalone versions (ie someone uses `--python=3.9 --fetch-missing-python` and wind up with v3.9.10 installed to `PIPX_HOME/py/3.9`. several months later, they come back and install another package with `--python=3.9 --fetch-missing-python`, and that package gets installed with python 3.9.10, even though v3.9.18 is now available)

That's a hard problem to solve automatically, so i think a manual process / command would be the way to go

alextremblay:
This latest error in the windows unit test is... interesting, to say the least.

is `PIPX_HOME` cached in CI between test runs?

I'm not quite sure how else to explain why `PIPX_HOME/py/3.11` already exists on the windows CI server...

chrysle:
I suppose that's [502d0cc](https://github.com/pypa/pipx/pull/1243/commits/502d0cc9978326668ca28190e5aba043e9e3cf86)?

chrysle:
Probably we should use pytest's `tmp_path` fixture?

alextremblay:
> Probably we should use pytest's `tmp_path` fixture?

to override the `PIPX_STANDALONE_PYTHON_CACHEDIR` constant?

I suppose so, but that somehow feels wrong. the unit test should as closely as possible reflect what will actually happen on a user's machine

ordinarily, this error we're seeing should be impossible. if standalone python 3.11's already been downloaded, the function should return early without trying to redownload it. this error seems to indicate that a previous run failed and produced an empy `PIPX_HOME/py/3.11` folder...

src/pipx/main.py
alextremblay: done!

alextremblay:
Well, apart from that rate limit error on the windows test, I think we're now good to go.

Are there any additional changes you need, or can this be merged?

src/pipx/interpreter.py
```diff
@@ -58,6 +61,14 @@ def find_python_interpreter(python_version: str) -> str:
 
     if shutil.which(python_version):
         return python_version
+
+    if fetch_missing_python or FETCH_MISSING_PYTHON:
+        try:
+            standalone_executable = download_python_build_standalone(python_version)
+            return standalone_executable
+        except PipxError as e:
+            raise InterpreterResolutionError(source="the python-build-standalone project", version=python_version) from e
```
Gitznik, line 70:
```suggestion
        except PipxError as e:
            logger.error(e)
            raise InterpreterResolutionError(source="the python-build-standalone project", version=python_version) from e
```

Should probably log the original error here. You've created nice errors with a lot of detail that would otherwise just not be shown, as the InterpreterResolutionError is caught and just it's content is printed.

tests/test_interpreter.py
```diff
@@ -31,6 +32,19 @@ def which(name):
     assert python_path.endswith("python.exe")
 
 
+@pytest.mark.skipif(not sys.platform.startswith("win"), reason="Looks for Python.exe")
+def test_windows_python_fetch_missing(monkeypatch):
+    def which(name):
+        return None
+
+    major, minor = _derive_target_python_version()
```
Gitznik, line 40: When patching `which`, is this still necessary? Same for `test_fetch_missing_python`. Would be nice to get rid of the entire `_derive_target_python_version` function if we can.

Gitznik (changes requested):

chrysle:
I'd like to raise awareness for https://github.com/pypa/pipx/pull/1243#discussion_r1478427702 that should still be implemented IMHO, as it tests basic functionality.

As for the rate limit, I see that as a possible issue, as these tests will be run many times. I propose skipping this test if `standalone_python.py` isn't modified (don't know how to achieve that, yet), or scheduling it.

tests/test_interpreter.py
alextremblay: that function de-duplicates the logic for selecting a valid target python version which is not already installed.

i thought patching `which` was necessary in the windows test to prevent pipx from detecting `py` and trying to use `py` to resolve the target version, but i could be wrong. As i've mentioned previously, i don't have a windows machine to test / debug with, so all my knowledge of the windows testing architecture is derived from just reading the source code

src/pipx/interpreter.py
alextremblay: The benefit of raising an exception *from* another exception (ie. `raise InterpreterResolutionError(...) from e`) is that the stack trace includes the surface error *and* the underlying error.

I think logging the error on top of that would be excessive

alextremblay:
> I'd like to raise awareness for [#1243 (comment)](https://github.com/pypa/pipx/pull/1243#discussion_r1478427702) that should still be implemented IMHO, as it tests basic functionality.
> 
> As for the rate limit, I see that as a possible issue, as these tests will be run many times. I propose skipping this test if `standalone_python.py` isn't modified (don't know how to achieve that, yet), or scheduling it.

The first suggestion from that comment (ensuring the target python isn't found on dev's machines) has already been implemented (by mocking out `which` in both the windows test and the unix test.

The second comment, adding a `python_version` variable to these two unit tests, doesn't make sense to me. none of the other tests in that test suite define such a variable, and i don't understand why we would want to add such an inconsistency or why the lack of this variable is a blocker towards getting this change merged?

alextremblay:
> I'd like to raise awareness for [#1243 (comment)](https://github.com/pypa/pipx/pull/1243#discussion_r1478427702) that should still be implemented IMHO, as it tests basic functionality.
> 
> As for the rate limit, I see that as a possible issue, as these tests will be run many times. I propose skipping this test if `standalone_python.py` isn't modified (don't know how to achieve that, yet), or scheduling it.

regarding the rate limit, that is definitely a concern. I'm going to try and write a fixture to mock out the API call specifically, that should take care of that problem

stand by

src/pipx/interpreter.py
Gitznik: but the PipxError is caught in `pipx/main`, line 191, so the stack trace will never be printed?

```
(develop-3-12) ➜  pipx git:(resolve_home_path_in_dir_env_vars) ✗ pipx install --python python99.99 -vv
pipx >(setup:858): 2024-02-06 16:43:45
pipx >(setup:859): /home/robert/private/code/pipx/.nox/develop-3-12/bin/pipx install --python python99.99 -vv
pipx >(setup:860): pipx version is 1.4.4.dev16+g3e423c2.d20240203
pipx >(setup:861): Default python interpreter is '/home/robert/private/code/pipx/.nox/develop-3-12/bin/python'
⚠️ No executable for the provided Python version 'python99.99' found in PATH. Please make sure the provided version is on your PATH or the file path is valid.
pipx >(cli:932): pipx finished.
```

tests/test_interpreter.py
Gitznik:
```
def find_python_interpreter(python_version: str, fetch_missing_python: bool = False) -> str:
    if Path(python_version).is_file():
        return python_version

    try:
        py_executable = find_py_launcher_python(python_version)
        if py_executable:
            return py_executable
    except (subprocess.CalledProcessError, FileNotFoundError) as e:
        raise InterpreterResolutionError(source="py launcher", version=python_version) from e

    if shutil.which(python_version):
        return python_version

    if fetch_missing_python or FETCH_MISSING_PYTHON:
        try:
            standalone_executable = download_python_build_standalone(python_version)
            return standalone_executable
        except PipxError as e:
            raise InterpreterResolutionError(source="the python-build-standalone project", version=python_version) from e
```
When patching which:
1. it's not a path
2. find_py_launcher_python won't find an executable because py was not found
3. no other python version will be found because `if shutil.which(python_version):` will be false
4. so it will go to downloading

I'm not hellbent on changing the test, but I think the way `find_python_interpreter` is written right now, `_derive_target_python_version` is no longer needed

chrysle:
> The first suggestion from that comment (ensuring the target python isn't found on dev's machines) has already been implemented (by mocking out `which` in both the windows test and the unix test.

However, implementing it would ensure the exception is raised correctly, which isn't entirely useless.

src/pipx/interpreter.py
alextremblay: ahhhhh, i see.

if `run_pipx_command` is swallowing exception details, wouldn't it be better to have `run_pipx_command` log those exception details rather than to add special handling to all functions which raise exception chains?

ie. add `log.error(e, exc_info=True)` to the exception-handling code in `run_pipx_command`

tests/test_interpreter.py
alextremblay: ahhhh, i think i understand. hang on, let me refactor this a bit

src/pipx/interpreter.py
Gitznik: Fair enough. I'd set it to `logger.info` though, as we would only want to log the traceback if the command is executed with verbose. The normal print should be enough information for normal use.

tests/test_interpreter.py
alextremblay: done

alextremblay:
> I'd like to raise awareness for [#1243 (comment)](https://github.com/pypa/pipx/pull/1243#discussion_r1478427702) that should still be implemented IMHO, as it tests basic functionality.

hi @chrysle, I've implemented test cases for versions of python which aren't available in the python-build-standalone project

in the future, we might want to expand the logic of the `standalone_python` module to collect data from multiple releases of the python-build-standalone project and merge it all together so that we can support fetching more python versions than are available in the current release (ie. `20231002` was the last release to include a version of python 3.8), but that's a massive increase in complexity for very little gain, and far outside the scope of this PR

tests/test_interpreter.py
Gitznik: Great stuff, thank you!

src/pipx/interpreter.py
alextremblay: do you want me to do this as part of the PR?

I would prefer it be done by one of the maintainers as a separate thing... it feels like something that belongs outside the scope of this PR, but if you disagree, i'm willing to write and commit the change

Gitznik: Sure, I can do it in a separate PR. I don't see any harm in doing it here, but it's up to you.

Gitznik (dismissed):

tests/test_interpreter.py
```diff
@@ -15,6 +18,17 @@
...
 
+@pytest.fixture()
+def mocked_github_api(monkeypatch, root):
+    """
+    Fixture to replace the github index with a local copy,
+    to prevent unit tests from exceeding github's API request limit.
+    """
+    with open(root / "testdata" / "standalone_python_index.json") as f:
+        index = json.load(f)
+    monkeypatch.setattr(pipx.standalone_python, "get_or_update_index", lambda: index)
```
chrysle, line 29: Actually, patching the `index_file` variable here would enable us dropping the `json` import?

```suggestion
    index_file = root / "testdata" / "standalone_python_index.json"
    monkeypatch.setattr(pipx.standalone_python, "get_or_update_index", lambda: index_file)
```

tests/test_interpreter.py
```diff
@@ -167,3 +205,27 @@ def test_find_python_interpreter_missing_on_path_raises(monkeypatch):
         find_python_interpreter(interpreter)
         assert "Python Launcher" in str(e)
         assert "on your PATH" in str(e)
+
+
+@pytest.mark.skipif(sys.platform.startswith("win"), reason="Looks for python3")
```
chrysle, line 210: I think we could deduplicate the tests and simply add a parametrisation for the executable names, then test for the platform within the test and present an according `assert` statement?

chrysle (commented):
Two suggestions, looks good to me other than that!

tests/test_interpreter.py
alextremblay: the return value of `get_or_update_index` is a dict, not a file

chrysle: Oh, I wasn't aware that one can't patch local variables!

chrysle (approved):

chrysle:
Thank you for all the work you have invested into this contribution!

> > Would be nice if there was a command to clean up the python versions, e.g. when a downloaded python version does not have a installed app that uses it anymore. We can treat that as a separate feature request though IMO.
> 
> Also a command to update one or all standalone versions

Additional concerns, as a `pipx clean` (?) command, shall be addressed in follow-ups.

PR #1240, opened 2024-02-03
Update requirement specification

oberbichler:
- [x] I have added a news fragment under `changelog.d/` (if the patch affects the end users)

## Summary of changes

- Update documentation for `pipx run` with dependencies in a local script (resolves #1227)

## Test plan

Tested by running

```python
# test.py

# /// script
# dependencies = ["requests"]
# ///

import sys
import requests
project = sys.argv[1]
pipx_data = requests.get(f"https://pypi.org/pypi/{project}/json").json()
print(pipx_data["info"]["version"])
```
with

```bash
pipx run test.py
```

PR #1244, opened 2024-02-05
docs: fix outdated "run with inline dependencies" example

katrinafyi:
## Summary of changes
The documentation was not updated since its first implementation in #916. Since then, #1100 and #1180 have changed the syntax in incompatible ways.

This is a really cool feature! It would be nice if a warning was given for using the old syntax, though I understand it was not available for a great deal of time.

## Test plan
Tested by running

```bash
cat <<EOF > test.py
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "requests<3",
#   "rich",
# ]
# ///

import requests
from rich.pretty import pprint

resp = requests.get("https://peps.python.org/api/peps.json")
data = resp.json()
pprint([(k, v["title"]) for k, v in data.items()][:10])
EOF

pipx run file:test.py
```
as in the updated docs/example.md

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -65,28 +65,38 @@ pipx run --spec test-py test.py # Always a package on PyPI
 
 You can also run scripts that have dependencies:
 
-If you have a script `test.py` that needs a 3rd party library like requests:
+If you have a script `test.py` that needs 3rd party libraries,
+you can add [inline script metadata](https://packaging.python.org/en/latest/specifications/inline-script-metadata/)
+in the style of PEP 723.
 
 ```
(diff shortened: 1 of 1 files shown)


Gitznik:
Hi @katrinafyi, thanks for contributing this :) I'm afraid there is already [another open PR](https://github.com/pypa/pipx/pull/1240) for updating this. Feel free to add your changes as suggestions there. If that PR goes stale, we can continue here.

katrinafyi:
Oops! I forgot to check the PRs. It is good you're aware, you can close this if you want.

PR #1246, opened 2024-02-06
Refactor release process in order to include changelog in release files

chrysle:

## Summary of changes

This ought to correctly include the changelog in the docs on release, after PR merge.

## Test plan

Tested by running

```
```


Gitznik:
I think this is an okay quick fix (potentially for the next release so we're not blocked by this?).

What I don't like about this is that the release artifacts in github releases would not contain the updated changelog. Not a huge deal I guess, but feels wrong.

chrysle:
> What I don't like about this is that the release artifacts in github releases would not contain the updated changelog. Not a huge deal I guess, but feels wrong.

I understand. What alternative do you suggest? My idea would be to transfer the release-related jobs from the `tests` workflow to a new `release` workflow with a `workflow_dispatch` event, then configuring the "Bump changelog on release" job to actually create the release, causing the other jobs to run. We would end up running that workflow manually to release.

I could rework the PR towards that, WDYT?

Gitznik:
> I understand. What alternative do you suggest? My idea would be to transfer the release-related jobs from the `tests` workflow to a new `release` workflow with a `workflow_dispatch` event, then configuring the "Bump changelog on release" job to actually create the release, causing the other jobs to run. We would end up running that workflow manually to release.
> 
> I could rework the PR towards that, WDYT?

Yes, that would be my suggestion as well. Should lead to the cleanest result overall. Just mentioning that @gaborbernat mentioned that this project tries to avoid this kind of workflow, but I don't see a good alternative.

gaborbernat:
Where would you specify the version to release in a workflow_dispatch?

gaborbernat:
Works for me 👍 remember to update https://github.com/pypa/pipx/blob/main/CONTRIBUTING.md#releasing-new-pipx-versions

chrysle:
There is one complication which requires a bit more discussion. If we want to stick with opening a pull request every release to update the changelog, how can the same workflow be configured to create a GH release _after_ that pull request is merged?

Gitznik:
> There is one complication which requires a bit more discussion. If we want to stick with opening a pull request every release to update the changelog, how can the same workflow be configured to create a GH release _after_ that pull request is merged?

I think [release-please](https://github.com/googleapis/release-please) does it by giving the changelog PR a specific tag and watching for that tag for creating the release.

Gitznik:
Will you work on this soon @chrysle? If not I'd suggest we create a release soon with the current setup. There's a pretty good amount of bug fixes and new features to be released now. 

If we do one with the current setup, I can manually create a PR updating the docs, so they show up correctly in the docs after the release?

chrysle:
Updated. See the result in https://github.com/chrysle/pipx/pull/2/commits/fff344872a6498f1b4561e380770de13802414dc.

Gitznik:
While your current solution works, what about giving this changelog PR a specific label and run the release when a PR with that label is merged? 

Label is a direct part of the `pull_request` object when using the `pull_request` trigger, and the version to release we can extract from the PR title, which is also in the `pull_request` object. 

Removes one manual step from the pipeline.

chrysle:
I'll look into it, it could take a little bit of time, though.

chrysle:
This isn't quite that easy, as it turns out.. After trial and error for a while I finally managed to get my workflow to trigger on   
a merged pull request with a specific label (see the [successful run](https://github.com/chrysle/pipx/actions/runs/8679732477)), but just as it creates the release, the `tests` workflow is already running on the `push` event, and the pushing of the release tag is somehow [swallowed](https://github.com/chrysle/pipx/actions/runs/8679732488), because of which the CD isn't triggered. Any suggestions?

chrysle:
@henryiii May I ask for some advice? The current state is at https://github.com/chrysle/pipx/tree/test-cd.

chrysle:
@henryiii Pinging you on this as requested.

Gitznik:
> This isn't quite that easy, as it turns out.. After trial and error for a while I finally managed to get my workflow to trigger on a merged pull request with a specific label (see the [successful run](https://github.com/chrysle/pipx/actions/runs/8679732477)), but just as it creates the release, the `tests` workflow is already running on the `push` event, and the pushing of the release tag is somehow [swallowed](https://github.com/chrysle/pipx/actions/runs/8679732488), because of which the CD isn't triggered. Any suggestions?

Just seeing this now. That sounds quite tricky indeed. As far as I'm concerned, unless someone that is really into github workflows wants to pick this up, I'm fine with having a 2 step release. Manually creating the release after merging the changelog changes is easy to document and not that difficult or prone to error. 

I'm sure your time is better spent somewhere else than this ;) Sorry I sent you on this goose hunt.

chrysle:
> IIUC, those release related jobs/steps should belong to a different workflow, which maybe `release.yml`

Yup, you're seeing an outdated version here, actual work happened at https://github.com/chrysle/pipx/tree/test-cd. But your comment inspired me to attempt moving all the release-related jobs into this workflow, and I finally managed, although I went half crazy (this is the last time I've done anything with GH Actions here, rest assured). Here the [successful run](https://github.com/chrysle/pipx/actions/runs/9082650617/job/24959566494) and the [release it created](https://github.com/chrysle/pipx/releases/tag/1.4.1). If you think that's all in order, I'll merge the development branch; otherwise, I'll give up.

huxuan:
> If you think that's all in order, I'll merge the development branch

IMO, it looks pretty good.

Previously, I just [use a tag in semver to trigger the release process](https://github.com/serious-scaffold/ss-python/blob/4e95caca0c0d33c5e268c57ab96e0ba065942664/.github/workflows/release.yml#L123). This two steps release process is quite a good idea. I may also apply it in my personal projects. :-)

chrysle:
Now merged. I extended the logic extracting the semantic version from the PR title to work with any with the title format `<Version>: Description` and updated the docs accordingly – that way we'll also be able to make manual amendments.

huxuan (dismissed):
LGTM and great job! Let us wait for the feedback from other guys.

CONTRIBUTING.md
```diff
@@ -220,10 +220,12 @@ nox -s watch_docs
 
 ## Releasing New `pipx` Versions
 
-To publish to PyPI simply create a "published" release on GitHub. This will trigger GitHub workflows that publishes:
+To release a new version, manually run the `bump-changelog` action under the *"Actions"* tab, passing it the version to
+be released. This will create a pull request updating the changelog for the upcoming version, with the `release-version` label. Attaching this label to any pull request of which the title follows the format `<Version>: Description` and merging it will trigger GitHub workflows that publish:
```
Gitznik, line 224: This to me is a bit clearer.

```suggestion
To release a new version, manually run the `bump-changelog` action under the *"Actions"* tab, passing it the version to
be released. This will create a pull request updating the changelog for the upcoming version, with the `release-version`
label. Merging this PR will automatically trigger the release workflows.

Attaching this label to any pull request of which the title follows the format `<Version>: Description` and merging it will 
trigger the release workflows as well.

The release workflow consists of publishing:
```

Gitznik (changes requested):
Great job! Developing workflows like this I know is very painful, so massive thanks for following through.

Gitznik (dismissed):
Great stuff

.github/workflows/release.yml
```diff
@@ -0,0 +1,81 @@
+name: Release
+
+on:
+  pull_request_target:
+    types:
+      - closed
+concurrency:
+  group: tests-${{ github.ref }}
```
huxuan, line 8: We might need to change the group name here. And maybe we should not "cancel-in-progress"?

.github/workflows/bump-changelog.yml
```diff
@@ -0,0 +1,52 @@
+name: Bump changelog before release
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version to be released'
+        required: true
+        type: string
```
huxuan, line 9: Better to add a "concurrency" section here to avoid dupliced creation?

huxuan (commented):
Just realize some concern about concurrency.

.github/workflows/release.yml
huxuan: It would be more general to use `${{ github.workflow }}-${{ github.ref }}` instead here.

chrysle:
Thanks for correcting my oversight! Updated.

huxuan (approved):
Great job!

gaborbernat (approved):

Issue #1249, opened 2024-02-07
Command to update local standalone python versions


Gitznik:
**How would this feature be useful?**

https://github.com/pypa/pipx/issues/1242 introduced the ability to have pipx download standalone python versions, if the requested python version is not available. Currently, there is no way through pipx to update them. Having a way for pipx to figure out which versions have updates available and applying them would be very useful for housekeeping.

Also see the discussions in the merge review of the feature for more information: https://github.com/pypa/pipx/pull/1243#issuecomment-1932170921

**Describe the solution you'd like**

A new command that checks the downloaded standalone python versions, checks for updates to these versions and updates them.

**Describe alternatives you've considered**

I don't really think there is one.

alextremblay:
Hey @Gitznik,

I've spent the last ~month of my spare time spinning out the code I wrote for managing python distributions into its own tool / library, and I've added support for updating installed distributions.

Check it out: https://github.com/alextremblay/pyvm
and on pypi: https://pypi.org/project/python-version-manager/

feel free to vendor it or use some of its code to improve the pipx implementation :)

Gitznik:
Interesting, thanks for creating, sharing and allowing us to vendor/use the code. Certainly a great resource for when this ticket is picked up :)

huxuan:
Hi @Gitznik,

I am interested in this issue, but have several questions:

1. Seems our first goal is to upgrade the Python version used by pipx-managed applications, so the behavior is more like `pipx reinstall --python <new_python_version> --fetch-missing-python <package_name>`, right? If so, is there any expectation for the "new command", maybe `pipx upgrade-python <package_name>`?
2. Any expectation to the upgrade strategy? From the current implementation [1], I think the intuitive idea is to upgrade the patch version, e.g., from `3.12.2` to `3.12.3` when available. Do we need to support something like from `3.11` to `3.12`? And this is also related to the CLI interaction.
3. Do we need an extra command to manage those installed standalone Python? The functionality will be more like `pyvm` provided. BTW, `pdm` also provide similar features [2].

[1] https://github.com/pypa/pipx/blob/49c998399ae0ff5fd97f748e7135214c83fe098e/src/pipx/standalone_python.py#L193
[2] https://pdm-project.org/en/latest/reference/cli/#python

Gitznik:
Hi @huxuan, great you're looking into this.

This command should allow updating patch versions. I don't think there's a use case for updating minor versions. As you said, switching the minor version should be a reinstall.

[...]

Issue #1248, opened 2024-02-07
Command for cleanup of installed standalone python versions


Gitznik:
**How would this feature be useful?**

#1242 introduced the ability to have pipx download standalone python versions, if the requested python version is not available. Currently, there is no way through pipx to delete them again. Having a way for pipx to figure out which interpreters are still in use and delete the ones that aren't would be very useful for housekeeping.

Also see the discussions in the merge review of the feature for more information: https://github.com/pypa/pipx/pull/1243#issuecomment-1932170921

**Describe the solution you'd like**

A new command that checks the interpreters of the installed applications, compares it with the available standalone python versions and deletes the versions that are no longer in use.

**Describe alternatives you've considered**

Manually going to the `PIPX_STANDALONE_PYTHON_CACHEDIR` folder and removing python versions, hoping they are not used anymore.

PR #1251, opened 2024-02-07
Add source interpreter to pipx metadata

Gitznik:

## Summary of changes

I added the source interpreter of a venv to the pipx metadata in preparation of #1248 and #1249. We need this to be able to find out if a standalone interpreter is still being used.

I've done it this way following the logic of the `python_version` part of the pipx metadata. Is there any reason we don't use the `pyvenv.cfg` for obtaining this information? It has the python version and the source interpreter in it, and it has been stable since python 3.3 AFAIK.

I have not added a news fragment yet, as IMO this is part of #1243 

## Test plan

Tested by running

```
pipx install --fetch-missing-python --python { a python version you don't have} pycowsay
pipx list
```


Gitznik:
FYI @chrysle and @alextremblay, it seems the standalone python downloaded on windows right now can't actually be executed. I've added executing it to the test to validate this. Did you come across this in your tests @alextremblay?

@chrysle for us this means we can't release this feature just yet.

alextremblay:
> FYI @chrysle and @alextremblay, it seems the standalone python downloaded on windows right now can't actually be executed. See the windows test workflow. I've added executing it to the test to validate this. Did you come across this in your tests @alextremblay?
> 
> @chrysle for us this means we can't release this feature just yet.

oh dear, that's unfortunate

as i mentioned in the PR for standalone python, i don't have a windows machine to test with, and i don't have much experience developing python projects in windows

maybe i can get a windows VM spun up for development or something... that might take me a few days

Gitznik:
> as i mentioned in the PR for standalone python, i don't have a windows machine to test with, and i don't have much experience developing python projects in windows
> 
> maybe i can get a windows VM spun up for development or something... that might take me a few days

Same here :+1:

Gitznik:
Great news. The `installed_python` path was wrong on windows. This is fixed now and ready for review.

chrysle:
Thanks for catching! You had a look at the source tarball, right?

Gitznik:
> Thanks for catching! You had a look at the source tarball, right?

I actually setup a windows VM and tried it out there.

chrysle:
> I actually setup a windows VM and tried it out there.

Well, I'm all for my way of debugging ;-)

```
$ tar tvf cpython-3.10.13+20240107-x86_64-pc-windows-msvc-shared-install_only.tar.gz | grep "python.exe"
-rwxrwxrwx root/root    255488 2021-01-01 09:00 python/Lib/venv/scripts/nt/python.exe
-rwxrwxrwx root/root     91136 2021-01-01 09:00 python/python.exe
$ tar tvf cpython-3.10.13+20240107-x86_64_v2-unknown-linux-gnu-install_only.tar.gz | grep "bin/python"
lrwxrwxrwx root/root         0 2021-01-01 09:00 python/bin/python3 -> python3.10
lrwxrwxrwx root/root         0 2021-01-01 09:00 python/bin/python3-config -> python3.10-config
-rwxrwxr-x root/root     20944 2021-01-01 09:00 python/bin/python3.10
-rwxrwxr-x root/root      3130 2021-01-01 09:00 python/bin/python3.10-config
```

Will review shortly.

Gitznik:
> Well, I'm all for my way of debugging ;-)
> 
> ```
> $ tar tvf cpython-3.10.13+20240107-x86_64-pc-windows-msvc-shared-install_only.tar.gz | grep "python.exe"
> -rwxrwxrwx root/root    255488 2021-01-01 09:00 python/Lib/venv/scripts/nt/python.exe
> -rwxrwxrwx root/root     91136 2021-01-01 09:00 python/python.exe
> $ tar tvf cpython-3.10.13+20240107-x86_64_v2-unknown-linux-gnu-install_only.tar.gz | grep "bin/python"
> lrwxrwxrwx root/root         0 2021-01-01 09:00 python/bin/python3 -> python3.10
> lrwxrwxrwx root/root         0 2021-01-01 09:00 python/bin/python3-config -> python3.10-config
> -rwxrwxr-x root/root     20944 2021-01-01 09:00 python/bin/python3.10
> -rwxrwxr-x root/root      3130 2021-01-01 09:00 python/bin/python3.10-config
> ```
> 
> Will review shortly.

Yeah that would've been probably the way more efficient approach :sweat_smile: ah well, now I have a windows VM for debugging.

src/pipx/pipx_metadata_file.py
```diff
@@ -109,6 +115,9 @@ def from_dict(self, input_dict: Dict[str, Any]) -> None:
         input_dict = self._convert_legacy_metadata(input_dict)
         self.main_package = PackageInfo(**input_dict["main_package"])
         self.python_version = input_dict["python_version"]
+        self.source_interpreter = (
+            Path(input_dict["source_interpreter"]) if input_dict["source_interpreter"] is not None else None
```
chrysle, line 119: This should work?

```suggestion
            Path(input_dict["source_interpreter"]) if input_dict["source_interpreter"] else None
```

src/pipx/venv.py
```diff
@@ -176,6 +177,9 @@ def create_venv(self, venv_args: List[str], pip_args: List[str], override_shared
 
         self.pipx_metadata.venv_args = venv_args
         self.pipx_metadata.python_version = self.get_python_version()
+        source_interpreter = shutil.which(self.python)
+        if source_interpreter is not None:
```
chrysle, line 181:
```suggestion
        if source_interpreter:
```

tests/test_interpreter.py
```diff
@@ -205,5 +193,7 @@ def which(name):
         assert str(PIPX_STANDALONE_PYTHON_CACHEDIR) in python_path
         if WINDOWS:
             assert python_path.endswith("python.exe")
+            subprocess.run([python_path, "-c", "import sys; print(sys.executable)"], check=True)
         else:
             assert python_path.endswith("python3")
+            subprocess.run([python_path, "-c", "import sys; print(sys.executable)"], check=True)
```
chrysle, line 199: How about:

```suggestion
        else:
            assert python_path.endswith("python3")
        subprocess.run([python_path, "-c", "import sys; print(sys.executable)"], check=True)
```

src/pipx/pipx_metadata_file.py
```diff
@@ -46,7 +46,7 @@ class PackageInfo(NamedTuple):
 
 class PipxMetadata:
     # Only change this if file format changes
-    __METADATA_VERSION__: str = "0.3"
+    __METADATA_VERSION__: str = "0.4"
```
chrysle, line 49: Should we add some documentation on what changes the different versions introduce in the future?

chrysle (commented):
Please also add a newsfragment for the bugfix.

Gitznik:
Thanks for the review.

> Please also add a newsfragment for the bugfix.

Do you reckon we need a news fragment for the bugfix? The bug was never released, so it's not really a change for the user.

Also what do you think about using the `pyvenv.cfg` to obtain the source interpreter, instead of writing it to our own metadata? I just didn't do it because it's not done for the python version either, but I don't know why it wasn't done there.

chrysle:
> Do you reckon we need a news fragment for the bugfix? The bug was never released, so it's not really a change for the user.

That's true, didn't think of it.

> Also what do you think about using the `pyvenv.cfg` to obtain the source interpreter, instead of writing it to our own metadata? I just didn't do it because it's not done for the python version either, but I don't know why it wasn't done there.

Well, the `executable` key to `pyvenv.cfg` was added in Python 3.11 ([changelog](https://docs.python.org/3.11/whatsnew/changelog.html#id153)). So I think we should delay switching the behaviour until then.

By the way, you'll need to update [`helpers.py`](https://github.com/pypa/pipx/blob/main/tests/helpers.py).

src/pipx/pipx_metadata_file.py
```diff
@@ -82,19 +83,24 @@ def to_dict(self) -> Dict[str, Any]:
...
             "pipx_metadata_version": self.__METADATA_VERSION__,
         }
 
     def _convert_legacy_metadata(self, metadata_dict: Dict[str, Any]) -> Dict[str, Any]:
-        if metadata_dict["pipx_metadata_version"] in ("0.2", self.__METADATA_VERSION__):
+        if metadata_dict["pipx_metadata_version"] in (self.__METADATA_VERSION__):
+            return metadata_dict
+        elif metadata_dict["pipx_metadata_version"] in ("0.2", "0.3"):
+            metadata_dict["source_interpreter"] = None
             return metadata_dict
```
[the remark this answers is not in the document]
Gitznik: Done

src/pipx/pipx_metadata_file.py
Gitznik: Done

src/pipx/venv.py
Gitznik: Done

tests/test_interpreter.py
Gitznik: Done

src/pipx/pipx_metadata_file.py
Gitznik: Any more suggestions for the docs I added? I personally think for more details one should dig through the commit history. 

Could maybe add links to the respective PRs, but even that is very easy to find via git.

chrysle (approved):

PR #1252, opened 2024-02-08
Inject additional packages from text file

### LINKED ISSUE #934: Support for `pipx inject <package> -r <requirements-file>`

jamesmyatt:

## Summary of changes

Fixes #934. Provides for injecting dependencies from a text (requirements) file.

Replaces #1037, which it's partially based on. Thanks, @dukecat0!

I don't think it's necessary to match the pip functionality when the "runpip" command also exists. Instead it replaces the workaround with xargs:

```bash
cat <requirements-file> | sed -e 's/#.*//' | xargs pipx inject <package>
```

It could be extended to support the full pip-requirements syntax in the future, if necessary, but I'd prefer "good" rather than "perfect".

## Test plan

Tested by running unit tests


src/pipx/main.py
```diff
@@ -415,9 +416,19 @@ def _add_inject(subparsers, venv_completer: VenvCompleter, shared_parser: argpar
...
     )
+    p.add_argument(
+        "-r",
+        "--requirement",
+        dest="requirements",
+        action="append",
+        default=[],
+        metavar="file",
+        help="file containing the packages to inject into the Virtual Environment--"
+        "one package name or pip package spec per line.",
```
chrysle, line 430:
```suggestion
        help=(
            "file containing the packages to inject into the Virtual Environment--"
            "one package name or pip package spec per line. "
            "May be specified multiple times."
         ),
```

tests/test_inject.py
```diff
@@ -47,3 +49,19 @@ def test_inject_include_apps(pipx_temp_env, capsys, with_suffix):
         assert run_pipx_cli(["inject", "pycowsay", PKG["black"]["spec"], "--include-deps"])
 
     assert not run_pipx_cli(["inject", f"pycowsay{suffix}", PKG["black"]["spec"], "--include-deps"])
+
+
+def test_inject_with_req_file(pipx_temp_env, capsys, tmp_path):
+    req_file = tmp_path / "requirements.txt"
```
chrysle, line 55: Since that is, as you mentioned, not a regular `requirements.txt` file, it's probably better to modify is slightly?

```suggestion
    req_file = tmp_path / "inject-requirements.txt"
```

src/pipx/commands/inject.py
```diff
@@ -116,3 +132,17 @@ def inject(
 
     # Any failure to install will raise PipxError, otherwise success
     return EXIT_CODE_OK if all_success else EXIT_CODE_INJECT_ERROR
+
+
+def parse_requirements(filename: Union[str, os.PathLike]) -> Generator[str, None, None]:
+    """
+    Extracts package specifications from requirements file.
+
+    Just returns all of the non-empty lines with comments removed.
```
chrysle, line 141: This version is PEP 257-compliant ;-)

```suggestion
    Extract package specifications from requirements file.

    Return all of the non-empty lines with comments removed.
```

src/pipx/commands/inject.py
```diff
@@ -11,6 +13,10 @@
 from pipx.util import PipxError
 from pipx.venv import Venv
 
+logger = logging.getLogger(__name__)
+
+COMMENT_RE = re.compile(r"(^|\s+)#.*$")
```
chrysle, line 18: I'd inline that?

changelog.d/1252.feature.md
```diff
@@ -0,0 +1,9 @@
+Add `--requirement` option to `inject` command.
+
+This reads the list of packages to inject from a text file,
+can be used multiple times,
+and can be used in parallel with command line packages.
```
chrysle, line 5:
```suggestion
and can be used in parallel with dependencies specified 
through the command line.

```

chrysle (commented):
Thank you for working on this! I'm a bit unsure regarding the naming, though, since _"Requirements File"_ is a reserved term.

src/pipx/main.py
```diff
@@ -415,9 +416,19 @@ def _add_inject(subparsers, venv_completer: VenvCompleter, shared_parser: argpar
...
+    )
```
Gitznik, line 431: Please add an example on how to use this and how a inject-requirements file would look like to our docs as well. Probably to `docs/examples.md`.

src/pipx/commands/inject.py
```diff
@@ -98,10 +105,19 @@ def inject(
     suffix: bool = False,
 ) -> ExitCode:
     """Returns pipx exit code."""
+    # Combined list of packages
+    packages = list(package_specs)
+    for filename in requirement_files:
+        packages.extend(parse_requirements(filename))
```
Gitznik, line 111: We probably should deduplicate packages here.

Gitznik (changes requested):
Also flagging that in the original PR by @dukecat0 the decision was made to make the requirements file and passing a requirement directly mutually exclusive. 

I personally don't see a technical reason to stick to that, so not a blocker for me.

src/pipx/commands/inject.py
jamesmyatt: That would be ideal, but we don't do that with `package_specs` and pip will just handle it, I expect. So IMO that's nice-to-have, rather than necessary.

src/pipx/commands/inject.py
jamesmyatt: Do you mean move it to `parse_requirements`? It's used in a loop, so it's worth compiling (I assume). Also, this is how it's implemented in https://github.com/pypa/pip/blob/main/src/pip/_internal/req/req_file.py.

jamesmyatt:
> Also flagging that in the original PR by @dukecat0 the decision was made to make the requirements file and passing a requirement directly mutually exclusive.
> 
> I personally don't see a technical reason to stick to that, so not a blocker for me.

I think that's how pip works, but like you say, I don't see a good technical reason for it. I don't even think it's necessary to match the pip functionality. If you want to match the pip functionality, then `runpip` is there. My preference here is to be fairly naive and rely on the users to do sensible things.

jamesmyatt:
> Thank you for working on this! I'm a bit unsure regarding the naming, though, since _"Requirements File"_ is a reserved term.

I agree that _"Requirements File"_ is a reserved term and shouldn't be over-used. However, we're using a strict sub-set of the _"Requirements File"_ syntax, won't accept changes that are incompatible and have ambitions to support the full syntax eventually.

jamesmyatt:
On the subject of full pip-requirements file support. I wonder if the better long-term implementation would be to change the way that `inject` works to just create a temporary requirements file from the CLI inputs and then install that using pip (I think this is what conda does with pip requirements). Then pipx would record everything that's listed as `requested` in the output (based on https://github.com/pypa/pipx/pull/1037#discussion_r1304075979).

But I think that needs more investigation and would be a much more significant change to pipx. Hence, my preference is to implement this solution with partial pip-requirements-file support that is probably good enough for 99% of use cases, and investigate a solution for the remaining 1% in slower time.

src/pipx/commands/inject.py
chrysle, line 111: How about this to de-duplicate?

```suggestion
    # Combined set of packages
    packages = set(package_specs)
    for filename in requirement_files:
        packages = packages | set(parse_requirements(filename))
```

chrysle:
> But I think that needs more investigation (e.g. checking how it works with `upgrade --include-injected`) and would be a much more significant change to pipx. Hence, my preference is to implement this solution with partial pip-requirements-file support that is probably good enough for 99% of use cases, and investigate a solution for the remaining 1% in slower time.

This sounds like a good plan to me.

src/pipx/commands/inject.py
jamesmyatt: That will only de-duplicate by exactly matching the dependency spec strings, including version numbers and white space. I'm not sure there's much value in that. I think that if we're going to de-duplicate, then you'd want to group by the package name and consolidate the version constraints into a single line, which will be very complicated.

src/pipx/commands/inject.py
```diff
@@ -98,10 +105,19 @@ def inject(
...
+    logger.debug("Injecting packages: %r", sorted(packages))
+    if not packages:
+        raise PipxError("No packages have been specified.")
```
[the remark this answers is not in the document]
jamesmyatt: Does it matter? Isn't it helpful to have exactly what `packages` is in the DEBUG logs before the error is raised?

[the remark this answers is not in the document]
chrysle: I don't insist on this.

> Isn't it helpful to have exactly what `packages` is in the DEBUG logs before the error is raised?


There aren't much possibilities apart from zero length, are there?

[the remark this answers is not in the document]
jamesmyatt: I can't think of any. I'm happy to do whatever you and @gaborbernat prefer.

Also do you prefer this logging at DEBUG or INFO? Probably INFO actually matches the other logging messages better.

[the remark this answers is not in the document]
chrysle: > Probably INFO actually matches the other logging messages better.

Agreed.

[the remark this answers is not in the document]
jamesmyatt: Done

src/pipx/commands/inject.py
jamesmyatt: I've implemented this. It at least removes duplicated package specifications.

chrysle:
@jamesmyatt Gentle ping on this.

jamesmyatt:
Thanks for the reminder. I think the outstanding actions are:

* [x] Fix merge
* [ ] Check docs are clear about relationship to pip-requirements files
* [ ] Add examples to documentation
* [ ] Check log and std outputs in unit tests

chrysle (commented):
Thanks for your initiative in getting this over the finish line! A few nits regarding the docs, otherwise LGTM.

jamesmyatt:
No problem. Thanks @chrysle .

jamesmyatt:
@chrysle Do you have an idea why the new unit tests are _sometimes_ failing?

tests/test_inject.py
```diff
@@ -1,12 +1,36 @@
...
 from package_info import PKG
 
 
-def test_inject_simple(pipx_temp_env, capsys):
+@pytest.mark.parametrize(
+    "pkg_spec,",
+    [
+        PKG["black"]["spec"],  # was test_inject_simple
+        "jaraco.clipboard==2.0.1",  # was test_inject_tricky_character
+        "pylint==3.0.4",  # was test_spec
```
jamesmyatt, line 16: This duplicates the `PKG["pylint"]["spec"]` test, but I wanted to explicitly match the previous tests. However I think this one could just be deleted, and the old test names removed.

tests/test_inject.py
```diff
@@ -27,16 +51,6 @@ def test_inject_simple_legacy_venv(pipx_temp_env, capsys, metadata_version):
...
-    assert not run_pipx_cli(["inject", "pycowsay", "jaraco.clipboard==2.0.1"])
-
-
-def test_spec(pipx_temp_env, capsys):
-    assert not run_pipx_cli(["install", "pycowsay"])
-    assert not run_pipx_cli(["inject", "pycowsay", "pylint==3.0.4"])
-
-
 @pytest.mark.parametrize("with_suffix,", [(False,), (True,)])
 def test_inject_include_apps(pipx_temp_env, capsys, with_suffix):
```
jamesmyatt, line 55: I don't understand this unit test

huxuan:
> @chrysle Do you have an idea why the new unit tests are _sometimes_ failing?

isort is one of the dependencies of pylint, so you may need to change the test case.

chrysle:
> isort is one of the dependencies of pylint, so you may need to change the test case.

I'd suggest to just drop `pylint` as a test package in this case.

jamesmyatt:
> > @chrysle Do you have an idea why the new unit tests are _sometimes_ failing?
> 
> isort is one of the dependencies of pylint, so you may need to change the test case.

Ah yes. I wonder why it works sometimes then.

tests/test_inject.py
chrysle: Yes, I agree.

tests/test_inject.py
chrysle: It's rather a functional test, and seems to combine testing including the exposed applications of deps to be injected with the suffix behaviour. However a bit more output debugging could be useful here.

jamesmyatt:
I think this is actually done now

chrysle (dismissed):

chrysle:
Waiting for @Gitznik's review before merging.

changelog.d/1252.feature.md
```diff
@@ -0,0 +1,16 @@
...
+The option can be abbreviated to `-r`.
+
+Each line is a separate package specifier with the same syntax as the command line.
+Comments are supported with a `#` prefix.
+Hence, the syntax is a strict subset of the pip [requirements file format][pip-requirements] syntax.
+
+If you require full pip functionality, then use the `runpip` command instead;
+however, the installed packages won't be recognised as "injected".
+
+[pip-requirements]: https://pip.pypa.io/en/stable/reference/requirements-file-format/
```
Gitznik, line 16: This breaks how the changelog is rendered unfortunately: https://pipx--1252.org.readthedocs.build/1252/changelog/
I think I fixed it for the news fragments in the last version by adding bullet points for everything but the first line, leading it to be rendered as a nested unordered list.

Gitznik (changes requested):
LGTM aside from the news fragment :+1:

changelog.d/1252.feature.md
jamesmyatt: Is it also too long?

jamesmyatt: Looking at https://github.com/pypa/pipx/pull/1310/files, I think it should be 1 title paragraph and maybe 1 supporting paragraph.

jamesmyatt: Actually, I don't think it automatically handles multipart log entries. So I've removed all of the detail and moved it to the examples.

However, https://github.com/mayeut/pipx/blob/main/changelog.d/1316.feature.md is broken.

Gitznik (approved):

changelog.d/1252.feature.md
Gitznik: Yeah I saw. If you feel like it feel free to fix it. Otherwise, I can do it in a separate PR

jamesmyatt: I'll leave it to you. I think the details need moving like I did here.

huxuan:
Great job!

Issue #609, opened 2021-01-23
Need documentation or fix to use private index with shared libs


itsayellow:
**How would this feature be useful?**

Related: #544 #604.

Right now users can use `pip_args` to install a package from a private index.  But there is no pipx-specific argument that directs the installation of the shared libs.

We used to borrow the pip_args from a package install, but this was an inconsistent mechanism, because the shared libs can be upgraded during `pipx list` and other commands which have no `pip_args`.

The most reliable way for a user to change such behavior is to set `PIP_*` environment variables, or to set the pip global config (e.g. see https://github.com/pipxproject/pipx/issues/604#issuecomment-763770502)

**Describe the solution you'd like**

My preference would be simply to document what is needed to configure pip.

**Describe alternatives you've considered**

We could add persistent pipx configuration, but this is a complicated solution because pipx doesn't have global persistent configuration at the moment, and it would have to be added.  Plus it's not clear this is easier or better than just configuring pip.

We could also have a dedicated option? or command? that would leave a persistent config in the `pipx/shared` directory, possibly similar or identical to pipx_metadata.json.  However this would add complexity and have to be recreated by the user if the shared directory were ever deleted.  Again, it's not clear this is better or easier than configuring pip.

uranusjr:
This is difficult because the user would expect the answer to be yes for some and no for some. There’s an argument to inherit `--proxy` etc., but for `--no-binary`, `--ignore-installed`, etc. the answer is very likely no. So the safe answer is likely no, and provide a separate command to update shared libs with `--pip-args` applied.

itsayellow:
> This is difficult because the user would expect the answer to be yes for some and no for some. There’s an argument to inherit `--proxy` etc., but for `--no-binary`, `--ignore-installed`, etc. the answer is very likely no. So the safe answer is likely no, and provide a separate command to update shared libs with `--pip-args` applied.

I'm not following--what's the "yes" or "no" questions you're talking about?

Are you advocating for something like `pipx shared --pip-args="..."`?  That would then leave a `pipx_metadata.json` in the `shared/` venv?

itsayellow:
My assumption is that the most likely thing config needed to be used with the shared libraries are things like mentioned in #604, like options affecting index, cert, trusted host, proxy.  I can't imagine anything else being useful for shared libs.  And those things seem pretty broadly applicable to me.

pawamoy:
[...]

PR #1256, opened 2024-02-14
Support pip_args for  shared_libs-enabled virtual environments

### LINKED ISSUE #964: Pipx install does not honour --index-url when it updates pip/setuptools/wheel
**Describe the bug**
Pipx install does not honour --index-url when it updates pip/setuptools/wheel.

I am calling `pipx install` with an `--index-url <replace_with_artifactory _url>` and `--pip-args "--index-url <replace_with_artifactory _url>"` which correctly installs the given package using the correct index url.  

However when I use pipx install with `--verbose` I can see that underneath the hood there is a call to the following, which does not pass through the `--index-url`:
`python -m pip --disable-pip-version-check install --force-reinstall --upgrade pip setuptools wheel` 

Are there any config settings that I can fix this with OR can I stop the reinstall of the pip/setuptools/wheel call please?

**How to reproduce**
Run:
`pipx install <whatever_package> --index-url <host_of_artifact_url>`

**Expected behavior**
Should see that it uses the passed in `--index-url`:
`python -m pip --disable-pip-version-check install --index-url <host_of_artifact_url> --force-reinstall --upgrade pip setuptools wheel`

Arpafaucon:

## Summary of changes

Fix #964

My diagnostic of the issue is:
- the `shared_libs` feature creates a verification task **at each `Venv` instance creation** that checks if shared libraries should be updated
-  this verification task can decide to run `pip` commands, and if so, we'd like to have it respect user-provided `pip_args` options
- the problem: **sometimes `Venv` instances are created in a context where the user does not expect pip to run**. An example of this is #1081 
  - the `list` command creates `Venv` instances, those run the shared_lib verification
  - the `list` CLI options do not enable `--pip_args` (and IMHO even if it did, users would not expect those here)

Change summary
- move the shared lib verification to a dedicated method `Venv. check_upgrade_shared_libs`
- call this method only in contexts where a user would expect `pip install/upgrade` operations (we can identify these because we have `pip_args` available), just after the `Venv` instance creation
- thus there will be some places where the shared lib check will not be run anymore
  - notably during `pipx list`, so `--skip-maintenance` is a no-op

## Test plan

Tested as part of the main test suite. Added tests
- `test_install.py::test_pip_args_forwarded_to_shared_libs`
- `test_reinstall_all.py::test_reinstall_all_triggers_shared_libs_upgrade`
- `test_run.py::test_pip_args_forwarded_to_shared_libs`
- in `test_list.py`, renamed `test_skip_maintenance`->`test_list_does_not_trigger_maintenance`


src/pipx/shared_libs.py
```diff
@@ -50,7 +50,8 @@ def create(self, verbose: bool = False) -> None:
 
             # ignore installed packages to ensure no unexpected patches from the OS vendor
             # are used
-            self.upgrade(pip_args=["--force-reinstall"], verbose=verbose)
+            pip_args = (pip_args or []) + ["--force-reinstall"]
```
chrysle, line 53:
```suggestion
            pip_args = (pip_args or []).append("--force-reinstall")
```

src/pipx/venv.py
```diff
@@ -77,9 +77,9 @@ def get_venv_dir(self, package_name: str) -> Path:
         """Return the expected venv path for given `package_name`."""
         return self._root.joinpath(canonicalize_name(package_name))
 
-    def verify_shared_libs(self) -> None:
+    def verify_shared_libs(self, pip_args: List[str], verbose: bool) -> None:
```
chrysle, line 80: I don't really understand the purpose of this function, BTW.

```suggestion
    def verify_shared_libs(self, pip_args: Optional[List[str]] = None, verbose: bool) -> None:
```

src/pipx/venv.py
```diff
@@ -97,12 +97,13 @@ def __init__(self, path: Path, *, verbose: bool = False, python: str = DEFAULT_P
         except StopIteration:
             self._existing = False
 
+    def check_upgrade_shared_libs(self, pip_args: List[str], verbose: bool):
```
chrysle, line 100:
```suggestion
    def check_upgrade_shared_libs(self, pip_args: Optional[List[str]] = None, verbose: bool):
```

chrysle (commented):
That was quick! I generally agree with the behaviour change. You'll need to create a functional test for the behaviour.

chrysle:
>   * notably during `pipx list`, so `--skip-maintenance` is a no-op

cc @Gitznik

src/pipx/shared_libs.py
Arpafaucon: mm this won't work `append` works in-place. I could do
```py
pip_args = pip_args or []
pip_args.append("--force-reinstall")
```

src/pipx/venv.py
Arpafaucon: a small grep tells me that it's not used anywhere - should I get rid of it ?

Arpafaucon:
> That was quick! I generally agree with the behaviour change. You'll need to create a functional test for the behaviour.

Could you help me with some pointers for that ? I don't know how to properly test that we actually forwarded the correct pip-args all the way down

src/pipx/venv.py
chrysle: Though I like the `--skip-maintenance` flag. I'm not sure we should always refrain from upgrading the libs when listing packages.

src/pipx/venv.py
Arpafaucon: I don't know the codebase well, so I'm confused by your suggestion here. (I may be missing contribution guidelines that clarifiy this - and I definitely lack experience in maintenance of widely-used open-source projects. )

If I understand, we agree on the fact that the point of the MR is to get rid of scenarios where shared libs operations ignore user-provided pip args under the hood.

So, isn't it "dangerous" to accept `None` as a default value here ? I felt that *not* providing an acceptable value would better document to future developpers that calling this function is not acceptable if we don't have  a user-provided `pip_args`

Arpafaucon:
> > I don't know how to properly test that we actually forwarded the correct pip-args all the way down
> 
> You could parse the verbose output?

Thanks: I added a test `test_install.py::test_pip_args_forwarded_to_shared_libs` that checks that providing `--no-index` to `pipx install` will effectively break the shared lib update. This is kind of indirect, but has the benefit of not needing any network - and the breakage is easy to test

src/pipx/venv.py
Arpafaucon: In that case should we provide a `--pip-args` there too ? 

Otherwise, if I understand correctly, that means that `pipx list` can arbitrarily connect to `pypi`, and user would not be able to provide a custom index_url nor certificate.

Arpafaucon:
> > > I don't know how to properly test that we actually forwarded the correct pip-args all the way down
> > 
> > 
> > You could parse the verbose output?
> 
> Thanks: I added a test `test_install.py::test_pip_args_forwarded_to_shared_libs` that checks that providing `--no-index` to `pipx install` will effectively break the shared lib update. This is kind of indirect, but has the benefit of not needing any network - and the breakage is easy to test.
> 
> If the general principle is OK for you, I can work on adding the same concept to other affected pipx commands
> 
>     * upgrade
> 
>     * run
> 
>     * inject
> 
>     * reinstall
> 
> 
> Let me know if that fits your expectations, or if you have further advice for me

Mmm my test methodology stops working at soon as multiple tests are executed. It looks like the shared libs mecanism is shared across tests ? If you have any advice on this that'd be helpful, on my side I'll keep on experimenting

Gitznik:
> > * notably during `pipx list`, so `--skip-maintenance` is a no-op
> 
> cc @Gitznik

Thanks for the ping. I don't mind this change. Is there any clear benefit of upgrading shared packages unless you want to use them (e.g. install, ...)?

If the `--skip-maintenance` flag is a no-op after this change we should log a deprecation warning and plan on removing the flag in an upcoming release.

src/pipx/venv.py
Gitznik: If we stick with `pip_args: List[str]` I think there's a lot of implicit meaning contained in that parameter. We should add a docstring that explains the logic behind it (commands that allow modifying the `pip_args` providing it as an empty list, even if there's no arguments provided)

src/pipx/venv.py
chrysle: > In that case should we provide a `--pip-args` there too ?

Yes, that would be the logical consequence.

Arpafaucon:
Hey - getting back after a brief gap. I'm still up to carry this change and get to a change that would satisfy pipx team.

I've seen you added the 'awaiting response' badge so I guess there is a misunderstanding - I was myself waiting for answers from you ^^ To clarify and unblock the discussion:

- regarding the maintenance operation during `pipx list`
  - (A) do we add a `--pip-args` to enable user customization
  - (B) or do we avoid any maintenance there? If so, we should get rid of the `--skip-maintenance` flag as suggested by @Gitznik ? 
  
  My humble opinion is **(B)**:  I wouldn't expect any maintenance task to happen during a `list` operation. 
- regarding the test strategy, if the sample test looks like something you'd accept:  I'll keep experimenting. I had stopped working on it awaiting feedback - but I will take the absence of criticism as a positive answer and carry on :)

chrysle:
>  My humble opinion is **(B)**:  I wouldn't expect any maintenance task to happen during a `list` operation.

Let's make it so, then.

Arpafaucon:
Thanks for the clarifications, I reworked on this to polish it

**skip-maintenance CLI argument**
I changed the CLI to make `--skip-maintenance` a no-op. I haven't found guidelines on how to implement deprecations here, nor an example of how it was done in the past, so I improvised a bit:
  - if the flag is enabled, I added a `logger.warning` message
  - I updated the doc to mention it is now a no-op
  - the `commands.list` function does not take it as an argument anymore
  - Should I also raise a DeprecationWarning ? With its current configuration, it seems that `pipx` does not show them, I could try and fix that

**failing tests**
I managed to fix the test by using a better fixture, now the whole test suite for 3.10 works on my machine - I'll check the full CI once you approve the workflow. I added the same for `pipx run` because that was trivial, but I am unsure whether it's useful to work more to test the same behaviour for the other commands

Gitznik:
> **skip-maintenance CLI argument** I changed the CLI to make `--skip-maintenance` a no-op. I haven't found guidelines on how to implement deprecations here, nor an example of how it was done in the past, so I improvised a bit:
> 
> * if the flag is enabled, I added a `logger.warning` message
> * I updated the doc to mention it is now a no-op
> * the `commands.list` function does not take it as an argument anymore
> * Should I also raise a DeprecationWarning ? With its current configuration, it seems that `pipx` does not show them, I could try and fix that

I don't think we have to dive deeper into deprecation warnings, as https://github.com/pypa/pipx/issues/1278 plans to introduce a global flag for `--skip-maintenance`, which will reintroduce it (even though it's a no-op for `list`).

src/pipx/main.py
```diff
@@ -279,14 +279,19 @@ def run_pipx_command(args: argparse.Namespace) -> ExitCode:  # noqa: C901
...
             skip=skip_list,
             force=args.force,
+            pip_args=pip_args,
         )
     elif args.command == "list":
+        if args.skip_maintenance:
+            # TODO: should I raise a DeprecationWarning here as well ?
+            logger.warning(
+                "Deprecation notice: `--skip-maintenance` is now a no-op, because maintenance tasks are now always skipped in `pipx list`. You can safely remove this flag from your calls."
+            )
```
Gitznik, line 289: Sorry about the back and forth around this. IMO you can remove the warning, as the flag will be globally introduced via https://github.com/pypa/pipx/issues/1278.

Gitznik (changes requested):

src/pipx/main.py
Arpafaucon: removed

Gitznik:
> I'm having trouble making the tests on windows pass - to be fully honest I don't understand the root cause of why this platform behaves differently from ubuntu+macOs (but hey, windows...)

We've all been there

> Sorry for the multiple CI runs, I don't have another way to test on windows ATM

You can also enable workflows on your fork so you don't have to wait for us approving each run :)

christian-krieg:
Hey there, first of all, thanks for this fix and great conversation so far! As I am waiting for this PR to be closed in order to start working on #1278, I would like to kindly ask when I may expect the merge? :)

src/pipx/venv.py
chrysle: Probably it _is_ better to be on the safe side and move the shared libs upgrade into a dedicated command, as said in https://github.com/pypa/pipx/issues/609#issuecomment-765833354. While that would be mutually exclusive with a global `--skip-maintenance` option. Introducing the latter, we could document to pass it whenever one does not want to forward the `pip` args to the `pipx` shared libs upgrade, which I think is a practical solution, since  
upgrading the lib by hand would mean extra efforts for the users and could be easily forgotten.

Generally, I doubt users will often make use of complex setups with `pipx` that could damage the shared library in any way. So I think this solution is good for now.

Gitznik (dismissed):

Gitznik:
@chrysle , @Arpafaucon I actually don't think there's anything missing?

changelog.d/1081.removal.md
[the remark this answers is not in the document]
Arpafaucon: done

Gitznik (approved):

chrysle (approved):

Arpafaucon:
> @chrysle , @Arpafaucon I actually don't think there's anything missing?

AFAIK not on my side. There are 2 points by @chrysle  waiting to be resolved
- the first one on changelogs should be good
- there was also a discussion about explicit-vs-implicit default value for `pip_args`. This PR leans strongly on the explicit side  (`pip_args: List[str]`), but I'm not certain this fits @chrysle 's vision as he had the opposite suggestion (I remember he suggested `pip_args: Optional[List[str]] = None`, but I can't see the comment anymore). This was some time ago so maybe the current state is now OK

PR #1267, opened 2024-02-25
Add `--install` option to `pipx upgrade` command

### LINKED ISSUE #1262: pipx upgrade fails when package is not installed
**Describe the bug**

The documentation at https://pipx.pypa.io/stable/docs/#pipx-upgrade says that `upgrade` is:
> Upgrade a package in a pipx-managed Virtual Environment by running `pip
install --upgrade PACKAGE`

`pipx upgrade` fails when a package is not installed, so that is not the behavior.

**How to reproduce**

```shell
$ pipx --version
1.4.3

$ pipx --verbose upgrade dummy_test
Package is not installed. Expected to find /Users/example/.local/pipx/venvs/dummy-test, but it does not exist.
```

It should have installed it if not installed, and upgraded it if installed.

chrysle:
Closes #1262.

## Test plan

Tested by running

```
pipx upgrade --install -p 3.10 pycowsay
```

### DIFF
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -267,10 +267,12 @@ def run_pipx_command(args: argparse.Namespace) -> ExitCode:  # noqa: C901
     elif args.command == "upgrade":
         return commands.upgrade(
             venv_dir,
+            args.python,
             pip_args,
             verbose,
             include_injected=args.include_injected,
             force=args.force,
+            install=args.install,
         )
     elif args.command == "upgrade-all":
         return commands.upgrade_all(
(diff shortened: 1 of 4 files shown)


src/pipx/main.py
```diff
@@ -486,6 +488,12 @@ def _add_upgrade(subparsers, venv_completer: VenvCompleter, shared_parser: argpa
         help="Modify existing virtual environment and files in PIPX_BIN_DIR and PIPX_MAN_DIR",
     )
     add_pip_venv_args(p)
+    p.add_argument(
```
Gitznik, line 491: I think you also have to run `add_python_options(p)` to add the `--python` flag.

Gitznik (changes requested):

src/pipx/main.py
Gitznik: Gotcha, sorry I missed that last night

Gitznik (approved):

Issue #1270, opened 2024-02-27
Opportunity for improvement of the type annotations of `pipx.util.dedup_ordered`


sam-xif:
Hello pipx team,

I am working as part of a research team developing a code analysis tool for Python. One of the issues the tool discovered in pipx's codebase is an opportunity to make the type annotations for `pipx.util.dedup_ordered` more specific. 

Here is the definition of `dedup_ordered`:
```
def dedup_ordered(input_list: List[Any]) -> List[Any]:
    output_list = []
    seen = set()
    for x in input_list:
        if x[0] not in seen:
            output_list.append(x)
            seen.add(x[0])

    return output_list
```

In the function body, it is clear that a List with an inner type that has more structure is expected. Based on the usage of the `input_list` argument in this function, and the call that is being made to it in `analyze_pip_output`, the parameter annotation can be changed with certainty to `List[Tuple[str, Any]]`. Since it is also obvious that the return type of this function matches the parameter type, that annotation can be changed as well. I have verified that making this change creates no additional type errors according to `mypy`:

```
# modification to repo is stashed
/.../pipx (main) » mypy ./src/pipx                                                                                                                                                                                                                                                                                                                                                                                                         
src/pipx/standalone_python.py:90: error: Argument 1 to "move" has incompatible type "Path"; expected "str"  [arg-type]
src/pipx/standalone_python.py:172: error: "dict" is not subscriptable, use "typing.Dict" instead  [misc]
src/pipx/commands/interpreter.py:15: error: "list" is not subscriptable, use "typing.List" instead  [misc]
src/pipx/commands/interpreter.py:37: error: "list" is not subscriptable, use "typing.List" instead  [misc]
Found 4 errors in 2 files (checked 30 source files)

# pop off stash
/.../pipx (main) » gstp                                                                                                                                                                                                                                                                                                                                                                                                                   On branch main
Your branch is up to date with 'origin/main'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   src/pipx/util.py

no changes added to commit (use "git add" and/or "git commit -a")
Dropped refs/stash@{0} (bcdde020658610723f83ae73efae94af71af78f0)

# run mypy again
/.../pipx (main*) » mypy ./src/pipx                                                                                                                                                                                                                                                                                                                                                                                                           
[...]

PR #1271, opened 2024-02-27
#1270 Make dedup_ordered type annotations more specific

sam-xif:
- [ x ] I have added a news fragment under `changelog.d/` (if the patch affects the end users)
(patch does not affect end users)
## Summary of changes
Parameter and return type annotations were made more specific for `pipx.util.dedup_ordered` as described in issue #1270.

## Test plan
There are no functional changes to the code.


chrysle (approved):
Looks good, thank you!

Issue #234, opened 2019-10-14
Add repology shield to README


jayvdb:
**How would this feature be useful?**
I believe it is important for pipx to be available as distro packages, and ensure they are kept updated.  A repology shield would help achieve that.  e.g. https://github.com/lilydjwg/nvchecker has it in their README.

Currently there are two entries in repology which cover pipx:

- [`pipx`](https://repology.org/project/pipx/versions)
- [`python:pipx`](https://repology.org/project/python:pipx/versions)

I assume that those two can be merged.

Here is what each shield looks like.  Obviously it will be more impressive once the two are merged.

[![Packaging status](https://repology.org/badge/vertical-allrepos/python:pipx.svg)](https://repology.org/metapackage/python:pipx/versions)

[![Packaging status](https://repology.org/badge/vertical-allrepos/pipx.svg)](https://repology.org/metapackage/pipx/versions)

**Describe the solution you'd like**
In markdown syntax it would be:

`[![Packaging status](https://repology.org/badge/vertical-allrepos/pipx.svg)](https://repology.org/metapackage/pipx/versions)`

gaborbernat:
PR welcome.

PR #1276, opened 2024-03-03
Add repology shield to installation.md

sstandre:

## Summary of changes
This PR adds a repology shield to README as per #234

It only includes [pipx](https://repology.org/project/pipx/versions) since [python:pipx](https://repology.org/project/python:pipx/versions) redirects to the former.

I used the Markdown syntax suggested in the issue, so the shield is left aligned. I can use the same syntax as the badges to center align it, if that is preferred.

## Test plan

Tested by running

```
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ _For comparison to other tools including pipsi, see
 
 <details>
   <summary>Packaging status</summary>
-  
+
   [![Packaging status](https://repology.org/badge/vertical-allrepos/pipx.svg?columns=3&exclude_unsupported=1)](https://repology.org/metapackage/pipx/versions)
 </details>
 
--- a/README.md
+++ b/README.md
@@ -27,12 +27,6 @@ _For comparison to other tools including pipsi, see
 
 ## Install pipx
 
-<details>
-  <summary>Packaging status</summary>
-
-  [![Packaging status](https://repology.org/badge/vertical-allrepos/pipx.svg?columns=3&exclude_unsupported=1)](https://repology.org/metapackage/pipx/versions)
-</details>
-
 > [!WARNING]
 >
 > It is not recommended to install `pipx` via `pipx`. If you'd like to do this anyway, take a look at the
--- a/README.md
+++ b/README.md
@@ -27,7 +27,11 @@ _For comparison to other tools including pipsi, see
 
 ## Install pipx
 
-[![Packaging status](https://repology.org/badge/vertical-allrepos/pipx.svg?columns=3&exclude_unsupported=1)](https://repology.org/metapackage/pipx/versions)
+<details>
+  <summary>Packaging status</summary>
+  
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -18,6 +18,8 @@
 
 </p>
 
+[![Packaging status](https://repology.org/badge/vertical-allrepos/pipx.svg)](https://repology.org/metapackage/pipx/versions)
```
Gitznik, line 21: IMO this would look nicer if was split into 3 columns. I'd also exclude unsupported, to only keep relevant entries.

```suggestion
[![Packaging status](https://repology.org/badge/vertical-allrepos/pipx.svg?columns=3&exclude_unsupported=1)](https://repology.org/metapackage/pipx/versions)
```

Gitznik, line 21: I'd prefer having this under the `Install pipx` section.

Gitznik (changes requested):
Thanks for working on this :)

sstandre:
Thanks for the review. I moved the shield to the Install pipx section, although I am unsure if it fits better above or below the warning:

- Shield above warning
pipx2
- Shield below warning
pipx1

Gitznik (dismissed):
Having the shield above the warning is fine for me.

README.md
```diff
@@ -27,6 +27,8 @@ _For comparison to other tools including pipsi, see
 
 ## Install pipx
 
+[![Packaging status](https://repology.org/badge/vertical-allrepos/pipx.svg?columns=3&exclude_unsupported=1)](https://repology.org/metapackage/pipx/versions)
```
dukecat0, line 30: How about putting it in a collapsible section so that it occupies less space?

```suggestion
<details>
  <summary>Packaging status</summary>
  
  [![Packaging status](https://repology.org/badge/vertical-allrepos/pipx.svg?columns=3&exclude_unsupported=1)](https://repology.org/metapackage/pipx/versions)
</details>
```

Gitznik: Seeing this I'd be +1 on not having it collapsed. It was not taking up that much space in the first place with 3 columns, and I think it being in a collapsible section makes it very hard to find, to the point where I think most people won't find it at all.

WDYT @dukecat0?

dukecat0: Makes sense to me, but do you prefer moving it to `installation.md` instead? I believe that's the place for a more detailed installation instruction.

Gitznik: If the shields work there that'd be fine for me. 

With the current setup, it's broken on the [main page of the docs](https://pipx--1276.org.readthedocs.build/1276/). @sstandre, can you please have a look at fixing that? It's probably because of the `<details>` section, so might be as simple as just moving the shield to the `installation.md` and not having it collapsible. 

> Or in the future, we should move some of the instructions from `README.md` to `installation.md` so we don't need to update both of them when there are updates.

I like the idea of uncluttering the main README :+1:

dukecat0 (approved):

PR #1088, opened 2023-10-15
Added --global option to perform actions for all users (#754)

haxwithaxe:
* [x] I have added an entry to `docs/changelog.md`

## Summary of changes
 Addressing issue #754. Added `--global` option to perform actions for all users. The global option is for pipx and affects all commands except `run` and `completions` by changing the base paths that are used to generate the install directories.
There are two important sets of changes. 
* One is a bare minimum implementation (first two commits).
* The other is a cleaner implementation (third commit).
  * This breaks out the paths into there own submodule since they definitely aren't constants anymore.

These changes also include new pytest tests and necessarily required changes to existing tests.

I'm perfectly happy to go with either of those or a complete rewrite if that's required to make this feature fit into this project. I'm not married to any of the implementation details.

## Test plan
Passes your CI process on github (via your unaltered actions in my fork https://github.com/haxwithaxe/pipx/actions/runs/6518200232).

### Manually tested by running
To verify the changes do what they are supposed to
```
python3 src/pipx --global environment
```

To test all the relevant commands for regressions.
```
python3 src/pipx --global ensurepath
sudo python3 src/pipx --global install pycowsay
sudo python3 src/pipx --global reinstall pycowsay
sudo python3 src/pipx --global inject pycowsay black
sudo python3 src/pipx --global uninject pycowsay black
sudo python3 src/pipx --global runpip pycowsay freeze
sudo python3 src/pipx --global list
sudo python3 src/pipx --global uninstall pycowsay
python3 src/pipx environment
python3 src/pipx ensurepath
python3 src/pipx install pycowsay
python3 src/pipx reinstall pycowsay
python3 src/pipx inject pycowsay black
python3 src/pipx uninject pycowsay black
python3 src/pipx runpip pycowsay freeze
python3 src/pipx list
python3 src/pipx uninstall pycowsay
```

Issue #754, opened 2021-10-25
Option to install package globally for multi-user access


alextremblay:
Hi All, not sure if this qualifies as a feature request, or if it's already possible and just needs to be documented?

pipx by default installs its virtualenvs in a folder inside the user's home directory (ie `~/.local/pipx/venvs/`), and installs the entrypoint binaries for those packages/virtualenvs in the user's personal path (ie `~/.local/bin`).

Is it possible to install packages with pipx such that they are globally executable by all users on a system (ie install entrypoint binaries in `/usr/local/bin`, install virtualenvs somewhere that all users can access them?

**How would this feature be useful?**

Having this ability would make pipx a prime candidate for people looking to deploy python software in multi-user environments. Without this ability, pipx is not usable at all in such situations

**Describe the solution you'd like**

Ideally, some kind of `--shared` or `--global` flag would be the most straightforward solution, but even something like an optional environment variable would be nice

Ex: `sudo pipx install --global cowsay`

**Describe alternatives you've considered**

There are currently no alternatives / workarounds that I'm aware of. I cannot find anything about this in the documentation, the github issues, or on the web

jack1142:
It appears that this is already mentioned in the documentation:
https://pypa.github.io/pipx/installation/#installation-options

Is there any interest in exposing this in pipx for easier use or should users just be expected to create a shell alias if they want to use this regularly?

jack1142:
Something I observed while trying out this solution today - it is impossible to run pretty much all commands (e.g. `pipx list`) without being the owner of `PIPX_HOME` because pipx uses the `logs` subdirectory of `PIPX_HOME` for its logs and that directory isn't accessible to regular users when set to `/opt/pipx`. A workaround for this is giving access to the `/opt/pipx/logs` for everyone but that doesn't seem ideal.

jack1142:
Presumably, it would be safer to take additional pre-cautions such as only using the elevated permissions for installing the package, while building it with a normal user account. I imagine that this, however, might be out of scope for pipx, even if it would be incredibly cool if it could do that :smile:

eode:
I *definitely* would like to have a `--global` option for pipx.  I've put some thought into how this could be done (fairly) reasonably:

* New environment vars:
  * `PIPX_GLOBAL_HOME`
  * `PIPX_GLOBAL_BIN_DIR`
* Normal 'local' usage is treated just as it is now (including when using as root) except:
  * If the global directory is present, global info is included in output
[...]

Issue #1257, opened 2024-02-19
revert using platformdirs by default for windows and macos


Gitznik:
### Discussed in https://github.com/pypa/pipx/discussions/1247

**How would this feature be useful?**

Makes the default behavior on Windows and MacOS more robust, by fixing some edge cases with sandboxing in windows and spaces in shebangs on macos. For more details, please see the discussion.

**Describe the solution you'd like**

Default behavior on Windows and MacOS reverts to pre #1001, for Linux it stays as is.
Revert the current path fallback behavior for paths in the pre #1001 location to fallback to the `platformdirs` locations post #1001 instead on Windows and MacOS.

**Describe alternatives you've considered**

Sticking to the `platformdirs` locations, as described in the discussion. There was no support for that idea though.

Gitznik:
Should probably wait for https://github.com/pypa/pipx/pull/1281 to be merged as that will make this change a lot easier and prevent a lot of conflicts.

PR #1281, opened 2024-03-07
Updated PR: Added --global option to perform actions for all users (#754)

Jendker:

## Summary of changes

Adds a `--global` option to perform actions for all users. This is a resubmission of PR  https://github.com/pypa/pipx/pull/1088 by https://github.com/haxwithaxe which I rebased on the latest `main`.

## Test plan

Passes your CI process on github (via your unaltered actions in my fork) apart from MacOS test which needs to be investigated. Help from your side is highly appreciated. Test run from my fork: https://github.com/Jendker/pipx/actions/runs/8182365101

To verify the changes do what they are supposed to

```
python3 src/pipx --global environment
```

To test all the relevant commands for regressions.

```
python3 src/pipx --global ensurepath
sudo python3 src/pipx --global install pycowsay
sudo python3 src/pipx --global reinstall pycowsay
sudo python3 src/pipx --global inject pycowsay black
sudo python3 src/pipx --global uninject pycowsay black
sudo python3 src/pipx --global runpip pycowsay freeze
sudo python3 src/pipx --global list
sudo python3 src/pipx --global uninstall pycowsay
python3 src/pipx environment
python3 src/pipx ensurepath
python3 src/pipx install pycowsay
python3 src/pipx reinstall pycowsay
python3 src/pipx inject pycowsay black
python3 src/pipx uninject pycowsay black
python3 src/pipx runpip pycowsay freeze
python3 src/pipx list
python3 src/pipx uninstall pycowsay
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ _For comparison to other tools including pipsi, see
 ```
 brew install pipx
 pipx ensurepath
+sudo pipx ensurepath --global # optional to allow pipx actions with --global argument
 ```
 
 Upgrade pipx with `brew update && brew upgrade pipx`.
@@ -49,6 +50,7 @@ Upgrade pipx with `brew update && brew upgrade pipx`.
 sudo apt update
 sudo apt install pipx
 pipx ensurepath
+sudo pipx ensurepath --global # optional to allow pipx actions with --global argument
 ```
 
 - Fedora:
@@ -56,6 +58,7 @@ pipx ensurepath
 ```
 sudo dnf install pipx
 pipx ensurepath
+sudo pipx ensurepath --global # optional to allow pipx actions with --global argument
 ```
 
 - Using `pip` on other distributions:
@@ -63,6 +66,7 @@ pipx ensurepath
 ```
 python3 -m pip install --user pipx
 python3 -m pipx ensurepath
+sudo pipx ensurepath --global # optional to allow pipx actions with --global argument
 ```
 
 Upgrade pipx with `python3 -m pip install --user --upgrade pipx`.
--- a/docs/how-pipx-works.md
+++ b/docs/how-pipx-works.md
@@ -17,7 +17,7 @@ When installing a package and its binaries on linux (`pipx install package`) pip
 - on operating systems which have the `man` command, as long as `~/.local/share/man` is a recognized search path of man,
   you can now view the new manual pages globally
 - adding `--global` flag to any `pipx` command will execute the action in global scope which will expose app to all
-  users - [reference](installation.md#global-installation)
+  users - [reference](installation.md#global-installation). Note that this is not available on Windows.
 
 When running a binary (`pipx run BINARY`), pipx will
 
--- a/docs/how-pipx-works.md
+++ b/docs/how-pipx-works.md
@@ -16,6 +16,8 @@ When installing a package and its binaries on linux (`pipx install package`) pip
 - as long as `~/.local/bin/` is on your PATH, you can now invoke the new binaries globally
 - on operating systems which have the `man` command, as long as `~/.local/share/man` is a recognized search path of man,
   you can now view the new manual pages globally
+- adding `--global` flag to any `pipx` command will execute the action in global scope which will expose app to all
+  users - [reference](installation.md#global-installation)
 
 When running a binary (`pipx run BINARY`), pipx will
 
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -162,7 +162,7 @@ can be overridden with environment variable `PIPX_GLOBAL_MAN_DIR`. Finally, defa
 is `/opt/pipx`, can be overridden with environment variable `PIPX_GLOBAL_HOME`. Make sure to run `sudo pipx ensurepath --global`
 if you intend to use this feature.
 
-Note that `--global` argument is not supported on Windows.
+Note that the `--global` argument is not supported on Windows.
 
 ## Upgrade pipx
 
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -133,10 +133,12 @@ sudo PIPX_HOME=/opt/pipx PIPX_BIN_DIR=/usr/local/bin PIPX_MAN_DIR=/usr/local/sha
 # Example: $ sudo PIPX_HOME=/opt/pipx PIPX_BIN_DIR=/usr/local/bin PIPX_MAN_DIR=/usr/local/share/man pipx install cowsay
 ```
 
-Pipx also comes with a `--global` which helps to execute actions globally to give access to any user. By default the
-global binary location is set to `/usr/local/bin`, can be overridden with `PIPX_GLOBAL_BIN_DIR`. Default global manual
-page location is `/usr/local/share/man`. This can be overridden with `PIPX_GLOBAL_MAN_DIR`. Default global virtual
-environment location is `/opt/pipx`, can be overridden with `PIPX_GLOBAL_HOME`.
+### Global installation
+
+Pipx also comes with a `--global` argument which helps to execute actions globally to give app access to any user. By
+default the global binary location is set to `/usr/local/bin`, can be overridden with `PIPX_GLOBAL_BIN_DIR`. Default
+global manual page location is `/usr/local/share/man`. This can be overridden with `PIPX_GLOBAL_MAN_DIR`. Default global
+virtual environment location is `/opt/pipx`, can be overridden with `PIPX_GLOBAL_HOME`.
 
 > [!NOTE]
 >
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -133,6 +133,11 @@ sudo PIPX_HOME=/opt/pipx PIPX_BIN_DIR=/usr/local/bin PIPX_MAN_DIR=/usr/local/sha
 # Example: $ sudo PIPX_HOME=/opt/pipx PIPX_BIN_DIR=/usr/local/bin PIPX_MAN_DIR=/usr/local/share/man pipx install cowsay
 ```
 
+Pipx also comes with a `--global` which helps to execute actions globally to give access to any user. By default the
+global binary location is set to `/usr/local/bin`, can be overridden with `PIPX_GLOBAL_BIN_DIR`. Default global manual
+page location is `/usr/local/share/man`. This can be overridden with `PIPX_GLOBAL_MAN_DIR`. Default global virtual
+environment location is `/opt/pipx`, can be overridden with `PIPX_GLOBAL_HOME`.
+
 > [!NOTE]
 >
 > After version 1.2.0, the default pipx paths have been moved from `~/.local/pipx` to specific user data directories on
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -19,6 +19,7 @@ pipx works on macOS, linux, and Windows.
 ```
 brew install pipx
 pipx ensurepath
+sudo pipx ensurepath --global # optional to allow pipx actions in global scope. See "Global installation" section below.
 ```
 
 ### On Linux:
@@ -29,6 +30,7 @@ pipx ensurepath
 sudo apt update
 sudo apt install pipx
 pipx ensurepath
+sudo pipx ensurepath --global # optional to allow pipx actions in global scope. See "Global installation" section below.
 ```
 
 - Fedora:
@@ -36,6 +38,7 @@ pipx ensurepath
 ```
 sudo dnf install pipx
 pipx ensurepath
+sudo pipx ensurepath --global # optional to allow pipx actions in global scope. See "Global installation" section below.
 ```
 
 - Using `pip` on other distributions:
@@ -43,6 +46,7 @@ pipx ensurepath
 ```
 python3 -m pip install --user pipx
 python3 -m pipx ensurepath
+sudo pipx ensurepath --global # optional to allow pipx actions in global scope. See "Global installation" section below.
 ```
 
 
@@ -155,7 +159,8 @@ Pipx also comes with a `--global` argument which helps to execute actions in glo
 all system users. By default the global binary location is set to `/usr/local/bin` and can be overridden with the
 environment variable `PIPX_GLOBAL_BIN_DIR`. Default global manual page location is `/usr/local/share/man`. This
 can be overridden with environment variable `PIPX_GLOBAL_MAN_DIR`. Finally, default global virtual environment location
-is `/opt/pipx`, can be overridden with environment variable `PIPX_GLOBAL_HOME`.
+is `/opt/pipx`, can be overridden with environment variable `PIPX_GLOBAL_HOME`. Make sure to run `sudo pipx ensurepath --global`
+if you intend to use this feature.
 
 Note that `--global` argument is not supported on Windows.
 
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -133,13 +133,6 @@ sudo PIPX_HOME=/opt/pipx PIPX_BIN_DIR=/usr/local/bin PIPX_MAN_DIR=/usr/local/sha
 # Example: $ sudo PIPX_HOME=/opt/pipx PIPX_BIN_DIR=/usr/local/bin PIPX_MAN_DIR=/usr/local/share/man pipx install cowsay
 ```
 
-### Global installation
-
-Pipx also comes with a `--global` argument which helps to execute actions globally to give app access to any user. By
-default the global binary location is set to `/usr/local/bin`, can be overridden with `PIPX_GLOBAL_BIN_DIR`. Default
-global manual page location is `/usr/local/share/man`. This can be overridden with `PIPX_GLOBAL_MAN_DIR`. Default global
-virtual environment location is `/opt/pipx`, can be overridden with `PIPX_GLOBAL_HOME`.
-
 > [!NOTE]
 >
 > After version 1.2.0, the default pipx paths have been moved from `~/.local/pipx` to specific user data directories on
@@ -155,6 +148,16 @@ virtual environment location is `/opt/pipx`, can be overridden with `PIPX_GLOBAL
 >
 > `user_data_dir()`, `user_cache_dir()` and `user_log_dir()` resolve to appropriate platform-specific user data, cache and log directories.
 > See the [platformdirs documentation](https://platformdirs.readthedocs.io/en/latest/api.html#platforms) for details.
+
+### Global installation
+
+Pipx also comes with a `--global` argument which helps to execute actions in global scope which exposes the app to
+all system users. By default the global binary location is set to `/usr/local/bin` and can be overridden with the
+environment variable `PIPX_GLOBAL_BIN_DIR`. Default global manual page location is `/usr/local/share/man`. This
+can be overridden with environment variable `PIPX_GLOBAL_MAN_DIR`. Finally, default global virtual environment location
+is `/opt/pipx`, can be overridden with environment variable `PIPX_GLOBAL_HOME`.
+
+Note that `--global` argument is not supported on Windows.
 
 ## Upgrade pipx
 
--- a/src/pipx/commands/ensure_path.py
+++ b/src/pipx/commands/ensure_path.py
@@ -98,6 +98,15 @@ def ensure_path(location: Path, *, force: bool) -> Tuple[bool, bool]:
 def ensure_pipx_paths(force: bool) -> ExitCode:
     """Returns pipx exit code."""
     bin_paths = {paths.ctx.bin_dir}
+    if paths.ctx.is_global:
+        paths.ctx.make_local()
+        another_path = paths.ctx.bin_dir
+        paths.ctx.make_global()
+    else:
+        paths.ctx.make_global()
+        another_path = paths.ctx.bin_dir
+        paths.ctx.make_local()
+    bin_paths.add(another_path)
 
     pipx_user_bin_path = get_pipx_user_bin_path()
     if pipx_user_bin_path is not None:
--- a/src/pipx/commands/ensure_path.py
+++ b/src/pipx/commands/ensure_path.py
@@ -6,7 +6,7 @@ from typing import Optional, Tuple
 
 import userpath  # type: ignore
 
-from pipx import constants
+from pipx import paths
 from pipx.constants import EXIT_CODE_OK, ExitCode
 from pipx.emojis import hazard, stars
 from pipx.util import pipx_wrap
@@ -97,7 +97,7 @@ def ensure_path(location: Path, *, force: bool) -> Tuple[bool, bool]:
 
 def ensure_pipx_paths(force: bool) -> ExitCode:
     """Returns pipx exit code."""
-    bin_paths = {constants.PIPX_DIRS.BIN_DIR}
+    bin_paths = {paths.ctx.bin_dir}
 
     pipx_user_bin_path = get_pipx_user_bin_path()
     if pipx_user_bin_path is not None:
--- a/src/pipx/commands/ensure_path.py
+++ b/src/pipx/commands/ensure_path.py
@@ -98,15 +98,6 @@ def ensure_path(location: Path, *, force: bool) -> Tuple[bool, bool]:
 def ensure_pipx_paths(force: bool) -> ExitCode:
     """Returns pipx exit code."""
     bin_paths = {paths.ctx.bin_dir}
-    if paths.ctx.is_global:
-        paths.ctx.make_local()
-        another_path = paths.ctx.bin_dir
-        paths.ctx.make_global()
-    else:
-        paths.ctx.make_global()
-        another_path = paths.ctx.bin_dir
-        paths.ctx.make_local()
-    bin_paths.add(another_path)
 
     pipx_user_bin_path = get_pipx_user_bin_path()
     if pipx_user_bin_path is not None:
--- a/src/pipx/commands/ensure_path.py
+++ b/src/pipx/commands/ensure_path.py
@@ -97,7 +97,7 @@ def ensure_path(location: Path, *, force: bool) -> Tuple[bool, bool]:
 
 def ensure_pipx_paths(force: bool) -> ExitCode:
     """Returns pipx exit code."""
-    bin_paths = {constants.LOCAL_BIN_DIR}
+    bin_paths = {constants.PIPX_DIRS.BIN_DIR}
 
     pipx_user_bin_path = get_pipx_user_bin_path()
     if pipx_user_bin_path is not None:
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -934,7 +934,9 @@ def setup(args: argparse.Namespace) -> None:
 
     mkdir(paths.ctx.venvs)
     mkdir(paths.ctx.bin_dir)
+    mkdir(paths.ctx.man_dir)
     mkdir(paths.ctx.venv_cache)
+    mkdir(paths.ctx.standalone_python_cachedir)
 
     for cachedir in [
         paths.ctx.venv_cache,
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -19,8 +19,7 @@ import argcomplete
 import platformdirs
 from packaging.utils import canonicalize_name
 
-import pipx.constants
-from pipx import commands, constants
+from pipx import commands, constants, paths
 from pipx.animate import hide_cursor, show_cursor
 from pipx.colors import bold, green
 from pipx.constants import (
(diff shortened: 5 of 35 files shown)


Jendker:
Pinging @Gitznik who offered help in the original PR.

tests/test_uninstall.py
```diff
@@ -22,14 +28,21 @@ def test_uninstall(pipx_temp_env):
     assert not run_pipx_cli(["uninstall", "pycowsay"])
 
 
+def test_uninstall_global(pipx_temp_env):
+    if sys.platform.startswith("win"):
+        pytest.skip("This behavior is undefined on Windows")
```
chrysle, line 33: There are other tests that could be modified similarly.

```suggestion
@pytest.mark.skipif(sys.platform.startswith("win"), reason="This behavior is undefined on Windows")
def test_uninstall_global(pipx_temp_env):
```

tests/test_upgrade.py
```diff
@@ -10,7 +12,15 @@ def test_upgrade(pipx_temp_env, capsys):
     assert not run_pipx_cli(["upgrade", "pycowsay"])
 
 
-@pytest.mark.parametrize("metadata_version", PIPX_METADATA_LEGACY_VERSIONS)
```
chrysle, line 13: This change feels unrelated..

tests/test_upgrade.py
```diff
@@ -10,7 +12,15 @@ def test_upgrade(pipx_temp_env, capsys):
...
+def test_upgrade_global(pipx_temp_env, capsys):
+    if sys.platform.startswith("win"):
+        pytest.skip("This behavior is undefined on Windows")
+    assert run_pipx_cli(["--global", "upgrade", "pycowsay"])
+    assert not run_pipx_cli(["--global", "install", "pycowsay"])
+    assert not run_pipx_cli(["--global", "upgrade", "pycowsay"])
```
chrysle, line 20: IMHO we should check the outputs here.

tests/test_environment.py
```diff
@@ -49,3 +57,21 @@ def test_resolve_user_dir_in_env_paths_env_not_set(monkeypatch):
...
+    assert "PIPX_BIN_DIR=/usr/local/bin" in captured.out
+    assert "PIPX_MAN_DIR=/usr/local/share/man" in captured.out
+    assert "PIPX_SHARED_LIBS=/opt/pipx/shared" in captured.out
+    assert "PIPX_LOCAL_VENVS=/opt/pipx/venvs" in captured.out
+    assert "PIPX_LOG_DIR=/opt/pipx/logs" in captured.out
+    assert "PIPX_TRASH_DIR=/opt/pipx/.trash" in captured.out
+    assert "PIPX_VENV_CACHEDIR=/opt/pipx/.cache" in captured.out
+    # Checking just for the sake of completeness
+    assert "PIPX_DEFAULT_PYTHON" in captured.out
+    assert "USE_EMOJI" in captured.out
```
chrysle, line 77: I think it would be better to check for the whole output of the command here.

src/pipx/paths.py
```diff
@@ -0,0 +1,86 @@
...
+        return home.resolve()
+
+    @property
+    def shared_libs(self) -> Path:
+        return Path(self._base_shared_libs or self.home / "shared").resolve()
+
+    def make_global(self) -> None:
+        self._base_home = DEFAULT_PIPX_GLOBAL_HOME
+        self._base_bin = DEFAULT_PIPX_GLOBAL_BIN_DIR
+        self._base_man = DEFAULT_PIPX_GLOBAL_MAN_DIR
```
chrysle, line 79: How about enabling users to set the global paths via environment variables, too?

src/pipx/paths.py
```diff
@@ -0,0 +1,86 @@
...
+    return val
+
+
+class _PathContext:
+    _base_home: Optional[Union[Path, str]] = get_expanded_environ("PIPX_HOME")
+    _base_bin: Optional[Union[Path, str]] = get_expanded_environ("PIPX_BIN_DIR")
+    _base_man: Optional[Union[Path, str]] = get_expanded_environ("PIPX_MAN_DIR")
+    _base_shared_libs: Optional[Union[Path, str]] = get_expanded_environ("PIPX_SHARED_LIBS")
+    _fallback_home: Path = Path.home() / ".local/pipx"
+    _in_home: bool = _base_home is not None or _fallback_home.exists()
```
chrysle, line 29: I'd propose naming this variable differently, like `home_exists`.

src/pipx/paths.py
```diff
@@ -0,0 +1,86 @@
...
+    def logs(self) -> Path:
+        if self._in_home:
+            return self.home / "logs"
+        return user_log_path("pipx")
+
+    @property
+    def trash(self) -> Path:
+        if self._in_home:
+            return self.home / ".trash"
+        return self.home / "trash"
```
chrysle, line 46: Could you explain why this is special-cased for the default pipx home? Since the `venvs` path isn't.

```suggestion
        return self.home / "trash"
```

src/pipx/main.py
```diff
@@ -621,8 +627,16 @@ def _add_interpreter(
         description="Get help for commands with pipx interpreter COMMAND --help",
         dest="interpreter_command",
     )
-    s.add_parser("list", help="List available interpreters", description="List available interpreters")
-    s.add_parser("prune", help="Prune unused interpreters", description="Prune unused interpreters")
+    s.add_parser(
```
chrysle, line 630: Why this sudden style change?

chrysle (commented):
Thanks for picking this up! However we'll definitely need to add some more documentation, and extend the ["How pipx works"](https://pipx.pypa.io/stable/how-pipx-works/) documentation.

tests/test_list.py
```diff
@@ -27,6 +27,14 @@ def test_cli(pipx_temp_env, monkeypatch, capsys):
     assert "nothing has been installed with pipx" in captured.err
 
 
+def test_cli_global(pipx_temp_env, monkeypatch, capsys):
+    if sys.platform.startswith("win"):
+        pytest.skip("This behavior is undefined on Windows")
+    assert not run_pipx_cli(["--global", "list"])
+    captured = capsys.readouterr()
+    assert "nothing has been installed with pipx" in captured.err
```
Gitznik, line 35: Like this this test seems not very useful. I think installing a package without `--global` before running `list` with `--global` would make this much more useful.

tests/test_uninstall.py
Gitznik: Big + on this. Way easier to read. Could also turn this skip reason into a shareable decorator: https://docs.pytest.org/en/latest/how-to/skipping.html#id1

Gitznik (changes requested):
Thanks for picking this up!

src/pipx/main.py
Jendker: Reverted. Thank you for catching this!

src/pipx/paths.py
Jendker: That's just an implementation of how the things are working on main: https://github.com/pypa/pipx/blob/0cfcdc1781d6bc366e7f72af5b3c59513d9ec118/src/pipx/constants.py#L28-L36
In this PR it probably makes sense to stay consistent with the old logic and change it in another PR if needed.

tests/test_environment.py
Jendker: Do you mean that I should have a one long string to compare output? This test is consistent with the test above which I think is a good thing: https://github.com/pypa/pipx/blob/0cfcdc1781d6bc366e7f72af5b3c59513d9ec118/tests/test_environment.py#L7-L20

tests/test_upgrade.py
Jendker: Added checks both here and above in `test_upgrade` function

src/pipx/paths.py
Jendker: Good idea! Added. Will document accordingly.

tests/test_list.py
Jendker: Good idea! Added install without `--global`.

Jendker:
Thank you for the review and all helpful comments! I had a look into https://pipx.pypa.io/stable/how-pipx-works/ but I am not sure if I will be able to adjust the doc without blurring everything given my limited knowledge of pipx codebase. Some help here from your side would be highly appreciated!

For now, I have updated the docs in docs/installation.md which already mention the env variables so I could add concise information regarding `--global` flag there.

chrysle:
> I had a look into https://pipx.pypa.io/stable/how-pipx-works/ but I am not sure if I will be able to adjust the doc without blurring everything given my limited knowledge of pipx codebase. 

Probably adding a bullet point mentioning the `--global` flag and linking to the new paragraph under _Installation_ (which I think looks good) will be enough. This doc is also missing information on how to override the common paths, so providing the link should serve a dual purpose.

tests/helpers.py
```diff
@@ -6,10 +6,11 @@
 from typing import Any, Dict, List, Optional
 from unittest import mock
 
+import pytest
```
chrysle, line 9: This ought to satisfy `mypy`.

```suggestion
import pytest  # type: ignore
```

docs/installation.md
```diff
@@ -133,6 +133,13 @@ sudo PIPX_HOME=/opt/pipx PIPX_BIN_DIR=/usr/local/bin PIPX_MAN_DIR=/usr/local/sha
...
+### Global installation
+
+Pipx also comes with a `--global` argument which helps to execute actions globally to give app access to any user. By
+default the global binary location is set to `/usr/local/bin`, can be overridden with `PIPX_GLOBAL_BIN_DIR`. Default
+global manual page location is `/usr/local/share/man`. This can be overridden with `PIPX_GLOBAL_MAN_DIR`. Default global
+virtual environment location is `/opt/pipx`, can be overridden with `PIPX_GLOBAL_HOME`.
+
 > [!NOTE]
 >
 > After version 1.2.0, the default pipx paths have been moved from `~/.local/pipx` to specific user data directories on
```
Gitznik, line 145: IMO this note should be at line 135, so before the `Global Installation` header.

docs/installation.md
```diff
@@ -133,6 +133,13 @@ sudo PIPX_HOME=/opt/pipx PIPX_BIN_DIR=/usr/local/bin PIPX_MAN_DIR=/usr/local/sha
 # Example: $ sudo PIPX_HOME=/opt/pipx PIPX_BIN_DIR=/usr/local/bin PIPX_MAN_DIR=/usr/local/share/man pipx install cowsay
 ```
 
+### Global installation
+
```
Gitznik, line 137: Please add a note that this is not available on Windows.

docs/how-pipx-works.md
```diff
@@ -16,6 +16,8 @@ When installing a package and its binaries on linux (`pipx install package`) pip
 - as long as `~/.local/bin/` is on your PATH, you can now invoke the new binaries globally
 - on operating systems which have the `man` command, as long as `~/.local/share/man` is a recognized search path of man,
   you can now view the new manual pages globally
+- adding `--global` flag to any `pipx` command will execute the action in global scope which will expose app to all
```
Gitznik, line 19: Please also mention here that this is not available on windows.

docs/installation.md
```diff
@@ -133,6 +133,13 @@ sudo PIPX_HOME=/opt/pipx PIPX_BIN_DIR=/usr/local/bin PIPX_MAN_DIR=/usr/local/sha
...
+Pipx also comes with a `--global` argument which helps to execute actions globally to give app access to any user. By
+default the global binary location is set to `/usr/local/bin`, can be overridden with `PIPX_GLOBAL_BIN_DIR`. Default
+global manual page location is `/usr/local/share/man`. This can be overridden with `PIPX_GLOBAL_MAN_DIR`. Default global
+virtual environment location is `/opt/pipx`, can be overridden with `PIPX_GLOBAL_HOME`.
```
Gitznik, line 141: The previous paragraph specifies for each of the env variables that they are env variables. While quite verbose, I think it makes it easier to understand for a new user. If you agree, please adjust this new paragraph accordingly.

Gitznik (changes requested):
Just docs stuff, otherwise lgtm :+1:

docs/installation.md
Jendker: Yes, being new user-friendly is important. I adjusted the paragraph :)

docs/how-pipx-works.md
Jendker: Done

docs/installation.md
Jendker: Nice catch, thank you!

Jendker:
Is there anything I can do to get the tests to pass on MacOS? I am not sure why I get permission denied accessing `/opt/pipx` on MacOS but it works fine on Ubuntu.

Gitznik:
> Is there anything I can do to get the tests to pass on MacOS? I am not sure why I get permission denied accessing `/opt/pipx` on MacOS but it works fine on Ubuntu.

`/opt` does not seem to be a thing on MacOS. You can use [platformdirs](https://github.com/platformdirs/platformdirs) `site_data_dir` to generate the appropriate shared path instead I think. You can use it for both Linux and Mac, staying consistent with the current way of doing things (even though it may have to be changed during https://github.com/pypa/pipx/issues/1257).

chrysle:
> You can use [platformdirs](https://github.com/platformdirs/platformdirs) `site_data_dir` to generate the appropriate shared path instead I think. You can use it for both Linux and Mac, staying consistent with the current way of doing things (even though it may have to be changed during #1257).

That sounds good for MacOS, but I think we should stick to `/opt/` on Linux, as I am fairly sure `/usr/local/share/` is mostly a location for system administrators to install their own software.

Gitznik:
> That sounds good for MacOS, but I think we should stick to `/opt/` on Linux, as I am fairly sure `/usr/local/share/` is mostly a location for system administrators to install their own software.

There's a lot of discourse around it online (spend about 20 minutes yesterday reading conflicting opinions on that topic). From what I gathered either would be fine, really. BSD derivatives favor `usr/local/share` (as with MacOS, `/opt` is not really a thing there) while Linux itself slightly favors `/opt`. 

IMO - stick to the decision `platformdirs` has made, as we're not just supporting Linux - we're also supporting BSD derivatives.

chrysle:
> IMO - stick to the decision `platformdirs` has made, as we're not just supporting Linux - we're also supporting BSD derivatives.

Good, but if downstream packagers complain because they feel the need to patch this behaviour I'll direct them to you ;-)

Gitznik:
> Good, but if downstream packagers complain because they feel the need to patch this behaviour I'll direct them to you ;-)

If you think that's a realistic scenario, let's go for `/opt/` on linux, and `/usr/local/share/` for Mac & BSD? We currently don't differentiate between those anywhere AFAIK, so some new logic will be required there.

Jendker:
I tested it locally on my Mac machine and writing to /opt works both on Intel and ARM Mac but I have to run commands as a root user with `sudo`. I’m surprised that the tests are passing on Ubuntu without any `sudo` calls.

Jendker:
To follow up on my last comment I just run tests in with python docker image (debian based) with a non-root user and I got very similar output like in MacOS tests:
```
________________________________________________________________________________________ test_fetch_missing_python _________________________________________________________________________________________

src = PosixPath('/tmp/tmp9nfdvti7/download/python'), dst = PosixPath('/opt/pipx/py/3.12'), copy_function = <function copy2 at 0x7f454efcdc60>

    def move(src, dst, copy_function=copy2):
        """Recursively move a file or directory to another location. This is
        similar to the Unix "mv" command. Return the file or directory's
        destination.

        If dst is an existing directory or a symlink to a directory, then src is
        moved inside that directory. The destination path in that directory must
        not already exist.

        If dst already exists but is not a directory, it may be overwritten
        depending on os.rename() semantics.

        If the destination is on our current filesystem, then rename() is used.
        Otherwise, src is copied to the destination and then removed. Symlinks are
        recreated under the new name if os.rename() fails because of cross
        filesystem renames.

        The optional `copy_function` argument is a callable that will be used
        to copy the source or it will be delegated to `copytree`.
        By default, copy2() is used, but any function that supports the same
        signature (like copy()) can be used.

        A lot more could be done here...  A look at a mv.c shows a lot of
        the issues this implementation glosses over.

        """
        sys.audit("shutil.move", src, dst)
        real_dst = dst
        if os.path.isdir(dst):
            if _samefile(src, dst) and not os.path.islink(src):
                # We might be on a case insensitive filesystem,
                # perform the rename anyway.
                os.rename(src, dst)
                return

            # Using _basename instead of os.path.basename is important, as we must
            # ignore any trailing slash to avoid the basename returning ''
            real_dst = os.path.join(dst, _basename(src))

            if os.path.exists(real_dst):
                raise Error("Destination path '%s' already exists" % real_dst)
        try:
>           os.rename(src, real_dst)
E           FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmp9nfdvti7/download/python' -> '/opt/pipx/py/3.12'

/usr/local/lib/python3.12/shutil.py:886: FileNotFoundError

During handling of the above exception, another exception occurred:

monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7f454caa10d0>, mocked_github_api = None

    def test_fetch_missing_python(monkeypatch, mocked_github_api):
        def which(name):
            return None

        monkeypatch.setattr(shutil, "which", which)

        major = sys.version_info.major
        minor = sys.version_info.minor
        target_python = f"{major}.{minor}"

        if target_python == "3.8":
            # 3.8 is not available in the standalone python project
            with pytest.raises(InterpreterResolutionError) as e:
                find_python_interpreter(target_python, fetch_missing_python=True)
                assert "not found" in str(e)
        else:
>           python_path = find_python_interpreter(target_python, fetch_missing_python=True)

tests/test_interpreter.py:191:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
src/pipx/interpreter.py:67: in find_python_interpreter
    standalone_executable = download_python_build_standalone(python_version)
src/pipx/standalone_python.py:90: in download_python_build_standalone
    shutil.move(extracted_dir, install_dir)
/usr/local/lib/python3.12/shutil.py:902: in move
    copytree(src, real_dst, copy_function=copy_function,
/usr/local/lib/python3.12/shutil.py:600: in copytree
    return _copytree(entries=entries, src=src, dst=dst, symlinks=symlinks,
/usr/local/lib/python3.12/shutil.py:498: in _copytree
    os.makedirs(dst, exist_ok=dirs_exist_ok)
<frozen os>:215: in makedirs
    ???
<frozen os>:215: in makedirs
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

name = '/opt/pipx', mode = 511, exist_ok = False

>   ???
E   PermissionError: [Errno 13] Permission denied: '/opt/pipx'
```
The same tests passes with a root user.

I'm checking for the core issue but from what it looks like it's not the problem of `/opt` path existance on MacOS as many global tests are passing apart from this single test.

Jendker:
On a separate branch just to test what I said above I have added a call in the failing MacOS test to set paths to local (which resets what previous tests did) and now MacOS tests are passing.

Commit: https://github.com/Jendker/pipx/commit/969f0390df81197a99a871c32153a6355e79c65a
Test run: https://github.com/Jendker/pipx/actions/runs/8224676966

It doesn't solve the core issue with pytest, is only a workaround. I don't have enough expertise in pytest to understand why that's needed in the first place but if you think that this test workaround is good enough for now I'll add this commit to PR branch and I think we can merge.

Jendker:
I rebased this branch on the latest origin/main.

Jendker:
I just included the workaround commit for pytest issue on PR branch. The tests on MacOS will pass now. I think it would make sense to fix the pytest issue in another PR since the failing test shouldn't even be affected by the `--global` argument changes.

Gitznik:
> I just included the workaround commit for pytest issue on PR branch. The tests on MacOS will pass now. I think it would make sense to fix the pytest core issue in another PR since the failing test shouldn't even be affected by the `--global` argument changes.

I reckon the tests leak into the next running test. The way the test suite is written, globals in the application persist across runs. Maybe you can fix it by calling `pipx.paths.ctx.make_local()` at the end of all the tests that use the `--global` flag, instead of the tests where it fails? Just to make sure you always leave the globals in a clean state?

Jendker:
That's definitely better than what I had :) Done! Tests are passing: https://github.com/Jendker/pipx/actions/runs/8257157064

src/pipx/commands/ensure_path.py
```diff
@@ -97,7 +97,7 @@ def ensure_path(location: Path, *, force: bool) -> Tuple[bool, bool]:
 
 def ensure_pipx_paths(force: bool) -> ExitCode:
     """Returns pipx exit code."""
-    bin_paths = {constants.LOCAL_BIN_DIR}
```
Gitznik, line 100: We probably need to add the global binary location here.

Gitznik (changes requested):
> That's definitely better than what I had :) Done! Tests are passing: https://github.com/Jendker/pipx/actions/runs/8257157064

Great that that worked :+1: 

Maybe we should move it into the `pipx_temp_env_helper` function in `conftest` then, so one does not have to remember to add it to tests that make use of the `--global` flag? It would run for all tests that use a `temp_env`, but unless I'm missing something, that should not have any impact.

src/pipx/commands/ensure_path.py
Jendker: Updated. The code became a bit ugly there, let me know if you have any suggestions.

Jendker:
I have tried moving the `make_local` into `conftest` but that isn't enough. Failing test: https://github.com/Jendker/pipx/actions/runs/8267713069/job/22618840110
On `rebase_main` (PR branch) I have added the [make_local on fixture cleanup](https://github.com/pypa/pipx/pull/1281/commits/a918febccd99053b3ba73287b4301b38295ee211) commit while still keeping the `make_local` at the end of all `--global` tests and that obviously still works.

Do you have any suggestions how to improve the situation?

Gitznik:
> I have tried moving the `make_local` into `conftest` but that isn't enough. Failing test: https://github.com/Jendker/pipx/actions/runs/8267713069/job/22618840110 On `rebase_main` (PR branch) I have added the [make_local on fixture cleanup](https://github.com/pypa/pipx/pull/1281/commits/a918febccd99053b3ba73287b4301b38295ee211) commit while still keeping the `make_local` at the end of all `--global` tests and that obviously still works https://github.com/Jendker/pipx/actions/runs/8267933048.
> 
> Do you have any suggestions how to improve the situation?

`test_environment` `test_cli_global` does not use `pipx_temp_env`, so that will still leak into one other test. Maybe adding `pipx_temp_env` there will fix the tests?

src/pipx/commands/ensure_path.py
Gitznik: Reviewing this - maybe it's actually a good idea to only add the global location to the PATH if ensurepath is run with the `--global` flag, in which case no code change in the `ensurepath` command is necessary? That way you only need `sudo` to run `ensurepath`, if you want the global bins on your path.

I'd suggest adding that to the docs though. Sorry for sending you on that goose hunt.

Jendker:
That's a good point but that's still something the developer needs to add to avoid tests leaking so there is no substantial difference between adding `make_local` or `pipx_temp_env` to each test so it doesn't fix the core issue. Please correct me if I'm missing something.

Gitznik:
> That's a good point but that's still something the developer needs to add to avoid tests leaking so there is no substantial difference between adding `make_local` or `pipx_temp_env` to each test so it doesn't fix the core issue. Please correct me if I'm missing something.

Yes you're right. A lot of test have `pipx_temp_env` anyways, so having it there just allows you to not worry about `make_local` - most of the time. But yeah, having it in the tests that make use of the `--global` flag makes it very obvious that it needs to be done.

I'm fine with either version

Jendker:
I have removed all the explicit `make_local` calls in the tests and used `pipx_temp_env` in all tests with `--global`. I had to update one of the tests to pass now. Please let me know if you have any suggestions.

Gitznik (changes requested):
Please add a quick note to the `ensurepath` docs that you need to run it with `--global` to add the global location to the `PATH`. Aside from that - looks good to me know. Great work :)

tests/conftest.py
```diff
@@ -69,18 +69,22 @@ def pipx_temp_env_helper(pipx_shared_dir, tmp_path, monkeypatch, request, utils_
     bin_dir = Path(tmp_path) / "otherdir" / "pipxbindir"
     man_dir = Path(tmp_path) / "otherdir" / "pipxmandir"
 
-    monkeypatch.setattr(constants, "PIPX_SHARED_LIBS", pipx_shared_dir)
+    global_home_dir = Path(tmp_path) / "global" / "pipxhome"
+    global_bin_dir = Path(tmp_path) / "global_otherdir" / "pipxbindir"
+    global_man_dir = Path(tmp_path) / "global_otherdir" / "pipxmandir"
```
chrysle, line 74: This setup doesn't really test the behaviour in a realistic manner, IMHO. I wonder whether making use of [`pyfakefs`](https://pytest-pyfakefs.readthedocs.io/en/latest/) here would be worth the trouble?

Jendker: I am not sure why that's needed considering that the logic is still tested if the global directories instead of local directories are being used in tests? See e.g. 
https://github.com/pypa/pipx/blob/3e965b1d3c28fea1405d590e27f5095821a29a26/tests/test_environment.py#L60-L74
I can change `test_cli` function in this file to check if local paths are being used to contrast it with `test_cli_global` if that makes the things more robust.

Jendker:
> Please add a quick note to the `ensurepath` docs that you need to run it with `--global` to add the global location to the `PATH`. Aside from that - looks good to me know. Great work :)

Done in [fbbdd4c](https://github.com/pypa/pipx/pull/1281/commits/fbbdd4cebcc50a23d553945aafc628812395559a) :)

tests/conftest.py
Jendker: Done in [9597273](https://github.com/pypa/pipx/commit/959727300e42516968d1057cb374001e0be065a3)

README.md
```diff
@@ -49,20 +50,23 @@ Upgrade pipx with `brew update && brew upgrade pipx`.
...
 pipx ensurepath
+sudo pipx ensurepath --global # optional to allow pipx actions with --global argument
 ```
 
 - Using `pip` on other distributions:
 
 ```
 python3 -m pip install --user pipx
 python3 -m pipx ensurepath
+sudo pipx ensurepath --global # optional to allow pipx actions with --global argument
```
chrysle, line 69:
```suggestion
sudo pipx ensurepath --global # optional on Linux systems to allow pipx actions with --global argument
```

docs/installation.md
```diff
@@ -149,6 +153,17 @@ sudo PIPX_HOME=/opt/pipx PIPX_BIN_DIR=/usr/local/bin PIPX_MAN_DIR=/usr/local/sha
...
+### Global installation
+
+Pipx also comes with a `--global` argument which helps to execute actions in global scope which exposes the app to
+all system users. By default the global binary location is set to `/usr/local/bin` and can be overridden with the
+environment variable `PIPX_GLOBAL_BIN_DIR`. Default global manual page location is `/usr/local/share/man`. This
+can be overridden with environment variable `PIPX_GLOBAL_MAN_DIR`. Finally, default global virtual environment location
+is `/opt/pipx`, can be overridden with environment variable `PIPX_GLOBAL_HOME`. Make sure to run `sudo pipx ensurepath --global`
+if you intend to use this feature.
+
+Note that `--global` argument is not supported on Windows.
```
chrysle, line 165:
```suggestion
Note that the `--global` argument is not supported on Windows.
```

README.md
Gitznik: This is already under the `Linux` section, so I think this is redundant.

mairas:
I'm a bit late with the comment, but `/opt` is a thing on Macos, at least in practice. Specifically, Homebrew does install itself on `/opt/homebrew`. You need to create `/opt/pipx` with sudo and then set the permissions, but other than that, it works fine and I've already installed some packages there with the environment variable hacks described in Issue #754.

Homebrew used to install itself in `/usr/local` but switched the location some years ago. Don't remember the specifics but I think Macos wasn't happy with relaxing permisssions of `/usr/local` like Homebrew did. Risking a bit of cargo cult programming here, my preference would be to stick with `/opt/pipx`.

Gitznik (approved):

chrysle (approved):

Jendker:
If all looks good to you can this be merged? I am worried about the possible conflicts with other PR which may emerge if it doesn’t get merged into master some time soon ☺️

Gitznik:
Thanks for picking this up and following all the way through @Jendker :rocket:

gaby:
@Gitznik Is there an ETA for the next release? I can't wait to try this feature.

chrysle:
And until I've refurbished the release pipeline to do so, please feel free to install directly from the `main` branch. Pre-release testing is always welcome!

PR #1286, opened 2024-03-10
Added --extra-index-url and --use-input options to be available in pipx install.

search5:
Add --extra-index-url as PIPX install default option to use --extra-index-url=https://domain/simple instead of * --pip-args="--extra-index-url=https://domain/simple"
* Added --use-input option to temporarily disable the --no-input option when using PYPI from Google Artifact Registry, as the --no-input option was being enforced and preventing packages from being installed.

- [X ] I have added a news fragment under `changelog.d/` (if the patch affects the end users)

## Summary of changes

When using `pipx install`, specifying `--index-url` as the private repo and `https://pypi.org/simple` as the `--pip-args` option didn't seem to be very efficient, so we fixed it to accept `--extra-index-url` separately.

In addition, the `--no-input` option was enforced when installing packages from the google artifact registry, which prevented installation using the methods discussed in the discussion, so we added the `--use-input` option to prevent the `--no-input` option from being enforced only when this option is used.

## Test plan

```
pipx install --extra-index-url=https://asia-northeast3-python.pkg.dev/myproject/my-repo/simple --use-input --verbose my_package
```

Tested by running

```
pyenv exec pipx install \
  --extra-index-url=https://asia-northeast3-python.pkg.dev/myproject/my-repo/simple \
  --preinstall=keyrings.google-artifactregistry-auth \
  --use-input \
  my_package
⚠️  Note: my_package was already on your PATH at /Users/user_name/.pyenv/shims/gcloud_changer
  installed package my_package 0.1, installed using Python 3.10.6
    - my_command
done! ✨ 🌟 ✨
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,15 @@ This project uses [*towncrier*](https://towncrier.readthedocs.io/) for keeping t
 
 <!-- towncrier release notes start -->
 
+## [Unreleased] - 2024-03-10
+
+### Features
+
+- Add `--extra-index-url` options for the `pipx` subcommands
+- Add `--use-input` options for the `pipx install` subcommand:
+  The `--use-input` option was being forced by pipx install when installing from the Google Artifact Registry, causing the package to fail to install because `--no-input` was enforced by pipx install, so the `--use-input` option enabled the `--no-input` option to be the default but temporarily disabled. We covered this in https://github.com/pypa/pipx/discussions/1280#discussion-6329294.
+
+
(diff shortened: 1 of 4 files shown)


Gitznik:
I don't think we're quite ready for a PR here. I have some comments on the code for now but I don't think we actually need to change any code for this.

Gitznik (changes requested):

Gitznik:
Closing this PR, as the problem was resolved without a code change.

PR #1291, opened 2024-03-17
Introduce `pipx pin` and `pipx unpin` commands

### LINKED ISSUE #802: pipx doesn't respect manually selected point releases on `upgrade`
##  Describe the bug
In the following example, the system's headers and library of _gdal_ are in v3.2.2. Latest _gdal_ is on v3.3.2, so is _pygdal_ looking for this version, can't find it and breaks. Even when installed with `pygdal==3.2.2.*`, _pipx upgrade_ doesn't respect this limitation and the upgrade fails, since the installer can't find latest _gdal_:
```
_main__.GDALConfigError: Version mismatch 3.2.2 != 3.3.2
```

## How to reproduce
``` sh
$ apt-install libgdal-dev gdal-bin gdal-data # should be 3.2.2 on Debian
$ pipx install --include-deps pygdal==3.2.2.* # works
$ pipx upgrade pygdal # fails
```

It gets worse when _pygdal_ has been injected into another package. This completely fails then:
``` sh
$ pipx upgrade --include-injected --force mapproxy
```

## Expected behavior
Read `package_or_url` from pipx_metadata.json and limit it to the version stated in there.

## Workarounds
``` sh
pipx upgrade --pip-args 'install pygdal==3.2.2.*' pygdal
pipx upgrade --include-injected --pip-args 'install pygdal==3.2.2.*' mapproxy
```

## ~/.local/pipx/venvs/pygdal/pipx_metadata.json
``` json
{
    "injected_packages": {},
    "main_package": {
        "app_paths": [],
        "app_paths_of_dependencies": {
            "numpy": [
                {
                    "__Path__": "/home/tobwen/.local/pipx/venvs/pygdal/bin/f2py",
                    "__type__": "Path"
                },
                {
                    "__Path__": "/home/tobwen/.local/pipx/venvs/pygdal/bin/f2py3",
                    "__type__": "Path"
                },
                {
                    "__Path__": "/home/tobwen/.local/pipx/venvs/pygdal/bin/f2py3.9",
                    "__type__": "Path"
                }
            ]
        },
        "apps": [],
        "apps_of_dependencies": [
            "f2py",
            "f2py3",
            "f2py3.9"
        ],
        "include_apps": true,
        "include_dependencies": true,
        "package": "pygdal",
        "package_or_url": "pygdal==3.2.2.*",
        "package_version": "3.2.2.10",
        "pip_args": [],
        "suffix": ""
    },
    "pipx_metadata_version": "0.2",
    "python_version": "Python 3.9.2",
    "venv_args": []
}
```

### LINKED ISSUE #891: Pin the version of an installed package so that `pipx upgrade-all` doesn't inadvertently upgrade the version
Currently, it seems that the new `1.2.0` version of `poetry` has some pretty serious bugs. To avoid these, I have rolled back to the earlier `1.1.15` version. I use `pipx` to manage this, which works great, but now, if I use `pipx upgrade-all`, it will also upgrade `poetry` back to the newer version that I don't want to use. I use `pipx` for quite a few different utilities, so using `pipx upgrade-all` to keep everything up to date is a common part of my workflow.

It seems like an additional option for `pipx install` could be introduced (something like `--pin`), that would make it so the package will not be upgraded when running `pipx upgrade-all`. I'm not sure how "unpinning" would be handled, and it would probably also be important to have additional feedback that a package is pinned when running `pipx list`, for example.

Alternatively, maybe a new `pipx pin` subcommand could be added that would be used to manage "pinning/unpinning" specific packages.

Obviously, it is possible to just upgrade individual packages, or rollback `poetry` after it gets upgraded, but it is pretty annoying to have to do that.

dukecat0:

## Summary of changes
Introduce `pipx pin` and `pipx unpin` commands, which can be used to pin or unpin the version 
of an installed package, so it will not be upgraded by `pipx upgrade` or `pipx upgrade-all`.

Closes #891 
Closes #802
## Test plan

Tested by running

```
nox -s tests-3.12
```


src/pipx/commands/pin.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def pin(venv_dir: Path, verbose: bool, unpin: bool) -> ExitCode:
+    venv = Venv(venv_dir, verbose=verbose)
+    try:
+        package_metadata = venv.package_metadata[venv.main_package_name]
+        if package_metadata.pinned and not unpin:
+            logger.warning(f"{hazard} Package {package_metadata.package} already pinned")
+        elif not package_metadata.pinned and unpin:
+            logger.warning(f"{hazard} Package {package_metadata.package} not pinned")
```
chrysle, line 19: Nit: I'd rather use `emojis.sleep` here and log at `INFO` level, as this shouldn't be "hazardous" in any way?

src/pipx/commands/pin.py
```diff
@@ -0,0 +1,34 @@
...
+
+from pipx.constants import ExitCode
+from pipx.emojis import hazard
+from pipx.util import PipxError
+from pipx.venv import Venv
+
+logger = logging.getLogger(__name__)
+
+
+def pin(venv_dir: Path, verbose: bool, unpin: bool) -> ExitCode:
```
chrysle, line 12: To comply with the function name, we might use a default value of `False` for `unpin`?

chrysle (commented):
Thanks for submitting this! I wonder whether it would be a valid use case to extend this behaviour to work with injected packages, too?

Another idea: Provide protection from removal for a `pipx`-installed package by `pipx pin` via an option.

dukecat0:
>  I wonder whether it would be a valid use case to extend this behaviour to work with injected packages, too?

That sounds like a good idea! Probably I will add a new option for it.

> Another idea: Provide protection from removal for a pipx-installed package by pipx pin via an option.

Not sure if it's a common usage, but we can add it in the future if there are requests about this.

src/pipx/commands/pin.py
dukecat0: However if we log at `INFO` level, no information will be displayed to the user unless they pass `--verbose`.

chrysle: Ah, you're right. Forgot that we changed the default logging level.

dukecat0:
> I wonder whether it would be a valid use case to extend this behaviour to work with injected packages, too?

I am implementing the logic of `--injected-packages-only` and it will only pin injected packages only instead of main package. (As running `pipx pin <package>` should prevent both main package and injected packages from upgrading during `pipx upgrade-all --include-injected`.)

chrysle:
> I am implementing the logic of `--injected-packages-only` and it will only pin injected packages only instead of main package.

Sounds good. You don't think one should be able to select _which_ injected package to work with?

dukecat0:
Yes, forgot to mention it in the above comment. 😅 I will mark this PR as draft first.

src/pipx/commands/upgrade.py
```diff
@@ -28,6 +28,15 @@ def _upgrade_package(
...
         raise PipxError(f"Internal Error: package {package_name} has corrupt pipx metadata.")
+    elif package_metadata.pinned and package_metadata.package != venv.main_package_name:
+        logger.warning(
+            f"Not upgrading pinned package {package_metadata.package} in venv {venv.name}. "
+            f"Run `pipx unpin {venv.name}` to unpin it."
+        )
+        return 0
+    elif package_metadata.pinned:
+        logger.warning(f"Not upgrading pinned package {venv.name}. Run `pipx unpin {venv.name}` to unpin it.")
+        return 0
```
uranusjr, line 39: I wonder if this should return an error instead.

dukecat0: Returning 0 here means the version of package is not changed, and raising error may terminate the upgrade-all process.

justinmayer: From my perspective, returning an error here would defeat the purpose of this feature, which is to be able to run `pipx upgrade-all` to upgrade all packages _except_ pinned packages — without errors and with return code of `0`. In terms of prior art, Homebrew emits warnings in these cases:

```fish
~ ➤ brew upgrade --formula

Warning: Not upgrading 1 pinned package:
foo-package 1.7.2
```

src/pipx/commands/pin.py
```diff
@@ -0,0 +1,97 @@
...
+        raise PipxError("--skip must be used with --injected-only")
+    elif injected_only:
+        pinned_packages_count = 0
+        pinned_packages_list = []
+        for package_name in venv.package_metadata:
+            if package_name == venv.main_package_name or package_name in skip:
+                continue
+
+            if venv.package_metadata[package_name].pinned:
+                logger.warning(f"{package_name} was pinned. Not modifying.")
```
chrysle, line 53: Should this be a warning? The user can only try to pin all injected packages at once, right now, so he can't do anything about this. I understand the default logging level is warning; probably it could be printed instead?

src/pipx/commands/pin.py
```diff
@@ -0,0 +1,97 @@
...
+                continue
+
+            pinned_packages_count += _update_pin_info(venv, package_name, is_main_package=False, unpin=False)
+            pinned_packages_list.append(package_name)
+
+        if pinned_packages_count != 0:
+            print(bold(f"Pinned {pinned_packages_count} packages in venv {venv.name}"))
+            for package in pinned_packages_list:
+                print("  -", package)
```
chrysle, line 62: Should this print the version to which it was pinned, too?

src/pipx/main.py
```diff
@@ -579,6 +584,38 @@ def _add_uninject(subparsers, venv_completer: VenvCompleter, shared_parser: argp
...
+        "pin",
+        description="Pin the specified package to prevent it from being upgraded",
+        parents=[shared_parser],
+    )
+    p.add_argument("package", help="Installed package to pin")
+    p.add_argument(
+        "--injected-only",
+        action="store_true",
+        help=(
+            "Pin injected packages in main app only, so that they will not be upgraded during `pipx upgrade-all --include-injected`. "
```
chrysle, line 598:
```suggestion
            "Pin injected packages in venv only, so that they will not be upgraded during upgrade operations. "
```

src/pipx/commands/pin.py
```diff
@@ -0,0 +1,97 @@
...
+    venv = Venv(venv_dir, verbose=verbose)
+    try:
+        main_package_metadata = venv.package_metadata[venv.main_package_name]
+    except KeyError as e:
+        raise PipxError(f"Package {venv.name} is not installed") from e
+
+    if main_package_metadata.pinned:
+        logger.warning(f"Package {main_package_metadata.package} already pinned {sleep}")
+    elif skip and not injected_only:
+        raise PipxError("--skip must be used with --injected-only")
```
chrysle, line 44: Raise or "skip skip" and log a warning?

src/pipx/main.py
```diff
@@ -579,6 +584,38 @@ def _add_uninject(subparsers, venv_completer: VenvCompleter, shared_parser: argp
...
+        nargs="+",
+        default=[],
+        help="Skip these packages. Must be used with `--injected-only`.",
+    )
+
+
+def _add_unpin(subparsers, venv_completer: VenvCompleter, shared_parser: argparse.ArgumentParser) -> None:
+    p = subparsers.add_parser(
+        "unpin",
+        description="Unpin the specified package and all injected packages in its venv to allow them being upgraded",
```
chrysle, line 613:
```suggestion
        description="Unpin the specified package and all injected packages in its venv to allow them to be upgraded",
```

tests/test_pin.py
```diff
@@ -0,0 +1,65 @@
...
+
+def test_pin_warning(monkeypatch, capsys, pipx_temp_env, caplog):
+    assert not run_pipx_cli(["install", PKG["nox"]["spec"]])
+    assert not run_pipx_cli(["pin", "nox"])
+    assert not run_pipx_cli(["pin", "nox"])
+
+    assert "Package nox already pinned 😴" in caplog.text
+
+
+def test_pin_not_installed_package(monkeypatch, capsys, pipx_temp_env):
```
chrysle, line 29: This doesn't seem to be used in any of the tests. Accidentally copied?

```suggestion
def test_pin_not_installed_package(capsys, pipx_temp_env):
```

dukecat0: Yes 😅, thanks for reminding!

src/pipx/commands/pin.py
dukecat0: A warning sounds reasonable to me, since `--skip` actually implies `--injected-only`.

dukecat0:
The failed test seems unrelated to this PR.

huxuan:
> The failed test seems unrelated to this PR.

Rerun makes everything OK. :-)

src/pipx/commands/pin.py
jaklan: >since --skip actually implies --injected-only

Why though? What if I want to upgrade the core package + all dependencies except one?

dukecat0: I believe `pipx upgrade <venv> --skip <all packages except for the one you want to pin>` should work for this case.

jaklan: @dukecat0 yeah, but it doesn't seem to be convenient to list all the excluded packages... I believe it would be much simpler (and less surprising) if we don't imply `--injected-only` when using `--skip`, but allow people to use them together explicitly.

kpfleming:
This feature was included in the 1.6.0 release but the documentation appears to be incomplete. When I run `pipx --help` then 'pin' and 'unpin' are included in the command list but don't have descriptions:

```
subcommands:
  Get help for commands with pipx COMMAND --help

  {install,install-all,uninject,inject,pin,unpin,upgrade,upgrade-all,upgrade-shared,uninstall,uninstall-all,reinstall,reinstall-all,list,interpreter,run,runpip,ensurepath,environment,completions}
    install             Install a package
    install-all         Install all packages
    uninject            Uninstall injected packages from an existing Virtual Environment
    inject              Install packages into an existing Virtual Environment
    upgrade             Upgrade a package
    upgrade-all         Upgrade all packages. Runs `pip install -U <pkgname>` for each package.
    upgrade-shared      Upgrade shared libraries.
    uninstall           Uninstall a package
    uninstall-all       Uninstall all packages
    reinstall           Reinstall a package
    reinstall-all       Reinstall all packages
    list                List installed packages
    interpreter         Interact with interpreters managed by pipx
    run                 Download the latest version of a package to a temporary virtual environment, then run an app from it. Also compatible with local `__pypackages__` directory (experimental).
    runpip              Run pip in an existing pipx-managed Virtual Environment
    ensurepath          Ensure directories necessary for pipx operation are in your PATH environment variable.
    environment         Print a list of environment variables and paths used by pipx.
    completions         Print instructions on enabling shell completions for pipx
```

They also don't appear in the online docs at https://pipx.pypa.io/stable/docs/ except (again) in the list of subcommands, but without descriptions of their behavior or entries in the table of contents.

dukecat0:
This was fixed in #1438.

src/pipx/commands/pin.py
jaklan: @dukecat0 I have just tested the new `pipx` with `pin` released - and as I expected, implied `--injected-only` when using `--skip` makes it impossible to pin the main package and its dependencies except one, because if you use `--skip` - the main package won't be pinned. That's why I suggested to make these 2 options work separately.

jaklan: Sure (just random libraries to show the issue):
```
❯ pipx install black==24.3.0
❯ pipx inject black flake8 mypy
❯ pipx pin black --skip mypy
Pinned 1 packages in venv black
  - flake8 7.1.1
❯ pipx upgrade black
upgraded package black from 24.3.0 to 24.8.0 (location:
/Users/lanskij/.local/pipx/venvs/black)
```
And if I run `pipx pin black` - I would pin the main package, but also all the dependencies.

dukecat0: Sounds reasonable to me, I will open a PR to change this behavior. :)

jaklan: Thanks!

jaklan: Btw, that change would also allow to introduce `--include` (as opposed to `--skip`) in the future:
`❯ pipx pin black --include flake8`

dukecat0: Yes actually I was thinking about that as well, but I was not sure whether it would be redundant.

jaklan: Exactly, I found both `--skip` and `--include` flags usable, just for various use-cases

PR #1293, opened 2024-03-17
Fix using run with file descriptors

### LINKED ISSUE #1292: `pipx run` produces PermissionError when combined with bash process substitution
**Describe the bug**
Attempting to run a Python script through bash process substitution with pipx results in a PermissionError. The same script runs without issues when executed directly with Python. This indicates a potential issue with how pipx handles file descriptors or temporary files created by process substitution.

**How to reproduce**
```bash
$ pip install pipx-in-pipx
```
```bash
$ pipx run <(echo 'print("Hello World!")')
  File "/Users/bobronium/.local/bin/pipx", line 8, in <module>
    sys.exit(cli())
  File "/Users/bobronium/.local/pipx/venvs/pipx/lib/python3.10/site-packages/pipx/main.py", line 921, in cli
  File "/Users/bobronium/.local/pipx/venvs/pipx/lib/python3.10/site-packages/pipx/main.py", line 204, in run_pipx_command
    commands.run(
  File "/Users/bobronium/.local/pipx/venvs/pipx/lib/python3.10/site-packages/pipx/commands/run.py", line 205, in run
    run_package(
  File "/Users/bobronium/.local/pipx/venvs/pipx/lib/python3.10/site-packages/pipx/commands/run.py", line 141, in run_package
    run_pypackage_bin(pypackage_bin_path, app_args)
  File "/Users/bobronium/.local/pipx/venvs/pipx/lib/python3.10/site-packages/pipx/util.py", line 102, in run_pypackage_bin
    exec_app(
  File "/Users/bobronium/.local/pipx/venvs/pipx/lib/python3.10/site-packages/pipx/util.py", line 385, in exec_app
    os.execvpe(str(cmd[0]), [str(x) for x in cmd], env)
  File "/Users/bobronium/.pyenv/versions/3.10.1/lib/python3.10/os.py", line 583, in execvpe
    _execvpe(file, args, env)
  File "/Users/bobronium/.pyenv/versions/3.10.1/lib/python3.10/os.py", line 596, in _execvpe
    exec_func(file, *argrest)
PermissionError: [Errno 13] Permission denied: '/dev/fd/11'
```
**Expected behavior**

Pipx should execute the script provided through process substitution, printing "Hello World!" to the console, similar to the direct Python execution:
```bash
$ python <(echo 'print("Hello World!")')
Hello World!
```

**Context**
Supporting this behaviour would allow to run scripts in-place using a clipboard:
```py
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "requests<3",
#   "rich",
# ]
# ///

import requests
from rich.pretty import pprint

resp = requests.get("https://peps.python.org/api/peps.json")
data = resp.json()
pprint([(k, v["title"]) for k, v in data.items()][:10])
```

```bash
$ pipx run <(pbpaste)
```

Bobronium:

Fixes #1292

## Summary of changes
Remove `Path(app).is_file()` check and ask for forgiveness instead.

## Test plan

Tested by running
```bash
python -m pipx run <(echo 'print("Hello world!")')
```

### DIFF
--- a/src/pipx/commands/run.py
+++ b/src/pipx/commands/run.py
@@ -42,16 +42,18 @@ Available executable scripts:
     {app_lines}"""
 
 
-def maybe_script_content(app: str, is_path: bool) -> Optional[Union[str, Path]]:
+def maybe_script_content(app: str, is_path: bool) -> Optional[str]:
     # If the app is a script, return its content.
     # Return None if it should be treated as a package name.
+    # Look for a local file first
+    try:
+        return Path(app).read_text(encoding="utf-8")
+    except (FileNotFoundError, IsADirectoryError):
+        pass
 
-    # Look for a local file first.
-    app_path = Path(app)
-    if app_path.is_file():
-        return app_path
-    elif is_path:
+    if is_path:
         raise PipxError(f"The specified path {app} does not exist")
+
 
     # Check for a URL
     if urllib.parse.urlparse(app).scheme:
--- a/tests/test_run.py
+++ b/tests/test_run.py
@@ -354,6 +354,32 @@ def test_run_script_by_relative_name(caplog, pipx_temp_env, monkeypatch, tmp_pat
         run_pipx_cli_exit(["run", "test.py"])
     assert out.read_text() == test_str
 
+@mock.patch("os.execvpe", new=execvpe_mock)
+@pytest.mark.skipif(sys.platform.startswith("win"), reason="uses file descriptor")
+def test_run_script_by_file_descriptor(caplog, pipx_temp_env, monkeypatch, tmp_path):
+    read_fd, write_fd = os.pipe()
+    out = tmp_path / "output.txt"
+    test_str = "Hello, world!"
+
+    os.write(
+        write_fd,
+        textwrap.dedent(
+            f"""
+                from pathlib import Path
+                Path({repr(str(out))}).write_text({repr(test_str)})
+            """
+        ).strip().encode("utf-8")
+    )
+    os.close(write_fd)
+    
+    with monkeypatch.context() as m:
+        m.chdir(tmp_path)
+        try:
+            run_pipx_cli_exit(["run", f'/dev/fd/{read_fd}'])        
+        finally:
+            os.close(read_fd)
+    assert out.read_text() == test_str
+
 
 @pytest.mark.skipif(not sys.platform.startswith("win"), reason="uses windows version format")
 @mock.patch("os.execvpe", new=execvpe_mock)


Bobronium:
Hmm. I just noticed that `maybe_script_content` now might return Path unchanged on main as opposed to 1.4.3 where it always read out content.

Have to do things differently then. Checking additionally for `app_path.is_fifo()` doesn't work right away, need more debugging to figure out why.

src/pipx/commands/run.py
```diff
@@ -45,12 +45,15 @@
 def maybe_script_content(app: str, is_path: bool) -> Optional[Union[str, Path]]:
     # If the app is a script, return its content.
     # Return None if it should be treated as a package name.
-
-    # Look for a local file first.
+    # Look for a local file first
     app_path = Path(app)
```
Bobronium, line 49:
```suggestion

    # Look for a local file first
    app_path = Path(app)
```

src/pipx/commands/run.py
```diff
@@ -46,11 +46,15 @@ def maybe_script_content(app: str, is_path: bool) -> Optional[Union[str, Path]]:
     # If the app is a script, return its content.
     # Return None if it should be treated as a package name.
 
-    # Look for a local file first.
+    # Look for a local file first
```
Bobronium, line 49:
```suggestion
    # Look for a local file first.
```

changelog.d/1293.bugfix.md
chrysle, line 1: I'd rather classify this as a feature, actually.

Bobronium: I also thought of it as a feature at first, however, decided to view it as a bug since it diverges from vanilla python behaviour which run supposed to mimic when it comes to executing files. 

It also doesn't seem like a deliberate design choice or a documented limitation. 

Is there any particular reason you prefer to classify it as a feature?

chrysle: > Is there any particular reason you prefer to classify it as a feature?

Well, I personally have never thought of using `pipx run` like this and also am fairly sure it is not a very common use case?

Bobronium: You're right. I'm not sure using `pipx run script.py` is a common usecase to begin with.

Yet, classifying something as a bugfix or a feature shouldn't be just a question of how frequently it is used, won't you agree?

Nevertheless, a good use case for this would be asking LLM to generate a script and specify its dependencies via [PEP 723](https://peps.python.org/pep-0723/), then copying it and simply running `pipx run <(pbpaste)`. I suspect it might become slightly more common as well, considering the growing popularity of LLMs and ergonomics of PEP 723.

That said, I don't really have a strong opinion on that and willing change it to feature, either if it will become a requirement for merging, or if I'll understand the reason behind it.

chrysle: > Yet, classifying something as a bugfix or a feature shouldn't be just a question of how frequently it is used, won't you agree?

I tend to disagree on that; if a behaviour is only cared about by a minority of users, while most are not even aware that the util disallows this kind of usage, then I would describe adding support for that behaviour as implementing a feature. But I do not insist on this minor categorizing decision, either, so feel free to resolve this thread if this argument does not feel strong enough for you.

> I suspect it might become slightly more common as well, considering the growing popularity of LLMs and ergonomics of PEP 723.

Sounds like a valid use case to me.

chrysle:
Please add an example for this functionality to `docs/examples.md`.

gaborbernat:
This seems to have stalled, closing we can reopen if you pick it up again

PR #1296, opened 2024-03-18
Update Instructions for zsh in `pipx completions` Output

mjs271:

(Note: change does not affect _usage_ by users, so I haven't added to `changelog.d/`)

## Summary of changes

The output of `pipx completions` (`[constants.py](https://github.com/pypa/pipx/blob/master/src/pipx/constants.py#L39):39`) has been modified to reflect the current capability of `argcomplete` as of v3 ([changelog](https://github.com/kislyuk/argcomplete/blob/9077be97459abab129d02ad937eff1618cf67e57/Changes.rst#changes-for-v300-2023-03-19)), namely that [zsh is fully supported](https://stackoverflow.com/a/75791863/6031795), so there is no need for `bashcompinit`. However, `argcomplete` < v3 is accounted for by referencing the former workflow.

Additionally, I've added to the top of the output to indicate that one must **_first_** install `argcomplete`, if not already installed or if using zipapp. This is due to the fact that, in setting up completions for myself, I had to first install `argcomplete`, and am running zsh on macOS 14.4.

## Test plan

N/A

### DIFF
--- a/src/pipx/constants.py
+++ b/src/pipx/constants.py
@@ -58,19 +58,18 @@ zsh:
 
     autoload -U compinit && compinit
 
-    Afterwards you can enable completion for pipx:
+    Afterwards you can enable completions for pipx:
 
     eval "$(register-python-argcomplete pipx)"
 
-    NOTE: If your version of argcomplete is earlier than v3, you may need to
+    > [!NOTE]
+    >
+    > If your version of argcomplete is earlier than v3, you may need to
     have bashcompinit enabled in zsh by running:
+    >
+    > autoload -U bashcompinit
+    > bashcompinit
 
-    autoload -U bashcompinit
-    bashcompinit
-
-    and then following with:
-
-    eval "$(register-python-argcomplete pipx)"
 
 tcsh:
     eval `register-python-argcomplete --shell tcsh pipx`
--- a/src/pipx/constants.py
+++ b/src/pipx/constants.py
@@ -38,7 +38,7 @@ MINGW: bool = is_mingw()
 
 completion_instructions = dedent(
     """
-If you do not already have argcomplete installed,
+If you encountered register-python-argcomplete command not found error,
 or if you are using zipapp, run
 
     pipx install argcomplete
--- a/src/pipx/constants.py
+++ b/src/pipx/constants.py
@@ -38,8 +38,12 @@ MINGW: bool = is_mingw()
 
 completion_instructions = dedent(
     """
-If you are using zipapp, run `pipx install argcomplete` before
-running any of the following commands.
+If you do not already have argcomplete installed,
+or if you are using zipapp, run
+
+    pipx install argcomplete
+
+before running any of the following commands.
 
 Add the appropriate command to your shell's config file
 so that it is run on startup. You will likely have to restart
@@ -49,13 +53,22 @@ bash:
     eval "$(register-python-argcomplete pipx)"
 
 zsh:
-    To activate completions for zsh you need to have
-    bashcompinit enabled in zsh:
+    To activate completions in zsh, first make sure compinit is marked for
+    autoload and run autoload:
+
+    autoload -U compinit && compinit
+
+    Afterwards you can enable completion for pipx:
+
+    eval "$(register-python-argcomplete pipx)"
+
+    NOTE: If your version of argcomplete is earlier than v3, you may need to
+    have bashcompinit enabled in zsh by running:
 
     autoload -U bashcompinit
     bashcompinit
 
-    Afterwards you can enable completion for pipx:
+    and then following with:
 
     eval "$(register-python-argcomplete pipx)"
 
--- a/src/pipx/constants.py
+++ b/src/pipx/constants.py
@@ -64,7 +64,7 @@ zsh:
 
     NOTE: If your version of argcomplete is earlier than v3, you may need to
     have bashcompinit enabled in zsh by running:
-    
+
     autoload -U bashcompinit
     bashcompinit
 
--- a/src/pipx/constants.py
+++ b/src/pipx/constants.py
@@ -62,13 +62,11 @@ zsh:
 
     eval "$(register-python-argcomplete pipx)"
 
-    > [!NOTE]
-    >
-    > If your version of argcomplete is earlier than v3, you may need to
+    NOTE: If your version of argcomplete is earlier than v3, you may need to
     have bashcompinit enabled in zsh by running:
-    >
-    > autoload -U bashcompinit
-    > bashcompinit
+    
+    autoload -U bashcompinit
+    bashcompinit
 
 
 tcsh:
--- a/changelog.d/1296.misc.md
+++ b/changelog.d/1296.misc.md
@@ -1,0 +1,1 @@
+Update screen output provided by `pipx completions` for zsh/argcomplete v3


dukecat0 (changes requested):
Please add a changelog entry

src/pipx/constants.py
```diff
@@ -49,13 +53,22 @@ def is_mingw() -> bool:
...
+    NOTE: If your version of argcomplete is earlier than v3, you may need to
+    have bashcompinit enabled in zsh by running:
 
     autoload -U bashcompinit
     bashcompinit
 
-    Afterwards you can enable completion for pipx:
+    and then following with:
 
     eval "$(register-python-argcomplete pipx)"
```
chrysle, line 73: I don't think we have to repeat that?

```suggestion
```

src/pipx/constants.py
```diff
@@ -49,13 +53,22 @@ def is_mingw() -> bool:
...
 
 zsh:
-    To activate completions for zsh you need to have
-    bashcompinit enabled in zsh:
+    To activate completions in zsh, first make sure compinit is marked for
+    autoload and run autoload:
+
+    autoload -U compinit && compinit
+
+    Afterwards you can enable completion for pipx:
```
chrysle, line 61:
```suggestion
    Afterwards you can enable completions for pipx:
```

src/pipx/constants.py
```diff
@@ -49,13 +53,22 @@ def is_mingw() -> bool:
...
+
+    eval "$(register-python-argcomplete pipx)"
+
+    NOTE: If your version of argcomplete is earlier than v3, you may need to
+    have bashcompinit enabled in zsh by running:
 
     autoload -U bashcompinit
     bashcompinit
```
chrysle, line 69: This should look a bit better.

```suggestion
    > [!NOTE]
    >
    > If your version of argcomplete is earlier than v3, you may need to
    have bashcompinit enabled in zsh by running:
    >
    > autoload -U bashcompinit
    > bashcompinit
```

mjs271: Is that a special syntax or stylistic convention? I don't believe I'm familiar with it.

src/pipx/constants.py
```diff
@@ -49,16 +53,24 @@ def is_mingw() -> bool:
...
 
     eval "$(register-python-argcomplete pipx)"
 
+    > [!NOTE]
+    >
+    > If your version of argcomplete is earlier than v3, you may need to
+    have bashcompinit enabled in zsh by running:
+    >
+    > autoload -U bashcompinit
+    > bashcompinit
```
chrysle, line 71: Sorry, I must have thought this documentation was inside a markdown file, for a moment (well, for more than a moment, since creating GH suggestions with the right indentation is a pain)!

```suggestion
    NOTE: If your version of argcomplete is earlier than v3, you may need to
    have bashcompinit enabled in zsh by running:
    
    autoload -U bashcompinit
    bashcompinit
```

mjs271: Haha, no worries! I certainly had to correct some markdown notation when I was first writing it, too 😂

chrysle (dismissed):

Gitznik (dismissed):

dukecat0 (changes requested):

src/pipx/constants.py
```diff
@@ -38,8 +38,12 @@ def is_mingw() -> bool:
 
 completion_instructions = dedent(
     """
-If you are using zipapp, run `pipx install argcomplete` before
-running any of the following commands.
+If you do not already have argcomplete installed,
```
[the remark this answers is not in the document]
mjs271: Thanks for the heads-up on this. It led me down a rabbit hole that answered some questions for me...

What started me on this path and led to this PR is when I installed pipx via `brew install pipx`. I've recreated by deleting everything `brew uninstall -f --zap pipx` (and removing brew's originally-downloaded zip file). After running a new `brew install pipx`, I see the following
```shell
$ pipx environment

PIPX_HOME=
PIPX_BIN_DIR=
PIPX_MAN_DIR=
PIPX_SHARED_LIBS=
PIPX_DEFAULT_PYTHON=
USE_EMOJI=

PIPX_HOME=/Users/<uname>/Library/Application Support/pipx
PIPX_BIN_DIR=/Users/<uname>/.local/bin
PIPX_MAN_DIR=/Users/<uname>/.local/share/man
PIPX_SHARED_LIBS=/Users/<uname>/Library/Application Support/pipx/shared
PIPX_LOCAL_VENVS=/Users/<uname>/Library/Application Support/pipx/venvs
PIPX_LOG_DIR=/Users/<uname>/Library/Logs/pipx
PIPX_TRASH_DIR=/Users/<uname>/Library/Application Support/pipx/trash
PIPX_VENV_CACHEDIR=/Users/<uname>/Library/Caches/pipx
USE_EMOJI=true

$ pipx list
nothing has been installed with pipx

$ which register-python-argcomplete
register-python-argcomplete not found

$ pipx reinstall argcomplete
Nothing to reinstall for argcomplete

$ pipx install argcomplete
  installed package argcomplete 3.2.3, installed using Python 3.12.2
    - register-python-argcomplete
done!

$ pipx list
venvs are in /Users/<uname>/Library/Application Support/pipx/venvs
apps are exposed on your $PATH at /Users/<uname>/.local/bin
manual pages are exposed at /Users/<uname>/.local/share/man
   package argcomplete 3.2.3, installed using Python 3.12.2
    - register-python-argcomplete

$ register-python-argcomplete pipx
zsh: /Users/<uname>/.local/bin/register-python-argcomplete: bad interpreter: /Users/<uname>/Library/Application: no such file or directory
``` 

So, after all that... I think macOS's terrible convention of putting a space in `/Application Support` is to blame. Based on what I can tell from the pipx install workflow and brew's recipe file, it ***should*** work, but the fatal flaw is the shebang in `register-python-argcomplete` with the path containing the `/Application Support` with space. I see that this shebang issue is generally addressed in `troubleshooting.md`, but I certainly didn't find that on my first try 🫥

I also discovered that this _space in the shebang interpreter_ is a longstanding annoyance, and appreciated this ancient thread on the topic:

> The "#! interpreter argument" is what is known as the Berkeley exec
> hack.  As described in the first faq referenced above it was a simple
> way to enable script interpreters and doesn't handle non-traditional
> filenames.  This has been in existence for so many years (decades!)
> that this is unlikely to change now.
> 
> >         IMHO bash should support every valid unix path for shebang lines.
> 
> This is part of the exec(2) kernel system call interface and is not
> part of bash.  Bash doesn't have anything to do with the "#!" line.
> The # is a comment character and bash sees that line as a comment.
> 
> Bob
> 
> Patient asks: "Doctor, doctor, it hurts when I use spaces in filenames."
> Doctor responds:  "... then don't use spaces in filenames."

In the end, I don't have strong opinions on the way to address this, but I'm happy to add some verbiage here, or whatever else might be helpful. Though I do see a new PR reverting `platformdirs()`, so perhaps that's the choice right there.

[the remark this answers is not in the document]
mjs271: {PR in question is #1297}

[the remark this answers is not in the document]
dukecat0: Oops, thanks for the explanation! How about we change it to something like:
```suggestion
If you encountered register-python-argcomplete command not found error,
```

dukecat0 (approved):
Thanks!

PR #1297, opened 2024-03-22
Revert platformdirs on MacOS and Windows

### LINKED ISSUE #1257: revert using platformdirs by default for windows and macos

Gitznik:

## Summary of changes

On MacOS and Windows, we now use the old paths again, while using the post 1.2.0 paths as fallback.
On Linux - nothing changed.

Resolves #1257 

## Test plan

Tested by running

Existing test suite

### DIFF
--- a/changelog.d/1257.feature.md
+++ b/changelog.d/1257.feature.md
@@ -1,4 +1,4 @@
-Revert platform specific directories on MacOS and Windows
+Revert platform-specific directories on MacOS and Windows
 
 They were leading to a lot of issues with Windows sandboxing
 and spaces in shebangs on MacOS.
--- a/changelog.d/1257.feature.md
+++ b/changelog.d/1257.feature.md
@@ -1,0 +1,4 @@
+Revert platform specific directories on MacOS and Windows
+
+They were leading to a lot of issues with Windows sandboxing
+and spaces in shebangs on MacOS.
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -130,7 +130,8 @@ on Windows or `~/Library/Application Support/pipx` on MacOS exists, it will be u
 This can be overridden with the `PIPX_HOME` environment variable.
 
 In case one of these fallback locations exist, we recommend either manually moving the pipx files to the new default location
-(see the `Troubleshooting` section of the docs), or setting the `PIPX_HOME` environment variable.
+(see the `Troubleshooting` section of the docs), or setting the `PIPX_HOME` environment variable (discarding files existing in
+the fallback location).
 
 As an example, you can install global apps accessible by all users on your system with the following command (on MacOS,
 Linux, and Windows WSL):
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -129,6 +129,9 @@ and `~/.local/pipx` on Windows and MacOS. For compatibility reasons, if `~/.loca
 on Windows or `~/Library/Application Support/pipx` on MacOS exists, it will be used as the default location instead.
 This can be overridden with the `PIPX_HOME` environment variable.
 
+In case one of these fallback locations exist, we recommend either manually moving the pipx files to the new default location
+(see the `Troubleshooting` section of the docs), or setting the `PIPX_HOME` environment variable.
+
 As an example, you can install global apps accessible by all users on your system with the following command (on MacOS,
 Linux, and Windows WSL):
 
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -124,10 +124,10 @@ The default binary location for pipx-installed apps is `~/.local/bin`. This can 
 variable `PIPX_BIN_DIR`. The default manual page location for pipx-installed apps is `~/.local/share/man`. This can be
 overridden with the environment variable `PIPX_MAN_DIR`.
 
-pipx's default virtual environment location is typically `~/.local/share/pipx` on Linux/Unix,
-`%USERPROFILE%\AppData\Local\pipx` on Windows and `~/Library/Application Support/pipx` on macOS, and for compatibility
-reasons, if `~/.local/pipx` exists, it will be used as the default location instead. This can be overridden with the
-`PIPX_HOME` environment variable.
+pipx's default virtual environment location is typically `~/.local/share/pipx` on Linux/Unix
+and `~/.local/pipx` on Windows and MacOS. For compatibility reasons, if `~/.local/pipx` on Linux, `%USERPROFILE%\AppData\Local\pipx`
+on Windows or `~/Library/Application Support/pipx` on MacOS exists, it will be used as the default location instead.
+This can be overridden with the `PIPX_HOME` environment variable.
 
 As an example, you can install global apps accessible by all users on your system with the following command (on MacOS,
 Linux, and Windows WSL):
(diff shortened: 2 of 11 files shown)


docs/installation.md
```diff
@@ -124,10 +124,10 @@ The default binary location for pipx-installed apps is `~/.local/bin`. This can
...
 overridden with the environment variable `PIPX_MAN_DIR`.
 
-pipx's default virtual environment location is typically `~/.local/share/pipx` on Linux/Unix,
-`%USERPROFILE%\AppData\Local\pipx` on Windows and `~/Library/Application Support/pipx` on macOS, and for compatibility
-reasons, if `~/.local/pipx` exists, it will be used as the default location instead. This can be overridden with the
-`PIPX_HOME` environment variable.
+pipx's default virtual environment location is typically `~/.local/share/pipx` on Linux/Unix
+and `~/.local/pipx` on Windows and MacOS. For compatibility reasons, if `~/.local/pipx` on Linux, `%USERPROFILE%\AppData\Local\pipx`
+on Windows or `~/Library/Application Support/pipx` on MacOS exists, it will be used as the default location instead.
+This can be overridden with the `PIPX_HOME` environment variable.
```
dukecat0, line 130: How about we suggest users setting PIPX_HOME to `~/.local/pipx` on Windows and macOS to avoid further issues?

Gitznik: Updated it accordingly

changelog.d/1257.feature.md
```diff
@@ -0,0 +1,4 @@
+Revert platform specific directories on MacOS and Windows
```
dukecat0, line 1: Should this be considered as `bugfix` instead of `feature`?

Gitznik: As this significantly changes the installation behavior I opted for `feature`, for visibility in the changelog as well. But I don't feel strongly about this. Let me know what you think and I'll change it accordingly.

dukecat0: Sounds reasonable 👍

dukecat0 (dismissed):

src/pipx/constants.py
```diff
@@ -29,11 +29,21 @@ def is_windows() -> bool:
     return sys.platform == "win32"
 
 
+def is_macos() -> bool:
+    return sys.platform == "darwin"
+
+
+def is_linux() -> bool:
+    return not WINDOWS and not MACOS
```
chrysle, line 37: I do think we should make this more explicit.

```suggestion
    return sys.platform == "linux"
```

changelog.d/1257.feature.md
chrysle, line 1:
```suggestion
Revert platform-specific directories on MacOS and Windows
```

docs/installation.md
```diff
@@ -124,10 +124,13 @@ The default binary location for pipx-installed apps is `~/.local/bin`. This can
...
+
+In case one of these fallback locations exist, we recommend either manually moving the pipx files to the new default location
+(see the `Troubleshooting` section of the docs), or setting the `PIPX_HOME` environment variable.
```
chrysle, line 133:
```suggestion
(see the `Troubleshooting` section of the docs), or setting the `PIPX_HOME` environment variable (discarding files existing in the fallback location).
```

src/pipx/constants.py
Gitznik: This would not cover a number of cases, see the `sys.platform` docs: https://docs.python.org/3/library/sys.html#sys.platform

Maybe we can call it `is_unix` instead, and keep the logic as is?

chrysle: > Maybe we can call it `is_unix` instead, and keep the logic as is?

Yes, sounds good.

chrysle:
By the way, should we have `pipx` automatically remove the old location if the new one exists and is populated?

src/pipx/constants.py
jaraco: My preference would be to use `platform.system() == "Linux"`

jaraco:
IMHO - if platformdirs isn't providing any value on Windows or Mac, it should probably be removed and pipx should fallback to a consistent behavior across platforms.

Gitznik:
> IMHO - if platformdirs isn't providing any value on Windows or Mac, it should probably be removed and pipx should fallback to a consistent behavior across platforms.

The benefit is specifically on linux, adhering to the xdg spec. IMO that's a good reason to keep it. we need to keep the package anyways to be backwards compatible with anyone that used the default location since 1.2


> By the way, should we have `pipx` automatically remove the old location if the new one exists and is populated?

I don't think I'm fond of automatically removing it, but maybe we can print a hint that it can safely be deleted by the user?

chrysle:
> The benefit is specifically on linux, adhering to the xdg spec. IMO that's a good reason to keep it.

As especially downstream packagers are really pleased if it is being followed (see e.g. [Debian](https://wiki.debian.org/UpstreamGuide#User_home_directories)).

> I don't think I'm fond of automatically removing it, but maybe we can print a hint that it can safely be deleted by the user?

Let's do this at info level.

src/pipx/paths.py
```diff
@@ -77,17 +93,43 @@ def make_local(self) -> None:
...
     @property
     def standalone_python_cachedir(self) -> Path:
         return self.home / "py"
 
+    def log_warnings(self):
+        if " " in str(self.home):
+            logger.warning(
+                pipx_wrap(
+                    (
+                        ":hazard: Found a space in the home path. We heavily discourage this, due to "
```
chrysle, line 113: Let's use `emojis.hazard` here.

src/pipx/paths.py
```diff
@@ -25,8 +40,9 @@ class _PathContext:
     _base_bin: Optional[Union[Path, str]] = get_expanded_environ("PIPX_BIN_DIR")
     _base_man: Optional[Union[Path, str]] = get_expanded_environ("PIPX_MAN_DIR")
     _base_shared_libs: Optional[Union[Path, str]] = get_expanded_environ("PIPX_SHARED_LIBS")
-    _fallback_home: Path = Path.home() / ".local/pipx"
-    _home_exists: bool = _base_home is not None or _fallback_home.exists()
+    _fallback_homes: list[Path] = FALLBACK_PIPX_HOMES
```
chrysle, line 43: You'll need to use `typing.List`.

src/pipx/paths.py
```diff
@@ -77,17 +93,43 @@ def make_local(self) -> None:
...
+                    (
+                        ":hazard: Found a space in the home path. We heavily discourage this, due to "
+                        "multiple incompatibilities. Please check our docs for more information on this, "
+                        "as well as some pointers on how to migrate to a different home path."
+                    ),
+                    subsequent_indent=" " * 4,
+                )
+            )
+
+        if self._fallback_homes and DEFAULT_PIPX_HOME.exists():
```
chrysle, line 121: Could also be that `PIPX_HOME` has been set? We should check if `_PathContext.home` exists.

src/pipx/paths.py
```diff
@@ -77,17 +93,43 @@ def make_local(self) -> None:
...
+            logger.info(
+                pipx_wrap(
+                    (
+                        f"Both the default pipx home folder ({DEFAULT_PIPX_HOME}) and the fallback "
+                        f"pipx home folder ({self._fallback_homes}) exist. If you are done migrating from the"
```
chrysle, line 126:
```suggestion
                        f"pipx home folder(s) ({self._fallback_homes}) exist. If you are done migrating from the"
```

chrysle (approved):

PR #1301, opened 2024-03-26
Add "install-all" command to install packages according to spec metadata file

### LINKED ISSUE #687: Support provisioning of pipx-managed sites from the output of `pipx list --json`.

huxuan:

## Summary of changes

Fix #687 

## Test plan

Tested by running

```
pipx list --json > pipx.json
pipx install-all pipx.json
```

## Known issues

1. https://github.com/pypa/pipx/issues/1304
2. Can not move utility functions to `pipx.util` due to circular import problem. https://github.com/pypa/pipx/pull/1301#issuecomment-2020512320


changelog.d/687.feature.md
```diff
@@ -0,0 +1 @@
+Add "install-all" command to install packages according to spec metadata file.
```
chrysle, line 1:
```suggestion
Add `install-all` command to install packages according to spec metadata file.
```

src/pipx/commands/install.py
```diff
@@ -117,3 +119,88 @@ def install(
 
     # Any failure to install will raise PipxError, otherwise success
     return EXIT_CODE_OK
+
+
+def extract_venv_metadata(spec_metadata_file: Path) -> Iterator[PipxMetadata]:
+    """Extracts venv metadata from spec metadata file."""
```
chrysle, line 125:
```suggestion
    """Extract venv metadata from spec metadata file."""
```

src/pipx/commands/install.py
```diff
@@ -117,3 +119,88 @@ def install(
...
+            venv_dir = paths.ctx.venvs.joinpath(package_path_name)
+            venv_metadata = PipxMetadata(venv_dir, read=False)
+            venv_metadata.from_dict(venvs_metadata_dict[package_path_name]["metadata"])
+            yield venv_metadata
+
+
+def generate_package_spec(package_info: PackageInfo) -> str:
+    """Generate more precise package spec from package info."""
+    if not package_info.package_or_url:
+        raise PipxError(f"The package_or_url of {package_info.package} is empty.")
```
chrysle, line 150:
```suggestion
        raise PipxError(f"A package spec is not available for {package_info.package}")
```

src/pipx/commands/install.py
```diff
@@ -117,3 +119,88 @@ def install(
...
+    spec_metadata_file: Path,
+    local_bin_dir: Path,
+    local_man_dir: Path,
+    pip_args: List[str],
+    venv_args: List[str],
+    verbose: bool,
+    *,
+    force: bool,
+) -> ExitCode:
+    """Returns pipx exit code."""
```
chrysle, line 167:
```suggestion
    """Return pipx exit code."""
```

src/pipx/commands/install.py
```diff
@@ -117,3 +119,88 @@ def install(
...
+            None,
+            [generate_package_spec(main_package)],
+            local_bin_dir,
+            local_man_dir,
+            venv_metadata.python_version,
+            pip_args,
+            venv_args,
+            verbose,
+            force=force,
+            reinstall=False,
```
chrysle, line 185: What if we have a package already installed?

chrysle (commented):
Thank you for this piece of work, it will come in handy! Could we have an example in `docs/examples.md`?

src/pipx/commands/install.py
huxuan: If the user does not specify the `--force` flag, everything will be kept as it is. And when the user specify the `--force` flag, the package will be overwrite.

I specify the `reinstall` as `False` here, because it means to first uninstall the package and then install it again, that is not an ordinary scenario in `install-all` I think.

chrysle:
Could you move the two helper functions to `util.py`? We're trying to keep the command files as clear as possible.

huxuan:
Seems the "macos-latest"

> docs/examples.md


`from pipx.pipx_metadata_file import PackageInfo, PipxMetadata, _json_decoder_object_hook` will import `PipxError` before it is initialized which will lead to "circular import" problem. Any ideas?

Maybe we need a `pipx.model` for those entities? If so, I would suggest to do it in a separate pull request.

huxuan:
I find some bug for the `python` args, will request another review when fix that problem.

huxuan:
@chrysle All comments except for moving the utility functions should be done. PTAL :-)

chrysle (approved):

PR #1303, opened 2024-03-26
Add support for listing packages from pyproject.toml files

### LINKED ISSUE #1302: Support list packages in toml format.

Nowadays, `pyproject.toml` is becoming a standard configuration file for Python projects. It would be useful if we can list related Python applications in the corresponding Python project, especially when combining with features like `install-all` (#687). 

Add a new option `--pyproject` to `pipx list`, just like `--json` and `--short`, it will print the package spec in toml and user can just copy it into `pyproject.toml` for management. In the future, we can even support `pipx install-all` that will by default install packages listed in the `pyproject.toml`.

No alternative solutions to the best of my knowledge.

ChaseKnowlden:

## Summary of changes
Adds a new argument to copy as a toml file for pyproject.toml.

Fix #1302 

## Test plan

Tested by running

```
pipx list --help
pipx list --pyproject
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,7 +28,7 @@ repos:
     - "packaging>=20"
     - "platformdirs>=2.1"
     - "tomli; python_version < '3.11'"
-    - "toml"
+    - "types-toml"
 # Configuration for codespell is in pyproject.toml
 - repo: https://github.com/codespell-project/codespell
   rev: v2.2.6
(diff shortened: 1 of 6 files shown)


Gitznik:
Thanks for working on this. 

As we don't have anything in place to read a `pyproject.toml` file and make decisions based on a `pipx` block in there, I don't see a reason to support this way of listing our requirements that way at the moment.

On top, I think this would need more research on how to structure a pyproject output. The solution you propose does not cover injected packages, the python version to use, ... It also seems generally broken, creating a new pipx block for each of the installed packages, if there are multiple.

Therefore I'm -1 on merging this at this time. Until a way of reading from a `pyproject.toml` is merged, I don't think there's any reason to continue developing on this either.

PR #1306, opened 2024-03-27
bug: Removed --python flag dependency when --force is passed

### LINKED ISSUE #1304: `pipx install --force` without `--python` complains about the ignorance of `--python`
**Describe the bug**

When run `pipx install --force` without `--python`, `pipx` still complains about the ignorance of `--python`

**How to reproduce**

```shell
> pipx install pycowsay
> pipx install pycowsay --force
--python is ignored when --force is passed. If you want to reinstall pycowsay with <path/to/default/python>, run `pipx reinstall pycowsay
--python <path/to/default/python>` instead.
```

**Expected behavior**

No `--python` ignorance warning.

BTW, seems the bug is introduced by the default value of `--python` [1]. And there are at least two approaches to resolve it. One is to remove the default value and check whether there is any side effects. The other way is to change the condition when warning [2]. Please let me know if there is any preference. 

[1] https://github.com/pypa/pipx/blob/ac70dc0d2221bc6a78e39bef86f43b5c73ef03d7/src/pipx/main.py#L376
[2] https://github.com/pypa/pipx/blob/ac70dc0d2221bc6a78e39bef86f43b5c73ef03d7/src/pipx/commands/install.py#L60

SwarajBaral:
Fixes #1304

## Test plan

Tested by running

```
```


src/pipx/commands/install.py
```diff
@@ -57,7 +56,7 @@ def install(
         venv = Venv(venv_dir, python=python, verbose=verbose)
         venv.check_upgrade_shared_libs(pip_args=pip_args, verbose=verbose)
         if exists:
-            if not reinstall and force and python_flag_was_passed:
+            if not reinstall and force:
```
Gitznik, line 59: Would this now not always print the warning when `--force` is passed, even when `--python` is not passed by design? I understood that is the exact thing the issue was supposed to adress?

Gitznik (changes requested):
Thanks for having a look. I'm not convinces this actually adresses the issue though. Please see my other review comment.

src/pipx/commands/install.py
SwarajBaral: Sorry, I misunderstood the issue. How do you suggest I go about it ? This warning should only pop up if `--force` and `--python` are used together right ?

Gitznik: Yes, exactly.

My first hunch would be to change the logic for `python_flag_was_passed` to
```python
python_flag_was_passed = python != DEFAULT_PYTHON
```
This is only an approximation though. If the user explicitly provided the default python via the `--python` flag, this would not print the warning.

The other option, as described in the issue, is to not set `DEFAULT_PYTHON` as the default argument to the `--python` flag. That way, we can actually check, if the flag was set or not.

If it does not make the code a lot more complex, I'd lean towards option 2, as it's more accurate.

Gitznik: Yes that's the one.

You have to check all usages of `args.python` and see if it's replaced with `DEFAULT_PYTHON` somewhere down the line. Just setting it to `None` and running the test suite should already give you a decent idea if it might be a problem, as most of the test suite does not use the `--python` flag.

SwarajBaral: Tests fails at this point, after the change, as python is `None`
```python3
    m.update(python.encode())
```
Do you think we should have a check for `python` being `None` here ?

huxuan:
As explained in #1304, there are two approaches to fix the problem:

1. Remove the default value of python [1]. We need to verify whether there is any side effects caused by this change.
2. Change the condition for warning. From my assumption, we should check whether `python` arg is the same as `DEFAULT_PYTHON` instead of `python_flag_was_passed` (not just removing it), and warning if it is not.

I personally perfer the first approach since it is more straightforward. The default value makes it hard to differentiate whether the user actually sets the `python` arg. But the change could affect many places, so we need to ensure it does not break anything. Also the preference depends, any other comments are welcome. 

[1] https://github.com/pypa/pipx/blob/ac70dc0d2221bc6a78e39bef86f43b5c73ef03d7/src/pipx/main.py#L376

src/pipx/commands/install.py
huxuan: How about setting the default value to an empty string?

SwarajBaral: That causes `InterpreterResolutionError` in almost all tests.

```python3
DEBUG    pipx.main:main.py:207 Failed to resolve interpreter:
  File "/Users/swaraj/personal/pipx/src/pipx/main.py", line 204, in run_pipx_command
    interpreter = find_python_interpreter(args.python, fetch_missing_python=fetch_missing_python)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/swaraj/personal/pipx/src/pipx/interpreter.py", line 72, in find_python_interpreter
    raise InterpreterResolutionError(source="PATH", version=python_version)
pipx.interpreter.InterpreterResolutionError: No executable for the provided Python version '' found in PATH. Please make
sure the provided version is on your PATH or the file path is valid. The
provided version looks like a version for Python Launcher, but `py` was not
found on PATH.
```

huxuan: OK, maybe empty string is not a good option. Thanks for the trying.

huxuan: Then I think need to do something like `python = python or DEFAULT_PYTHON` [1] in the corresponding places.

https://github.com/pypa/pipx/blob/ac70dc0d2221bc6a78e39bef86f43b5c73ef03d7/src/pipx/commands/install.py#L38

Gitznik: > Then I think need to do something like `python = python or DEFAULT_PYTHON` [1] in the corresponding places.
> 
> https://github.com/pypa/pipx/blob/ac70dc0d2221bc6a78e39bef86f43b5c73ef03d7/src/pipx/commands/install.py#L38

Yes, this.

SwarajBaral: > [1] in the corresponding places.

Does this mean everywhere we are using args.python or just fixing it `run_pipx_command` will do (see the snippet below) ? Sorry if I misunderstood the comment.
```python3
interpreter = find_python_interpreter(args.python or DEFAULT_PYTHON, fetch_missing_python=fetch_missing_python)
```

Gitznik: > Does this mean everywhere we are using args.python or just fixing it `run_pipx_command` will do (see the snippet below) ? Sorry if I misunderstood the comment.
> 
> ```python
> interpreter = find_python_interpreter(args.python or DEFAULT_PYTHON, fetch_missing_python=fetch_missing_python)
> ```

Doing it here will be fine I think. You'll have to store to a variable whether or not `args.python` was set before this line though, so you can pass it to where it's needed for the check.

SwarajBaral: Oh OK, so I can do something like this then ?
```python3
        ...
        python_flag_was_passed = args.python or DEFAULT_PYTHON
        try:
            interpreter = find_python_interpreter(args.python, python_flag_was_passed, fetch_missing_python=fetch_missing_python)
        ...
```
And then change `find_python_interpreter` accordingly ?
```python3
def find_python_interpreter(python_version: str, python_flag_was_passed: bool = True, fetch_missing_python: bool = False) -> str:
    if not python_flag_was_passed:
        return DEFAULT_PYTHON
    ...
```

SwarajBaral: I think the point where this line is written, `args.python` will always be set (refer below). But if we make the above fix, it will let us set the default value to an empty string as it will pick up `DEFAULT_PYTHON` at this point.
```python3
if "python" in args and args.python is not None:
        fetch_missing_python = args.fetch_missing_python
        python_flag_was_passed = bool(args.python)
        try:
            interpreter = find_python_interpreter(args.python, python_flag_was_passed, fetch_missing_python=fetch_missing_python)
            args.python = interpreter
        except InterpreterResolutionError as e:
            logger.debug("Failed to resolve interpreter:", exc_info=True)
            print(
                pipx_wrap(
                    f"{hazard} {e}",
                    subsequent_indent=" " * 4,
                )
            )
            return EXIT_CODE_SPECIFIED_PYTHON_EXECUTABLE_NOT_FOUND
```

SwarajBaral: We also will need to change `find_python_interpreter` 's signature a bit. Something like this
```python3
def find_python_interpreter(python_version: str, python_flag_was_passed: bool = True, fetch_missing_python: bool = False) -> str:
    if not python_flag_was_passed:
        return DEFAULT_PYTHON
    ...
```

SwarajBaral: Let me know if this fix works

Gitznik: Why not just change it to?

```python
    if "python" in args:
        fetch_missing_python = args.fetch_missing_python
        python_flag_passed = bool(args.python)
        try:
            interpreter = find_python_interpreter(args.python or DEFAULT_PYTHON, fetch_missing_python=fetch_missing_python)
            args.python = interpreter
        except:
            ...
```

Gitznik: no, but we need to create it here as the information whether the flag was set or not will be lost afterwards, due to
```python
        try:
            interpreter = find_python_interpreter(args.python or DEFAULT_PYTHON, fetch_missing_python=fetch_missing_python)
            args.python = interpreter
```

SwarajBaral: Ah okay, makes sense. Made the changes

src/pipx/main.py
```diff
@@ -373,7 +374,7 @@ def add_include_dependencies(parser: argparse.ArgumentParser) -> None:
 def add_python_options(parser: argparse.ArgumentParser) -> None:
     parser.add_argument(
         "--python",
-        default=DEFAULT_PYTHON,
+        default="",
```
Gitznik, line 379: Why even set a default and not let it default to None (which is the default behavior of argparse)?

src/pipx/commands/install.py
```diff
@@ -33,6 +33,7 @@ def install(
     """Returns pipx exit code."""
     # package_spec is anything pip-installable, including package_name, vcs spec,
     #   zip file, or tar.gz file.
+    python_flag_was_passed = python != DEFAULT_PYTHON
```
Gitznik, line 36: This, as discussed, is not really accurate which is the reason we evaluate `python_flag_passed` in `pipx/main`. You should get the information from there.

Gitznik (changes requested):

src/pipx/commands/install.py
SwarajBaral: Missed this commit. Updated. Also moved the `python_flag_passed=bool(args.python)` out of the if condition. Just to make it simpler to pass it to various methods if required.
```python3
    python_flag_passed = bool(args.python)
    if "python" in args:
        fetch_missing_python = args.fetch_missing_python
        try:
            interpreter = find_python_interpreter(
                args.python or DEFAULT_PYTHON, fetch_missing_python=fetch_missing_python
            )
            args.python = interpreter
```

Gitznik (changes requested):

src/pipx/commands/install.py
```diff
@@ -29,11 +29,11 @@ def install(
     include_dependencies: bool,
     preinstall_packages: Optional[List[str]],
     suffix: str = "",
+    python_flag_passed: Optional[bool] = False
```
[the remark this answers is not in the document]
SwarajBaral: `install` is being called from a few other places as well. Should we evaluate `python_flag_was_passed` everywhere it is being invoked and pass it down ?

src/pipx/main.py
```diff
@@ -197,10 +197,10 @@ def run_pipx_command(args: argparse.Namespace, subparsers: Dict[str, argparse.Ar
         logger.info(f"Virtual Environment location is {venv_dir}")
     if "skip" in args:
         skip_list = [canonicalize_name(x) for x in args.skip]
-
+    
+    python_flag_passed = bool(args.python)
```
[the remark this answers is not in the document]
Gitznik: Edit - your solution is also fine.

src/pipx/commands/install.py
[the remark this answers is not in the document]
Gitznik: But even if `install` is being called without `python_flag_passed`, the default will set it to `False`, so the `Optional` type is unnecessary. `python_flag_passed` will always be a boolean, never `None`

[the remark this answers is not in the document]
Gitznik: And preferably yes, we should always pass the correct `python_flag_passed` value.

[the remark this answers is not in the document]
SwarajBaral: Done. Inferring the flag from the other place is just checking if `python` param was passed to the method or not. So I think `python_flag_passed=bool(python)` will work.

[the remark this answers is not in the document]
Gitznik: It will always be passed, due to these lines, so inferring it the way you do is not accurate:
https://github.com/pypa/pipx/blob/6d81069806745ce29b1e85d016463bbee206380f/src/pipx/main.py#L203-L210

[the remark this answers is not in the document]
SwarajBaral: Oh yeah, my bad. Then i guess we have to drill down the value of ‘python_was_passed’ to the point where install is used

Gitznik (changes requested):
Code changes look fine for me. Can you please add a test that verifies this behavior?

SwarajBaral:
> Code changes look fine to me. Can you please add a test that verifies this behavior?

@Gitznik Doesn't `test_passed_python_and_force_flag_warning` test this exact behaviour ? I am happy to write a test for this, just not sure what I should be testing here that is not covered already.

Gitznik:
> @Gitznik Doesn't `test_passed_python_and_force_flag_warning` test this exact behaviour ? I am happy to write a test for this, just not sure what I should be testing here that is not covered already.

That test didn't cover the bug you fixed, so we need a test that verifies that this bug can no longer happen.

SwarajBaral:
> That test didn't cover the bug you fixed, so we need a test that verifies that this bug can no longer happen.

Done. Also added changelog.md file

tests/test_install.py
```diff
@@ -361,3 +361,10 @@ def test_install_run_in_separate_directory(caplog, capsys, pipx_temp_env, monkey
     f.touch()
 
     install_packages(capsys, pipx_temp_env, caplog, ["pycowsay"], ["pycowsay"])
+
+
+def test_force_install_python_flag_warning(pipx_temp_env, capsys):
+    assert not run_pipx_cli(["install", "pycowsay"])
+    assert not run_pipx_cli(["install", "pycowsay", "--force"])
+    captured = capsys.readouterr()
+    assert "--python is ignored when --force is passed." not in captured.out
```
Gitznik, line 370: I don't think this needs it's own test case. Just changing the existing test like so should do the trick IMO. 

```python
def test_passed_python_and_force_flag_warning(pipx_temp_env, capsys):
    assert not run_pipx_cli(["install", "black"])
    assert not run_pipx_cli(["install", "--python", sys.executable, "--force", "black"])
    assert "--python is ignored when --force is passed." in captured.out

    assert not run_pipx_cli(["install", "black", "--force"])
    assert "--python is ignored when --force is passed." not in captured.out, "Should only print warning if both flags present"

    assert not run_pipx_cli(["install", "pycowsay", "--force"])
    assert "--python is ignored when --force is passed." not in captured.out, "Should not print warning if package does not exist yet"
```

Gitznik (changes requested):

changelog.d/1304.bugfix.md
```diff
@@ -0,0 +1 @@
+Do not show `--python` flag warning when using `--force`
```
Gitznik, line 1:
```suggestion
Only show `--python` and `--force` flag warning if both flags are present
```

Gitznik (changes requested):

tests/test_install.py
SwarajBaral: Modified the test case and removed `test_force_install_python_flag_warning`

Gitznik (approved):

Koppom94:
> 
> 
> Fixes #1304
> 
> 
> 
> - [ ] I have added a news fragment under `changelog.d/` (if the patch affects the end users)
> 
> 
> 
> ## Summary of changes
> 
> 
> 
> ## Test plan
> 
> 
> 
> 
> 
> 
> 
> Tested by running
> 
> 
> 
> ```
> 
> # command(s) to exercise these changes
> 
> ```
> 
>

PR #1310, opened 2024-03-29
Release 1.5.0

Gitznik:

## Summary of changes

Let's have this manual release until we have automated pipelines again?

Will create a release if this PR is accepted and merged.

Note: I had to fix our feature news fragments so they render correctly in the changelog. We should all be aware that his is the required format, so we can enforce it int PRs. Please look at the first commit for the necessary changes.

## Test plan

Tested by running

```
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,50 @@ This project uses [*towncrier*](https://towncrier.readthedocs.io/) for keeping t
 
 <!-- towncrier release notes start -->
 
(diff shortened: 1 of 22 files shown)


chrysle (approved):
Yes, it's due. Sorry for the lack of activity, I'll try to get back to this asap.

Gitznik:
> Yes, it's due. Sorry for the lack of activity, I'll try to get back to this asap.

No bother. Not like this way of releasing is that much work either ;)

PR #1316, opened 2024-04-01
feat: add upgrade-shared command

mayeut:

## Summary of changes

Add upgrade-shared command

This allows to create/upgrade shared libraries as a standalone command.

It might be useful to initialize shared libraries with a pinned version of pip in order to create e.g. reproducible docker images using a constraint file.

It also allows for different `--pip-args` that one would get with the `install` command.

## Test plan

Tested in tests/test_upgrade_shared.py


chrysle:
Thank you for doing this! Could you extend our documentation at _how-pipx-works.md_, stating clearly how the shared libs can be upgraded?

mayeut:
@chrysle, I'm not sure what to add to `how-pipx-works.md`. I added an example though.

chrysle:
> I'm not sure what to add to `how-pipx-works.md`. 

I thought of adding something like "This [shared lib] can later on be managed with the [`upgrade-shared`](anchor) command" at the bullet point introducing the shared libs concept. Might be unnecessary.

> I added an example though.

Looks good, thanks! We somehow ignored it for a while, but there is a certain `scripts/gen_doc_pages.py` script that takes care of filling the template for our command line documentation (https://pipx.pypa.io/stable/docs/). Could you amend that? And, please split the commit updating the test packages into another PR, since it feels unrelated.

mayeut:
> Could you amend that? 

Done

> And, please split the commit updating the test packages into another PR, since it feels unrelated.

It's related but I limited the update to what's necessary only (pip).

docs/examples.md
```diff
@@ -142,3 +142,17 @@ Installing to existing venv 'pipx'
...
+One use of the upgrade-shared command is to force a `pip` upgrade.
+
+```shell
+> pipx upgrade-shared
+```
+
+Another use of the upgrade-shared command is to pin (temporarily) `pip` to a specific version.
+
+```shell
+> pipx upgrade-shared --pip-args=pip==24.0
```
chrysle, line 157:
```suggestion
> pipx upgrade-shared --pip-args="pip == 24.0"
```

I think this improves readability.

docs/examples.md
```diff
@@ -142,3 +142,17 @@ Installing to existing venv 'pipx'
...
+
+## `pipx upgrade-shared` examples
+
+One use of the upgrade-shared command is to force a `pip` upgrade.
+
+```shell
+> pipx upgrade-shared
+```
+
+Another use of the upgrade-shared command is to pin (temporarily) `pip` to a specific version.
```
chrysle, line 154:
```suggestion
This example pins `pip` (temporarily, until the next automatic upgrade, if that is not explicitly turned off) to a specific version.
```

src/pipx/commands/upgrade.py
```diff
@@ -252,3 +252,15 @@ def upgrade_all(
         )
 
     return EXIT_CODE_OK
+
+
+def upgrade_shared(
+    verbose: bool,
+    pip_args: List[str],
+) -> ExitCode:
+    """Returns pipx exit code."""
```
chrysle, line 261:
```suggestion
    """Return pipx exit code."""
```

tests/test_upgrade_shared.py
```diff
@@ -0,0 +1,62 @@
+import subprocess
+
+from helpers import run_pipx_cli
+
+
+def test_upgrade_shared(pipx_ultra_temp_env, capsys, caplog, monkeypatch):
+    from pipx.shared_libs import shared_libs
```
chrysle, line 7: Should this be moved to the top?

tests/test_upgrade_shared.py
```diff
@@ -0,0 +1,62 @@
...
+    assert "Upgrading shared libraries in" in caplog.text
+    assert "Already upgraded libraries in" not in caplog.text
+    assert shared_libs.has_been_updated_this_run is False
+    assert shared_libs.is_valid is True
+
+
+def test_upgrade_shared_pin_pip(pipx_ultra_temp_env, capsys, caplog, monkeypatch):
+    from pipx.shared_libs import shared_libs
+
+    def pip_version():
```
chrysle, line 49: We might make a test fixture out of this?

tests/test_upgrade_shared.py
```diff
@@ -0,0 +1,62 @@
+import subprocess
+
+from helpers import run_pipx_cli
+
+
+def test_upgrade_shared(pipx_ultra_temp_env, capsys, caplog, monkeypatch):
```
chrysle, line 6:
```suggestion
def test_upgrade_shared(pipx_ultra_temp_env, capsys, caplog):
```

This doesn't seem to be used in the test?

tests/test_upgrade_shared.py
```diff
@@ -0,0 +1,62 @@
...
+    assert "creating shared libraries" not in captured.err
+    assert "upgrading shared libraries" in captured.err
+    assert "Upgrading shared libraries in" in caplog.text
+    assert "Already upgraded libraries in" not in caplog.text
+    assert shared_libs.has_been_updated_this_run is True
+    assert run_pipx_cli(["upgrade-shared", "-v"]) == 0
+    assert "Already upgraded libraries in" in caplog.text
+
+
+def test_upgrade_shared_pip_args(pipx_ultra_temp_env, capsys, caplog, monkeypatch):
```
chrysle, line 31:
```suggestion
def test_upgrade_shared_pip_args(pipx_ultra_temp_env, capsys, caplog):
```

tests/test_upgrade_shared.py
```diff
@@ -0,0 +1,62 @@
...
+    captured = capsys.readouterr()
+    assert "creating shared libraries" in captured.err
+    assert "upgrading shared libraries" in captured.err
+    assert "Upgrading shared libraries in" in caplog.text
+    assert "Already upgraded libraries in" not in caplog.text
+    assert shared_libs.has_been_updated_this_run is False
+    assert shared_libs.is_valid is True
+
+
+def test_upgrade_shared_pin_pip(pipx_ultra_temp_env, capsys, caplog, monkeypatch):
```
chrysle, line 46:
```suggestion
def test_upgrade_shared_pin_pip(pipx_ultra_temp_env, capsys, caplog):
```

tests/test_upgrade_shared.py
mayeut: I will add a comment in order to explain why it's not at the top.

changelog.d/1316.feature.md
```diff
@@ -0,0 +1,7 @@
+Add `pipx upgrade-shared` command.
+
+This allows to create/upgrade shared libraries as a standalone command.
+
+It might be useful to initialize shared libraries with a pinned version of pip in order to create e.g. reproducible docker images using a constraint file.
+
+It also allows for different `--pip-args` that one would get with the `install` command.
```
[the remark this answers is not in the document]
mayeut: The pre-commit hook just reverted the change.

docs/examples.md
mayeut: This fails on windows, c.f. https://github.com/mayeut/pipx/actions/runs/8771336857/job/24068962865

tests/test_upgrade_shared.py
mayeut: I don't see how but feel free to push on the branch.

docs/examples.md
chrysle: It should work with single quotes? Updated the suggestion.

tests/test_upgrade_shared.py
chrysle: Ah, I see. This isn't feasible because the Python interpreter for the shared lib doesn't exist at the time the fixture is called. I thought of creating a general way to retrieve the `pip` version based on a Python interpreter path.

tests/test_upgrade_shared.py
chrysle: I see, thank you.

docs/examples.md
mayeut: It also fails on windows: https://github.com/pypa/pipx/actions/runs/8779377197/job/24087325568

chrysle (approved):

mayeut:
just fixed conflicts with the main branch

chrysle (approved):

chrysle:
Thanks, @mayeut!

PR #1329, opened 2024-04-07
Fix `pipx reinstall` crash with absolute path

### LINKED ISSUE #1324: `reinstall` is destructive when misused with local source directory
pipx version 1.5.0

**Describe the bug**

I have a package installed with pipx, and I wanted to reinstall it from a local source repository. I ran 

```bash
pipx reinstall {absolute-path-to-repo}
```

thinking that it would do something equivalent to 

```bash
pipx install {absolute-path-to-repo} --force
```

Instead, it deleted the file tree at `{absolute-path-to-repo}`. 

From this experience, I have inferred that I have misunderstood the intended functionality and arguments of the `reinstall` command. (I'm inferring that reinstall accepts a path to a package's pipx virtual environment, rather than a path to package source.) However, I feel like this is an easy mistake to make, and the destructive side effect of this performing this action seems really bad. I did in fact lose some work that was only in my local git repository. 

**How to reproduce**

1. Create any directory
    ```bash
    mkdir testdir
    ```
2. Run `pipx reinstall` with an absolute path to that directory
    ```bash
    pipx reinstall $(pwd)/testdir
    ```
    This produces output
    ```
    uninstalled testdir! ✨ 🌟 ✨
        /Users/jqi/.local/pipx/logs/cmd_2024-04-05_19.03.13_pip_errors.log
 
        ERROR: Could not find a version that satisfies the requirement testdir (from versions: none)
        ERROR: No matching distribution found for testdir
    
    Error installing testdir.
    ```
    with the side effect of deleting `testdir/`.

**Expected behavior**

`pipx reinstall` should not accept a local file path as an argument to prevent misuse. Or if this is actually desired functionality, there should more checks in place to ensure that it is intentional and actually a path to a pipx virtual environment.

chrysle:

## Summary of changes

Closes #1324. This checks whether the argument given as package name inside `VenvContainer.get_venv_dir()` is an absolute path or existing path, raising `PipxError` if so. This is due to the behaviour that an absolute path passed into the `pathlib.PurePath.joinpath()` function to join with another path will return the same. 

## Test plan

Tested by running

```
pipx reinstall /some/absolute/path
mkdir existing_path ; pipx reinstall existing_path
```

### DIFF
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -202,7 +202,8 @@ def run_pipx_command(args: argparse.Namespace, subparsers: Dict[str, argparse.Ar
                     pipx_wrap(
                         f"""
                         Error: Path '{package}' given as
-                        package name.
+                        package. Expected the name of
+                        an installed package.
                         """
                     )
                 )
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -185,7 +185,8 @@ def run_pipx_command(args: argparse.Namespace, subparsers: Dict[str, argparse.Ar
 
     if "package" in args:
         package = args.package
-        if urllib.parse.urlparse(package).scheme:
+        url_parse_package = urllib.parse.urlparse(package)
+        if url_parse_package.scheme and url_parse_package.netloc:
             raise PipxError("Package cannot be a url")
 
         if "spec" in args and args.spec is not None:
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -193,6 +193,19 @@ def run_pipx_command(args: argparse.Namespace, subparsers: Dict[str, argparse.Ar
                 if "#egg=" not in args.spec:
                     args.spec = args.spec + f"#egg={package}"
 
+        if args.command == "reinstall":
+            # Passing paths into `reinstall` might have unintended
+            # side effects.
+            if Path(package).is_absolute() or Path(package).exists():
+                raise PipxError(
+                    pipx_wrap(
+                        f"""
+                        Error: Path '{package}' given as
+                        package name.
+                        """
+                    )
+                )
+
         venv_dir = venv_container.get_venv_dir(package)
         logger.info(f"Virtual Environment location is {venv_dir}")
 
--- a/tests/test_reinstall.py
+++ b/tests/test_reinstall.py
@@ -62,9 +62,9 @@ def test_reinstall_with_path(pipx_temp_env, capsys, tmp_path):
     assert run_pipx_cli(["reinstall", str(path)])
     captured = capsys.readouterr()
 
-    assert f"Path '{str(path)}' given as package" in captured.err.replace("\n", " ")
+    assert "Expected the name of an installed package" in captured.err.replace("\n", " ")
 
     assert run_pipx_cli(["reinstall", str(path.resolve())])
     captured = capsys.readouterr()
 
-    assert f"Path '{str(path.resolve())}' given as package" in captured.err.replace("\n", " ")
+    assert "Expected the name of an installed package" in captured.err.replace("\n", " ")
--- a/tests/test_reinstall.py
+++ b/tests/test_reinstall.py
@@ -1,5 +1,4 @@
 import sys
-from pathlib import Path
 
 import pytest  # type: ignore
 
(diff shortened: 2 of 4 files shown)


Gitznik:
What's the behavior if you pass a relative path? I assume it will also lead to unexpected results?

chrysle:
It just gets "normalized" and then appended to `pipx`'s venv path, due to the specialty noted above, therefore fails. I thought it was not worth the while catching this mistake, too, since few users should attempt something like that and be able to solve it easily by the output given. We can not disallow passing in existing paths by default, since installing from a local archive for example is also a case.

Gitznik (dismissed):

jayqi:
Ah, classic two-absolute-paths-where-one-clobbers-the-other.

I'm also generally kind of suspicious of allowing paths being passed in to reinstall at all. I get that relative paths will likely result in a no-op, and actually I think that is bad. One of the reasons I ended up trying an absolute path was because I tried a relative path and got a "nothing to do" message instead of something telling me that I was wildly misusing the argument. 

> We can not disallow passing in existing paths by default, since installing from a local archive for example is also a case.

This doesn't sound right to me for `reinstall`? If `reinstall` is supposed to expect the name of a package (which is also the name of pipx's venv for it), even if someone had originally installed from a local source path, shouldn't the name just be the name of the package? 

If this is the case, then it feels like we should be stricter about `reinstall` only accepting valid package names and not accepting paths at all. PyPA's specification has a regex for valid package names: https://packaging.python.org/en/latest/specifications/name-normalization/#name-format

I also think the error messages here could be clearer. If there's no venv named, e.g., `foobar`, then currently it says:

> Nothing to reinstall for foobar 😴

which is pretty ambiguous. I _did_ have a package installed (that I was trying to pass in a path for) and this message made it seem like pipx just didn't think reinstallation was necessary. 

This would be much more clear to the user if it says something like 

> No package named foobar installed

chrysle:
> This doesn't sound right to me for `reinstall`? If `reinstall` is supposed to expect the name of a package (which is also the name of pipx's venv for it), even if someone had originally installed from a local source path, shouldn't the name just be the name of the package?

That's because `pipx install` also makes use of this function. Probably we should handle this case separately for the `reinstall` command, as you suggested. Thinking of it, this seems to be much more reasonable. I could rework the PR towards that.

> This would be much more clear to the user if it says something like
> 
> > No package named foobar installed

I agree with that. Would you like to improve it in a separate PR?

chrysle:
Could you review again?

src/pipx/main.py
```diff
@@ -185,14 +185,28 @@ def run_pipx_command(args: argparse.Namespace, subparsers: Dict[str, argparse.Ar
...
+        if args.command == "reinstall":
+            # Passing paths into `reinstall` might have unintended
+            # side effects.
+            if Path(package).is_absolute() or Path(package).exists():
+                raise PipxError(
+                    pipx_wrap(
+                        f"""
+                        Error: Path '{package}' given as
+                        package name.
+                        """
```
Gitznik, line 206: The `pipx reinstall` help message calls it just `package`, not `package name`. I'd rephrase this accordingly.
```suggestion
                        f"""
                        Error: Path '{package}' given as
                        package. Expected the name of
                        an installed package.
                        """
```

tests/test_reinstall.py
```diff
@@ -54,3 +55,20 @@ def test_reinstall_specifier(pipx_temp_env, capsys):
...
+def test_reinstall_with_path(pipx_temp_env, capsys, path):
+    try:
+        Path(path).mkdir(parents=True, exist_ok=True)
+    except PermissionError:
+        pass
+
+    assert run_pipx_cli(["reinstall", path])
+    captured = capsys.readouterr()
+
+    assert "given as package name" in captured.err
```
Gitznik, line 74: Can you write this test in a way that does not pollute the runners system?
For example for the absolute paths, use paths within the `tmp_path` of the test, and use `monkeypatch.chdir` to use relative paths within the `tmp_path`.

Gitznik (changes requested):
I'm on vacation for the next few weeks, so someone else needs to take over the reviews for this.

tests/test_reinstall.py
chrysle: Ah, right, thanks for catching.

Gitznik (approved):
Nice, lgtm

PR #1338, opened 2024-04-10
docs: change --global parameter position on ensurepath command

marcosalberto:
…tation

## Summary of changes

Change --global parameter position on ensurepath command

## Test plan

Tested by running

```
➜  ~ sudo pipx ensurepath --global
Password:
usage: pipx [-h] [--quiet] [--verbose] [--global] [--version]
            {install,uninject,inject,upgrade,upgrade-all,uninstall,uninstall-all,reinstall,reinstall-all,list,interpreter,run,runpip,ensurepath,environment,completions} ...
pipx: error: unrecognized arguments: --global

➜  ~ sudo pipx --global ensurepath
/usr/local/bin is already in PATH.

⚠️  All pipx binary directories have been added to PATH. If you are sure you want to proceed, try again with the '--force' flag.

Otherwise pipx is ready to go! ✨ 🌟 ✨
➜  ~
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ _For comparison to other tools including pipsi, see
 ```
 brew install pipx
 pipx ensurepath
-sudo pipx ensurepath --global # optional to allow pipx actions with --global argument
+sudo pipx --global ensurepath # optional to allow pipx actions with --global argument
 ```
 
 Upgrade pipx with `brew update && brew upgrade pipx`.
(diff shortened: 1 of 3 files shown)


Gitznik (approved):
Thank you very much for raising and fixing the problem :+1:

chrysle:
I believe we could also add the `--global` option to the shared parser, so that the original command works? This syntax is a bit unpractical.

dukecat0:
> I believe we could also add the --global option to the shared parser, so that the original command works? 

I am +1 on this.

dukecat0:
@marcosalberto Thanks for your contribution, but I think it's better to fix the problem from the source, instead of asking the user to change the order.

marcosalberto:
@dukecat0 for sure, my PR aimed to fix the documentation fast to turn the experience for new users easy.

Thanks.

Issue #1132, opened 2023-12-03
Error logs overwrite each other when using `reinstall-all`


corneliusroemer:
**Describe the bug**

When running `pipx reinstall-all --python "python3.12"` and getting error logs from multiple reinstallations, the logs are overwritten.

Each reinstallation should get its own log path, or logs should be appended to and not overwritten.

**How to reproduce**

1. Run a command that errors for multiple installations, maybe something like `pipx reinstall-all --python "python3.16"` (because python3.16 doesn't exist)
2. Check out the error logs for earlier errors, i.e. `/Users/USERNAME/.local/pipx/logs/cmd_2023-12-03_14.10.06_pip_errors.log`

**Expected behavior**

The logs are appended to, rather than overwritten.

**Details**

This is the terminal output I get:

```log
 pipx reinstall-all --python "python3.12"
uninstalled parquet-cli! ✨ 🌟 ✨
Fatal error from pip prevented installation. Full pip output in file:
    /Users/corneliusromer/.local/pipx/logs/cmd_2023-12-03_14.10.06_pip_errors.log

pip seemed to fail to build package:
    six>=1.5

Some possibly relevant errors from pip install:
    TypeError: rmtree() got an unexpected keyword argument 'onexc'

Error installing parquet-cli.
uninstalled hatch! ✨ 🌟 ✨
Fatal error from pip prevented installation. Full pip output in file:
    /Users/corneliusromer/.local/pipx/logs/cmd_2023-12-03_14.10.06_pip_errors.log

pip seemed to fail to build package:
    more-itertools

Some possibly relevant errors from pip install:
    TypeError: rmtree() got an unexpected keyword argument 'onexc'

Error installing hatch.
uninstalled pycowsay! ✨ 🌟 ✨
Fatal error from pip prevented installation. Full pip output in file:
    /Users/corneliusromer/.local/pipx/logs/cmd_2023-12-03_14.10.06_pip_errors.log

pip seemed to fail to build package:
    pycowsay

Some possibly relevant errors from pip install:
    TypeError: rmtree() got an unexpected keyword argument 'onexc'

Error installing pycowsay.
uninstalled chatGPT! ✨ 🌟 ✨
Fatal error from pip prevented installation. Full pip output in file:
    /Users/corneliusromer/.local/pipx/logs/cmd_2023-12-03_14.10.06_pip_errors.log

pip seemed to fail to build package:
    git+https://github.com/mmabrouk/chatgpt-wrapper

[...]

PR #1348, opened 2024-04-16
feature: make install-all gather errors in batch

huxuan:

## Summary of changes

As discussed in https://github.com/pypa/pipx/issues/1132#issuecomment-2053680533, `install-all` needs a more robust strategy to deal with failed packages, so almost the same logic as `reinstall-all` is copied here. Besides, a test case for `install-all` is added.

## Test plan

Tested by running

```
nox -s tests
```

<del>BTW, this pull request is a refactor, so no news is added and there is also no referenced issue. Let me know if it is necessary.</del>

### DIFF
--- a/src/pipx/commands/install.py
+++ b/src/pipx/commands/install.py
@@ -10,6 +10,7 @@ from pipx.constants import (
     EXIT_CODE_OK,
     ExitCode,
 )
+from pipx.emojis import sleep
 from pipx.interpreter import DEFAULT_PYTHON
 from pipx.pipx_metadata_file import PackageInfo, PipxMetadata, _json_decoder_object_hook
 from pipx.util import PipxError, pipx_wrap
@@ -188,6 +189,7 @@ def install_all(
     """Return pipx exit code."""
     venv_container = VenvContainer(paths.ctx.venvs)
     failed: List[str] = []
+    installed: List[str] = []
 
     for venv_metadata in extract_venv_metadata(spec_metadata_file):
         # Install the main package
@@ -227,6 +229,10 @@ def install_all(
         except PipxError as e:
             print(e, file=sys.stderr)
             failed.append(venv_dir.name)
+        else:
+            installed.append(venv_dir.name)
+    if len(installed) == 0:
+        print(f"No packages installed after running 'pipx install-all {spec_metadata_file}' {sleep}")
     if len(failed) > 0:
         raise PipxError(f"The following package(s) failed to install: {', '.join(failed)}")
     # Any failure to install will raise PipxError, otherwise success
--- a/src/pipx/commands/install.py
+++ b/src/pipx/commands/install.py
@@ -232,7 +232,7 @@ def install_all(
         else:
             installed.append(venv_dir.name)
     if len(installed) == 0:
-        print(f"No packages installed after running 'pipx install-all {spec_metadata_file}' {sleep}")
+        print(f"{sleep} No packages installed after running 'pipx install-all {spec_metadata_file}'")
     if len(failed) > 0:
         raise PipxError(f"The following package(s) failed to install: {', '.join(failed)}")
     # Any failure to install will raise PipxError, otherwise success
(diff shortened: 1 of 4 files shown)


chrysle:
Thanks for the quick submission; please add a changelog entry of type _feature_, as this is a fundamental change to the behaviour.

huxuan:
> Thanks for the quick submission; please add a changelog entry of type _feature_, as this is a fundamental change to the behaviour.

Thanks for the response. So what the issue number should be? Or maybe I should create one for this?

tests/test_install_all.py
```diff
@@ -0,0 +1,18 @@
...
+    _ = capsys.readouterr()
+
+    assert not run_pipx_cli(["list", "--json"])
+    captured = capsys.readouterr()
+
+    pipx_list_path = Path(tmp_path) / "pipx_list.json"
+    with open(pipx_list_path, "w") as pipx_list_fh:
+        pipx_list_fh.write(captured.out)
+
+    assert not run_pipx_cli(["install-all", str(pipx_list_path)])
```
dukecat0, line 18:
```suggestion
    assert not run_pipx_cli(["install-all", str(pipx_list_path)])
    
    captured = capsys.readouterr()
    assert "black" in captured.out
    assert "pycowsay" in captured.out
```

src/pipx/commands/install.py
```diff
@@ -185,42 +186,52 @@ def install_all(
...
+                    verbose=verbose,
+                    include_apps=inject_package.include_apps,
+                    include_dependencies=inject_package.include_dependencies,
+                    force=force,
+                    suffix=inject_package.suffix == main_package.suffix,
+                )
+        except PipxError as e:
+            print(e, file=sys.stderr)
+            failed.append(venv_dir.name)
+        else:
```
dukecat0, line 231: I don't think this is necessary here according to this:

https://github.com/pypa/pipx/blob/44023628756588ca74c9a2256037e796cd02613a/src/pipx/commands/install.py#L120

tests/test_install_all.py
chrysle, line 1: Let's add a test for the new error-gathering functionality.

src/pipx/commands/install.py
dukecat0: For `reinstall_all`, it's a little bit different than this one cause it needs to set `first_reinstall` to False: https://github.com/pypa/pipx/blob/main/src%2Fpipx%2Fcommands%2Freinstall.py#L136-L136

huxuan: OK, will do.

huxuan: Tracked in #1357

tests/test_install_all.py
```diff
@@ -0,0 +1,30 @@
...
+    assert "black" in captured.out
+    assert "pycowsay" in captured.out
+
+
+def test_install_all_multiple_errors(pipx_temp_env, root, capsys):
+    pipx_metadata_path = root / "testdata" / "pipx_metadata_multiple_errors.json"
+    assert run_pipx_cli(["install-all", str(pipx_metadata_path)])
+    captured = capsys.readouterr()
+    assert "The following package(s) failed to install: dotenv, weblate" in captured.err
+    assert "No packages installed after running 'pipx install-all" in captured.out
```
dukecat0, line 30: Just a simple typo :)
```suggestion
    assert "No packages installed after running 'pipx install-all'" in captured.out
```

dukecat0 (dismissed):
Thanks!!

src/pipx/commands/install.py
```diff
@@ -186,41 +188,52 @@ def install_all(
...
+            installed.append(venv_dir.name)
+    if len(installed) == 0:
+        print(f"No packages installed after running 'pipx install-all {spec_metadata_file}' {sleep}")
```
chrysle, line 235: Stylistic nitpick, we usually format our output like this.

```suggestion
        print(f"{sleep} No packages installed after running 'pipx install-all {spec_metadata_file}'")
```

huxuan: Done.

huxuan:
> Looks good to me! Ha, and I was actually mistaken with the "stylistic convention" – but doesn't matter I guess.

That is OK, we can follow up this in #1357. I will mark it as ready for review then.

chrysle:
Thanks!

PR #1350, opened 2024-04-16
fix: version check for standalone python

### LINKED ISSUE #1349: Version check for standalone python is wrong.
**Describe the bug**

When specify python version to `3.1`, python 3.12 will be downloaded

**How to reproduce**

```shell
pipx install --python 3.1 --fetch-missing-python pycowsay
```

**Expected behavior**

Raise an error that no valid Python found.

huxuan:

## Summary of changes

Fix #1349 

BTW, the changes also include some code refactor for related part.

## Test plan

Tested by running

```
pipx install --python 3.1 --fetch-missing-python pycowsay
```

### DIFF
--- a/changelog.d/1349.bugfix.md
+++ b/changelog.d/1349.bugfix.md
@@ -1,0 +1,1 @@
+fix version check for standalone python
--- a/src/pipx/standalone_python.py
+++ b/src/pipx/standalone_python.py
@@ -44,10 +44,6 @@ GITHUB_API_URL = "https://api.github.com/repos/indygreg/python-build-standalone/
 PYTHON_VERSION_REGEX = re.compile(r"cpython-(\d+\.\d+\.\d+)")
 
 
-class NotAvailable(Exception):
-    """Raised when the asked Python version is not available."""
-
-
 def download_python_build_standalone(python_version: str):
     """When all other python executable resolutions have failed,
     attempt to download and use an appropriate python build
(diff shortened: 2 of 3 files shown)


chrysle:
Thanks for catching this. A test should be useful.

huxuan:
> Thanks for catching this. A test should be useful.

Thanks for the review. A test cases for the issue is added, PTAL

Actually, I am not sure whether we should add such kind of tests since it is more like a hard-coding. 

I also thought to add some positive test cases but found it will make everything a little complicated to ensure the specific python version is not available especially considering tests against all python versions. If necessary, maybe we should do it in a separate pull request?

chrysle:
> Actually, I am not sure whether we should add such kind of tests since it is more like a hard-coding.

I generally agree that adding tests that ensure certain features work as expected is a better way of coverage. However, this test case ensures an error is raised correctly on a certain occasion, thus I consider it useful.
 
> I also thought to add some positive test cases but found it will make everything a little complicated to ensure the specific python version is not available especially considering tests against all python versions and all platforms. 

It sounds reasonable to me to cover that.

huxuan:
> > I also thought to add some positive test cases but found it will make everything a little complicated to ensure the specific python version is not available especially considering tests against all python versions and all platforms.
> 
> It sounds reasonable to me to cover that.

OK, I just forgot the usage of monkey mock and learned from other places, will add some more cases later.

chrysle:
Thanks!

Issue #1354, opened 2024-04-19
Virtual Environment location is invalid when passing an absolute path


huxuan:
**Describe the bug**

When passing an absolute path as `package` argument, `VenvContainer.get_venv_dir()` returns the absolute path directly and brings bunch of abnormal behaviors. Esepcially when the path is normalized which gets identical result after `packaging.util.canonicalize_name()`, `reinstall` and `uninstall` will remove the absolute path unexpectedly (as reported in #1324).

1. All related commands has error in logs:  `pipx >(run_pipx_command:197): Virtual Environment location is /tmp/empty-project`
2. `pipx uninstall` and `pipx reinstall` will remove the directory `/tmp/empty-project` 

**How to reproduce**

```shell
$ cp -r /path/to/pipx/testdata/empty_project tmp/empty-project

$ pipx runpip --verbose /tmp/empty-project list
pipx >(setup:924): pipx version is 1.5.0
pipx >(setup:925): Default python interpreter is '/usr/local/pipx/venvs/pipx/bin/python'
pipx >(run_pipx_command:197): Virtual Environment location is /tmp/empty-project
venv for '/tmp/empty-project' was not found. Was '/tmp/empty-project' installed with pipx?

$ pipx upgrade --verbose /tmp/workspace
pipx >(setup:924): pipx version is 1.5.0
pipx >(setup:925): Default python interpreter is '/usr/local/pipx/venvs/pipx/bin/python'
pipx >(run_pipx_command:197): Virtual Environment location is /tmp/empty-project
pipx >(_upgrade_venv:136): Ignoring --python as not combined with --install
Not upgrading empty-project. It has missing internal pipx metadata.
It was likely installed using a pipx version before 0.15.0.0.
Please uninstall and install this package to fix.

$ pipx inject --verbose /tmp/empty-project pywcowsay
pipx >(setup:924): pipx version is 1.5.0
pipx >(setup:925): Default python interpreter is '/usr/local/pipx/venvs/pipx/bin/python'
pipx >(run_pipx_command:197): Virtual Environment location is /tmp/empty-project
Can't inject 'pywcowsay' into Virtual Environment 'empty-project'. 'empty-project' has missing internal pipx metadata. It was likely installed using a pipx version before 0.15.0.0.
Please uninstall and install 'empty-project', or reinstall-all to fix.

$ pipx uninject --verbose /tmp/empty-project pywcowsay
pipx >(setup:924): pipx version is 1.5.0
[...]

PR #1356, opened 2024-04-19
fix: validate absolute path as packages arg on Windows correctly

### LINKED ISSUE #1355: `pipx upgrade` should validate windows absolute path correctly
**Describe the bug**

Per the discussion here: https://github.com/pypa/pipx/pull/1339#discussion_r1564150055, when passing an absolute path to `pipx upgrade` on Windows, it will complains about `Package cannot be a url. Package name should be passed instead.`

**How to reproduce**

```
PS C:\> pipx.local upgrade --verbose C:\
pipx >(setup:990): pipx version is 1.5.1.dev14+g4402362
pipx >(setup:991): Default python interpreter is 'C:\Users\huxuan\AppData\Local\pipx\pipx\venvs\pipx-local\Scripts\python.exe'
Package cannot be a url. Package name should be passed instead.
```

**Expected behavior**

The error should complain about absolute path rather than url.

BTW, this issue is partially depends on #1354 as it may not cover the following processing of the absolute path. <del>Maybe we can just combine the changes together.</del>

huxuan:
…ctly

## Summary of changes

Fix #1355 

## Test plan

Tested by running

```
nox -s tests
```


tests/test_upgrade.py
```diff
@@ -116,3 +116,10 @@ def test_upgrade_multiple(pipx_temp_env, capsys):
     captured = capsys.readouterr()
     assert f"upgraded package {name} from {initial_version} to" in captured.out
     assert "pycowsay is already at latest version" in captured.out
+
+
+def test_upgrade_absolute_path(pipx_temp_env, capsys, root):
+    print(str((root / "testdata" / "empty_project").resolve()))
+    assert run_pipx_cli(["upgrade", "--verbose", str((root / "testdata" / "empty_project").resolve())])
+    captured = capsys.readouterr()
+    assert "Package cannot be a url" not in captured.err
```
chrysle, line 125:
```suggestion
    assert "Package cannot be a URL" not in captured.err
```

src/pipx/main.py
```diff
@@ -174,6 +174,12 @@ def get_venv_args(parsed_args: Dict[str, str]) -> List[str]:
     return venv_args
 
 
+def validate_package_arg(package: str):
```
chrysle, line 177:
```suggestion
def package_is_url(package: str, raise: bool = True): -> bool
```

src/pipx/main.py
```diff
@@ -174,6 +174,12 @@ def get_venv_args(parsed_args: Dict[str, str]) -> List[str]:
...
+    url_parse_package = urllib.parse.urlparse(package)
+    if url_parse_package.scheme and url_parse_package.netloc:
+        raise PipxError("Package cannot be a url. Package name should be passed instead.")
```
chrysle, line 180:
```suggestion
    if url_parse_package.scheme and url_parse_package.netloc:
        if not raise:
            return True
        raise PipxError("Package cannot be a URL. A valid package name should be passed instead.")
    return False   
```

src/pipx/main.py
```diff
@@ -212,8 +216,8 @@ def run_pipx_command(args: argparse.Namespace, subparsers: Dict[str, argparse.Ar
         logger.info(f"Virtual Environment location is {venv_dir}")
 
     if "packages" in args:
-        if any(urllib.parse.urlparse(package).scheme for package in args.packages):
-            raise PipxError("Package cannot be a url. Package name should be passed instead.")
+        for package in args.packages:
+            validate_package_arg(package)
```
chrysle, line 220:
```suggestion
            package_is_url(package)
```

src/pipx/main.py
```diff
@@ -185,9 +191,7 @@ def run_pipx_command(args: argparse.Namespace, subparsers: Dict[str, argparse.Ar
 
     if "package" in args:
         package = args.package
-        url_parse_package = urllib.parse.urlparse(package)
-        if url_parse_package.scheme and url_parse_package.netloc:
-            raise PipxError("Package cannot be a url")
+        validate_package_arg(package)
 
         if "spec" in args and args.spec is not None:
             if urllib.parse.urlparse(args.spec).scheme:
```
chrysle, line 197:
```suggestion
        if "spec" in args and args.spec is not None:
            if package_is_url(args.spec, raise=False):
```

changelog.d/1355.bugfix.md
```diff
@@ -0,0 +1 @@
+ validate the absolute path as packages argument on Windows correctly
```
chrysle, line 1:
```suggestion
Validate whether a package is an URL correctly.
```

huxuan:
Hi @chrysle, thanks for your review.

I have a little concern whether we should combine the check for `package` argument in one place. Currently we need to check it is not a URL, and according to the discussion in #1354, we also need check it is not a absolute path. Any ideas?

chrysle:
Hmm, since we only need the path validation in one place, but can reuse the newly introduced function, I'd be inclined to keep it as-is. Or do you mean something else?

huxuan:
> Hmm, since we only need the path validation in one place, but can reuse the newly introduced function, I'd be inclined to keep it as-is. Or do you mean something else?

I thought to combine all the validation for `package` and `packages` argument in one function, but I did not releaize we also need to ensure the `spec` argument is not a URL but not for an absolute path. So it is fine to keep it as it is now.

In this way, we need another function to fix #1354, something like `package_is_abspath`. This will be used in the validation for `package` and `packages` but not for `spec`.

Let me know if there is anything wrong.

chrysle (approved):

chrysle:
Thanks!

PR #1367, opened 2024-04-24
ci: macos-latest now points at macos-14

henryiii:
ARM is fine, but no Python 3.8 or 3.9 is not.

Before change trigger: https://github.com/pypa/pipx/actions/runs/8810668602

(latest hasn't switched here yet, it's a rollout)

Edit: looks like they are preparing 3.8 and 3.9 for ARM.

### DIFF
--- a/.github/workflows/exhaustive_package_test.yml
+++ b/.github/workflows/exhaustive_package_test.yml
@@ -11,8 +11,19 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, windows-latest]
         python-version: ["3.12", "3.11", "3.10", "3.9", "3.8"]
+        include:
+          - os: macos-13
+            python-version: "3.8"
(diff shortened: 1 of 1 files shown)


chrysle:
Python 3.8 & 3.9 are now supported on MacOS 14.

henryiii:
Where is now is fine then unless you specifically would like to have an old Intel runner job too, but they will probably go away at some point.

chrysle:
> unless you specifically would like to have an old Intel runner job too, but they will probably go away at some point.

Probably that would be a good idea? This was already discussed for Pillow in https://github.com/python-pillow/Pillow/pull/8010#issuecomment-2079144483.

henryiii:
It's a bit (lot) more important for a package that deals with binaries. pipx doesn't, really.

chrysle:
Thanks for the PR nontheless, then! (It'd be appreciated if you could have a look at #1246, BTW; probably I'm doing something fundamentally wrong).

henryiii:
I'll try to look at it by Tuesday, ping me if I forget! Didn't see anything really glaring yet.

PR #1373, opened 2024-04-25
fix: update test packages to fix tests on macos-latest

huxuan:
- [x] I have added a news fragment under `changelog.d/` (if the patch affects the end users)

## Summary of changes

Fix #1372 

1. Update some packages to newer version to install on macos-latest successfully.
2. Update tests package lists
4. Update pylint related tests due to metadata change

## Test plan

Tested by running

```
# command(s) to exercise these changes
nox -s tests
```

PR #1374, opened 2024-04-26
fix: support python3.8 for standalone python builds

### LINKED ISSUE #1375: Support Python 3.8 in Standalone Python Builds.
**Describe the bug**

Currently, we do not support Python 3.8 for the standalone python build because we specified the `x86_64_v3` for download links, but according to the official documentation [1]:

> The x86_64_v2, x86_64_v3, and x86_64_v4 binaries usually crash on startup when run on an incompatible CPU. We don’t recommend running the x86_64_v4 builds in production because they likely don’t yield a reliable performance benefit. Unless you are executing these binaries on a CPU older than ~2008 or ~2013, we recommend running the x86_64_v2 or x86_64_v3 binaries, as these should be slightly faster since they take advantage of more modern CPU instructions which are more efficient. But if you want maximum portability, stick with the baseline x86_64 builds.

As mentioned in the last scentence, I think we should stick to the `x86_64` builds.

[1] https://gregoryszorc.com/docs/python-build-standalone/main/running.html?highlight=x86_64_v3#obtaining-distributions

**How to reproduce**

```shell
$ pipx install --verbose --python 3.8 --fetch-missing-python black
pipx >(setup:924): pipx version is 1.5.0
pipx >(setup:925): Default python interpreter is '/usr/local/pipx/venvs/pipx/bin/python'
⚠ No executable for the provided Python version '3.8' found in the python-build-standalone project. Please make sure the provided version is listed in
    https://github.com/indygreg/python-build-standalone/releases/latest.
```

**Expected behavior**

Be able to install Python 3.8 with standalone python builds.

huxuan:

## Summary of changes

Fix #1375

## Test plan

Tested by running

```
nox -s tests
pipx install --verbose --python 3.8 --fetch-missing-python black
```

BTW, the tests on macos-latest is failed and I am trying to fix it in #1373 where I need python 3.8 interpreter for testing.

### DIFF
--- a/changelog.d/1375.bugfix.md
+++ b/changelog.d/1375.bugfix.md
@@ -1,0 +1,1 @@
+Support python3.8 for standalone python builds
--- a/src/pipx/standalone_python.py
+++ b/src/pipx/standalone_python.py
@@ -33,8 +33,8 @@ MACHINE_SUFFIX: Dict[str, Dict[str, Any]] = {
             # musl doesn't exist
         },
         "x86_64": {
-            "glibc": "x86_64_v3-unknown-linux-gnu-install_only.tar.gz",
-            "musl": "x86_64_v3-unknown-linux-musl-install_only.tar.gz",
+            "glibc": "x86_64-unknown-linux-gnu-install_only.tar.gz",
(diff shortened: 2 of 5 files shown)


dukecat0 (approved):
Thanks!

PR #1378, opened 2024-04-26
Fix `--system-site-packages` for `upgrade --install`

### LINKED ISSUE #1344: `pipx upgrade --install` ignores `--system-site-packages` option
**Describe the bug**

Recently in [this PR](https://github.com/pypa/pipx/pull/1267) the `--install` option was added to the `upgrade` command.
Unfortunately this doesn't seem to respect the `--system-site-packages` option.

**How to reproduce**

When using `upgrade --install` the `--system-site-packages` option isn't respected:
```sh
❯ pipx upgrade --install --system-site-packages pycowsay
❯ grep system $HOME/.local/share/pipx/venvs/pycowsay/pyvenv.cfg
include-system-site-packages = false
```

Comparison with the `install` command, where it works correctly:
```sh
❯ pipx install --system-site-packages pycowsay
❯ grep system $HOME/.local/share/pipx/venvs/pycowsay/pyvenv.cfg
include-system-site-packages = true
command = /usr/bin/python3 -m venv --without-pip --system-site-packages /home/sadamczyk/.local/share/pipx/venvs/pycowsay
```

**Expected behavior**

The `upgrade --install` command should handle the `--system-site-packages` option just like the `install` command does.

**Investigation**

Looking through the code, the `--system-site-packages` option seems to be passed to the `install` command in the `venv_args` variable from the `get_venv_args` function: https://github.com/pypa/pipx/blob/5482fac6c3dfc3775af398d630ef30102f15bdcc/src/pipx/main.py#L170-L174
https://github.com/pypa/pipx/blob/5482fac6c3dfc3775af398d630ef30102f15bdcc/src/pipx/main.py#L182

But `venv_args` is only passed to the `run`, `install` and `install-all` commands
https://github.com/pypa/pipx/blob/5482fac6c3dfc3775af398d630ef30102f15bdcc/src/pipx/main.py#L244-L253
... but not to the `upgrade` command:
https://github.com/pypa/pipx/blob/5482fac6c3dfc3775af398d630ef30102f15bdcc/src/pipx/main.py#L294-L304

It might be enough to add the required function arguments so that the `venv_args` from the main.py can be passed through all the way to the `install` call here:
https://github.com/pypa/pipx/blob/5482fac6c3dfc3775af398d630ef30102f15bdcc/src/pipx/commands/upgrade.py#L111-L114

sadamczyk:
Fix #1344

## Summary of changes

The `--system-site-package` option currently doesn't do anything when used with `upgrade --install`.
This PR adds `venv_args` to the upgrade command so that it can be passed through to the install command, when the `--install` option is set.

## Test plan

Tested by running

```
pipx upgrade --install --system-site-packages pycowsay
grep system $HOME/.local/share/pipx/venvs/pycowsay/pyvenv.cfg
```

Before the PR this returns:
```
include-system-site-packages = false
```

After applying the PR it returns:
```
include-system-site-packages = true
command = /usr/bin/python3.12 -m venv --without-pip --system-site-packages /home/sadamczyk/.local/share/pipx/venvs/pycowsay
```

Also successfully tested by installing https://github.com/essembeh/gnome-extensions-cli which requires the system site packages to work for the `--dbus` integration. Before this ran into an error because of the missing required 
system dependencies.
```
pipx upgrade --install --system-site-packages gnome-extensions-cli
gext --dbus update --install -y nightthemeswitcher@romainvigier.fr
```

### DIFF
--- a/changelog.d/1344.bugfix.md
+++ b/changelog.d/1344.bugfix.md
@@ -1,1 +1,1 @@
-Make the --system-site-packages option actually work with upgrade --install
+Make the Python `venv` module arguments work with `upgrade --install`
--- a/changelog.d/1344.bugfix.md
+++ b/changelog.d/1344.bugfix.md
@@ -1,0 +1,1 @@
+Make the --system-site-packages option actually work with upgrade --install
--- a/src/pipx/commands/upgrade.py
+++ b/src/pipx/commands/upgrade.py
@@ -104,15 +104,18 @@ def _upgrade_venv(
     upgrading_all: bool,
     force: bool,
     install: bool = False,
+    venv_args: Optional[List[str]] = None,
     python: Optional[str] = None,
     python_flag_passed: bool = False,
 ) -> int:
     """Return number of packages with changed versions."""
     if not venv_dir.is_dir():
         if install:
+            if venv_args is None:
+                venv_args = []
             commands.install(
                 venv_dir=None,
-                venv_args=[],
+                venv_args=venv_args,
                 package_names=None,
                 package_specs=[str(venv_dir).split(os.path.sep)[-1]],
                 local_bin_dir=paths.ctx.bin_dir,
@@ -134,6 +137,9 @@ def _upgrade_venv(
                 does not exist.
                 """
             )
+
+    if venv_args and not install:
+        logger.info("Ignoring " + ", ".join(venv_args) + " as not combined with --install")
 
     if python and not install:
         logger.info("Ignoring --python as not combined with --install")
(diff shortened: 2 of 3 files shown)


changelog.d/1344.bugfix.md
```diff
@@ -0,0 +1 @@
+Make the --system-site-packages option actually work with upgrade --install
```
chrysle, line 1:
```suggestion
Make the Python `venv` module arguments work with `upgrade --install`
```

chrysle (approved):

chrysle:
Thanks!

Issue #1377, opened 2024-04-26
pipx --preinstall ignores dependency version specifier


davidpeckham:
When I use pipx --preinstall to install a specific version of a dependency before installing the main package, pipx installs the latest version of the dependency instead.

**How to reproduce**

Ask pipx to install a dependency with a specific version before installing the main package:

```sh
% pipx install --preinstall virtualenv==20.25.3 tox --verbose
pipx >(setup:924): pipx version is 1.5.0
pipx >(setup:925): Default python interpreter is '/Library/Frameworks/Python.framework/Versions/3.11/bin/python3'
pipx >(package_name_from_spec:379): Determined package name: tox
pipx >(package_name_from_spec:380): Package name determined in 0.0s
creating virtual environment...
pipx >(run_subprocess:174): running /Library/Frameworks/Python.framework/Versions/3.11/bin/python3 -m venv --without-pip /Users/foo/.local/pipx/venvs/tox
pipx >(run_subprocess:174): running <checking pip's availability>
pipx >(run_subprocess:174): running /Users/foo/.local/pipx/venvs/tox/bin/python -c import sysconfig; print(sysconfig.get_path('purelib'))
pipx >(run_subprocess:174): running /Users/foo/.local/pipx/shared/bin/python -c import sysconfig; print(sysconfig.get_path('purelib'))
pipx >(run_subprocess:174): running /Users/foo/.local/pipx/venvs/tox/bin/python --version
pipx >(package_name_from_spec:379): Determined package name: virtualenv
pipx >(package_name_from_spec:380): Package name determined in 0.0s
upgrading virtualenv...
pipx >(run_subprocess:174): running /Users/foo/.local/pipx/venvs/tox/bin/python -m pip --no-input install --upgrade virtualenv
Looking in indexes: https://nexusrepo.app.mycompany.net/repository/pypilocal/simple/
Collecting virtualenv
  ERROR: HTTP error 403 while getting https://nexusrepo.app.mycompany.net/repository/pypilocal/packages/virtualenv/20.26.0/virtualenv-20.26.0-py3-none-any.whl#sha256=0846377ea76e818daaa3e00a4365c018bc3ac9760cbb3544de542885aad61fb3 (from https://nexusrepo.app.mycompany.net/repository/pypilocal/simple/virtualenv/) (requires-python:>=3.7)
[...]

PR #1379, opened 2024-04-26
pipx --preinstall ignores dependency version specifier #1377

davidpeckham:

## Summary of changes

[chrysle](https://github.com/chrysle) found the problem -- commands/install.py calls `package_name_from_spec` when it shouldn't. I removed that call and passed the dependencies unmodified to `venv.upgrade_package_no_metadata`.

## Test plan

I added this test case to tests/test_install.py:

```python
def test_preinstall_specific_version(pipx_temp_env, caplog):
    assert not run_pipx_cli(["install", "--preinstall", "black==22.8.0", "nox"])
    assert "black==22.8.0" in caplog.text
```

Tested by running

```
pipx install --preinstall black==22.8.0 nox
```

### DIFF
--- a/changelog.d/1377.bugfix.md
+++ b/changelog.d/1377.bugfix.md
@@ -1,0 +1,1 @@
+Install specified version of `--preinstall` dependency instead of latest version
--- a/src/pipx/commands/install.py
+++ b/src/pipx/commands/install.py
@@ -91,8 +91,7 @@ def install(
             override_shared = package_name == "pip"
             venv.create_venv(venv_args, pip_args, override_shared)
             for dep in preinstall_packages or []:
-                dep_name = package_name_from_spec(dep, python, pip_args=pip_args, verbose=verbose)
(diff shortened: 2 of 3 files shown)


chrysle:
How about running `pipx.package_specifier.parse_specifier_for_install` over the deps instead?

davidpeckham:
> How about running `pipx.package_specifier.parse_specifier_for_install` over the deps instead?

Thanks, I added that.

chrysle:
My bad, I've just seen the function is already called in `pipx.venv.install_package`. Could you revert the change?

chrysle:
@davidpeckham https://github.com/pypa/pipx/pull/1379/commits/7339ae01045f2bece87dd7641305fa3db9b694c2 seems to be empty...

davidpeckham:
@chrysle I think that last commit fixed it.

chrysle (approved):

chrysle:
Thanks!

PR #1390, opened 2024-05-04
Fix passing constraints file path into `pipx install` operation via `--pip-args` 

### LINKED ISSUE #1389: Error with `pipx install` Command Using Relative Path for Pip Arguments
**Describe the bug**

When using the `pipx install` command with pip arguments `-c` or `--constraint`, using a relative path, we encounter an error stating that the file does not exist, even though it does.

**How to reproduce**

`pipx install --pip-args=--constraint=some_constraint_file.txt some_package_inside_constraint_file`

**Expected behavior**

The package should be installed successfully.

I have already identified that this issue occurs following the changes introduced in PR https://github.com/pypa/pipx/pull/1237.

Is it necessary to run the commands in a different directory?

If so, I have already begun drafting a solution for this issue.
My approach involves checking for `-c` and `--constraint` flags and replacing relative paths with absolute paths before calling the subprocess.

Do you think this proposed solution is appropriate? If there is a simpler way to resolve this issue, please advise.

Thank you very much for this package!

guysalt:
closes #1389

## Summary of changes

Fix error with `pipx install` Command Using Relative Path for Pip Arguments

## Test plan

Tested by running

```
pipx install --pip-args=--constraint=some_constraint_file.txt some_package_inside_constraint_file
```

### DIFF
--- a/changelog.d/1389.bugfix.md
+++ b/changelog.d/1389.bugfix.md
@@ -1,0 +1,1 @@
+Fix passing constraints file path into `pipx install` operation via `pip` args
--- a/src/pipx/package_specifier.py
+++ b/src/pipx/package_specifier.py
@@ -163,7 +163,27 @@ def parse_specifier_for_install(package_spec: str, pip_args: List[str]) -> Tuple
         )
         pip_args.remove("--editable")
 
-    return (package_or_url, pip_args)
+    for index, option in enumerate(pip_args):
+        if not option.startswith(("-c", "--constraint")):
+            continue
+
+        if option in ("-c", "--constraint"):
+            argument_index = index + 1
+            if argument_index < len(pip_args):
+                constraints_file = pip_args[argument_index]
+                pip_args[argument_index] = str(Path(constraints_file).expanduser().resolve())
+
+        else:  # option == "--constraint=some_path"
+            option_list = option.split("=")
+
+            if len(option_list) == 2:
+                key, value = option_list
+                value_path = Path(value).expanduser().resolve()
+                pip_args[index] = f"{key}={value_path}"
+
+        break
+
+    return package_or_url, pip_args
 
 
 def parse_specifier_for_metadata(package_spec: str) -> str:
--- a/tests/test_install.py
+++ b/tests/test_install.py
@@ -284,6 +284,19 @@ def test_pip_args_with_constraint_relative_path(constraint_flag, pipx_temp_env, 
     assert subprocess_package_version_output != package_version
 
 
+@pytest.mark.parametrize("constraint_flag", ["-c ", "--constraint ", "--constraint="])
+def test_pip_args_with_wrong_constraint_fail(constraint_flag, pipx_ultra_temp_env, tmp_path, capsys):
+    constraint_file_name = "constraints.txt"
+    os.chdir(tmp_path)
+
+    assert run_pipx_cli(["install", f"--pip-args='{constraint_flag}{constraint_file_name}'", "pycowsay"])
+
+    assert (
+        f"ERROR: Could not open requirements file: [Errno 2] No such file or directory: '{constraint_file_name}'"
+        in capsys.readouterr().err
+    )
+
+
 def test_install_suffix(pipx_temp_env, capsys):
     name = "pbr"
 
--- a/tests/test_install.py
+++ b/tests/test_install.py
@@ -264,6 +264,26 @@ def test_pip_args_with_windows_path(pipx_temp_env, capsys):
     assert r"D:\\TEST\\DIR" in captured.err
 
 
+@pytest.mark.parametrize("constraint_flag", ["-c ", "--constraint ", "--constraint="])
+def test_pip_args_with_constraint_relative_path(constraint_flag, pipx_temp_env, tmp_path, caplog):
+    constraint_file_name = "constraints.txt"
+    package_name = "ipython"
+    package_version = "8.23.0"
+
+    os.chdir(tmp_path)
+    constraints_file = tmp_path / constraint_file_name
+    constraints_file.write_text(f"{package_name}!={package_version}")
+    constraints_file.touch()
+
+    assert not run_pipx_cli(["install", f"--pip-args='{constraint_flag}{constraint_file_name}'", package_name])
+
+    assert f"{constraint_flag}{constraints_file}" in caplog.text
+
+    subprocess_package_version = subprocess.run([package_name, "--version"], capture_output=True, text=True, check=False)
+    subprocess_package_version_output = subprocess_package_version.stdout.strip()
+    assert subprocess_package_version_output != package_version
+
+
 def test_install_suffix(pipx_temp_env, capsys):
     name = "pbr"
 
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -159,7 +159,9 @@ def get_pip_args(parsed_args: Dict[str, str]) -> List[str]:
         pip_args += ["--index-url", parsed_args["index_url"]]
 
     if parsed_args.get("pip_args"):
-        pip_args += shlex.split(parsed_args.get("pip_args", ""), posix=not WINDOWS)
+        # Stripping the single quote that can be parsed from several shells
+        pip_args_striped = parsed_args["pip_args"].strip("'")
+        pip_args += shlex.split(pip_args_striped, posix=not WINDOWS)
 
     # make sure --editable is last because it needs to be right before
     #   package specification


src/pipx/package_specifier.py
```diff
@@ -161,7 +161,24 @@ def parse_specifier_for_install(package_spec: str, pip_args: List[str]) -> Tuple
         )
         pip_args.remove("--editable")
 
-    return (package_or_url, pip_args)
+    pip_args = [option for pip_arg in pip_args for option in pip_arg.split(" ")]
```
chrysle, line 164: This doesn't seem necessary.

```suggestion
```

src/pipx/package_specifier.py
```diff
@@ -161,7 +161,24 @@ def parse_specifier_for_install(package_spec: str, pip_args: List[str]) -> Tuple
...
+    for option in pip_args:
+        if "-c" in option or "--constraint" in option:
+            option_list = option.split("=")
+
+            if len(option_list) != 2:
+                continue
+
+            key, value = option_list
+            value_path = Path(value)
+            if value_path.exists():
```
chrysle, line 175: We don't need to perform this check ourselves, `pip` will catch a non-existing constraint file for us.

changelog.d/1389.bugfix.md
```diff
@@ -0,0 +1 @@
+Fix error with `pipx install` Command Using Relative Path for Pip Arguments
```
chrysle, line 1:
```suggestion
Fix passing constraints file path into `pipx install` operation via `pip` args
```

src/pipx/package_specifier.py
```diff
@@ -161,7 +161,24 @@ def parse_specifier_for_install(package_spec: str, pip_args: List[str]) -> Tuple
         )
         pip_args.remove("--editable")
 
-    return (package_or_url, pip_args)
+    pip_args = [option for pip_arg in pip_args for option in pip_arg.split(" ")]
+
+    for option in pip_args:
+        if "-c" in option or "--constraint" in option:
```
chrysle, line 167: Small refactoring suggestion (don't apply, lacks proper indentation):

```suggestion
    if any(arg.startswith("--constraint") for arg in pip_args):
```

chrysle (commented):
Thanks for catching this! Please also extend the docstring of the changed function.

guysalt:
> Thanks for catching this! Please also extend the docstring of the changed function.

I have considered this `* Convert local paths to absolute paths` as the same description, doesn't it?

src/pipx/package_specifier.py
guysalt: If you have more than 1 pip args, the function get it as a single string.

https://pipx.pypa.io/stable/examples/ here for example there is the example:
`pipx install --pip-args='--index-url=<private-repo-host>:<private-repo-port> --trusted-host=<private-repo-host>:<private-repo-port>' private-repo-package`

In this example the `pip_args` argument will be:
```python
['--index-url=<private-repo-host>:<private-repo-port> --trusted-host=<private-repo-host>:<private-repo-port>']
```

So from this I assumed I need first to check for multiple flags in a single string and split it :)

src/pipx/package_specifier.py
guysalt: right, changed both lines to `value_path = Path(value).expanduser().resolve()`

changelog.d/1389.bugfix.md
guysalt: thank you!

src/pipx/package_specifier.py
guysalt: what about `"-c"`?

chrysle: True, updated the suggestion.

src/pipx/package_specifier.py
chrysle: > In this example the `pip_args` argument will be:
> 
> ```python
> ['--index-url=<private-repo-host>:<private-repo-port> --trusted-host=<private-repo-host>:<private-repo-port>']
> ```

That doesn't seem correct. The `pip` arguments get normalised and formatted already at an early stage:

https://github.com/pypa/pipx/blob/bd69963a4dbd90c95950921e99654dc9867ef10f/src/pipx/main.py#L161

src/pipx/package_specifier.py
```diff
@@ -161,7 +161,22 @@ def parse_specifier_for_install(package_spec: str, pip_args: List[str]) -> Tuple
...
+            option_list = option.split("=")
```
chrysle, line 168: This won't work when the constraints file is given without the equals sign.

src/pipx/package_specifier.py
guysalt: ok now I see, so it another bug... `shlex.split()` does not split anything between single quotes on posix platforms.

on shlex documentation:

> Enclosing characters in quotes which are not part of [escapedquotes](https://docs.python.org/3/library/shlex.html#shlex.shlex.escapedquotes) (e.g. "'") preserve the literal value of all characters within the quotes;

Do you have any idea for appropriate way to resolve it?

Since I dont know what exactly happening on windows I dont want just to strip the single quotes (or changing to regular function `string.split()`)

guysalt: also, what about the `-c` flag? (or other single dash flags)

now I understand I did wrong on the `-c` as on this flag should not be  with`=`... (e.g `-c some_file.txt`)

src/pipx/package_specifier.py
guysalt: done, but changed the inner block a little bit (as wrote above about the `-c` flag, please take a look :)

and again, thank you!

src/pipx/package_specifier.py
guysalt: changed

src/pipx/package_specifier.py
chrysle: > ok now I see, so it another bug... `shlex.split()` does not split anything between single quotes on posix platforms.

The passage you quoted means that any special characters in the string are preserved. It does not result in an incorrect splitting behaviour. I can assure you that works, as I reproduced it locally.

src/pipx/package_specifier.py
```diff
@@ -161,7 +161,27 @@ def parse_specifier_for_install(package_spec: str, pip_args: List[str]) -> Tuple
...
+            continue
+
+        if option == "-c":
+            next_option_index = index + 1
+            if next_option_index < len(pip_args):
+                next_option = pip_args[next_option_index]
+                pip_args[next_option_index] = str(Path(next_option).expanduser().resolve())
+
+        else:  # option == "--constraint"
+            option_list = option.split("=")
```
chrysle, line 175: The `--constraint` option can also be given without the equals sign, apparently.

src/pipx/package_specifier.py
guysalt: Are you saying you running this line and it split the strings to you?

```python
shlex.split("'-c .github/workflows/constraints.txt'")
```

Please double check it...

If so, do you have any idea why it does not work for me? maybe it related to python version? (using python 3.12.2)

src/pipx/package_specifier.py
guysalt: haha, im sorry for not checking this...
 
change this line:

`if option == "-c":`

to this:

`if option == "-c" or option == "--constraint":`

src/pipx/package_specifier.py
chrysle: The single quotes given on the command line aren't preserved when given to `shlex`, or rather there is no difference to double quotes in this case – it just tells the program (the argument parser) that the arguments given  belong as one single string to the option.

guysalt: Im sorry, but Im not sure I understand what do you mean...

I do the following steps:

1. install `pipx` locally
2. put print for pip_args inside `parse_specifier_for_install` function
3. calling `pipx install --pip-args='-c a' foo`
4. see printing this `['-c a']`

I do something wrong?
maybe we could talk at discord or something?

chrysle: I debugged the `pip_args` in `parse_specifier_for_install` directly, too (POSIX mode), calling exactly the same command, but it worked for me. What OS are you running? I can also investigate with `pyenv`, but unsure whether it's related to the Python version.

guysalt: I just realized it's probably because im using `xonsh` shell.

check it on `zsh` and it split the `pip_args` correctly.

I don't fully understand why this is happening on xonsh, but I will try to figure it out or maybe talk with them.

anyway is seems it does not relate to `pipx` or to this PR and I think we can proceed :)

chrysle:
Great, let's get down to business once again, then. Don't fail to inform us about your XONSH compatibility research.
Could you add a functional test, please?

guysalt:
> Great, let's get down to business once again, then. Don't fail to inform us about your XONSH compatibility research. Could you add a functional test, please?

sure, hope to do so soon...

guysalt:
> Great, let's get down to business once again, then. Don't fail to inform us about your XONSH compatibility research. Could you add a functional test, please?

hey, so I wrote a test, but i will glad if you help me with a question...

the test right now is not passing and I dont understand why.
first I wrote a test just for checking the path is absolute and the command working (with empty constraint file).

then I want to add a check for installing the right constraint from the file.
when I put in the constraint file `pycowsay==0.0.0.1` it does not installing the package (says something about conflict).

and if I run similar command on my terminal its working, so im trying to understand what is the different?
maybe you can help me understand its related to the tests architecture or something like this.

guysalt:
> > Great, let's get down to business once again, then. Don't fail to inform us about your XONSH compatibility research. Could you add a functional test, please?
> 
> hey, so I wrote a test, but i will glad if you help me with a question...
> 
> the test right now is not passing and I dont understand why. first I wrote a test just for checking the path is absolute and the command working (with empty constraint file).
> 
> then I want to add a check for installing the right constraint from the file. when I put in the constraint file `pycowsay==0.0.0.1` it does not installing the package (says something about conflict).
> 
> and if I run similar command on my terminal its working, so im trying to understand what is the different? maybe you can help me understand its related to the tests architecture or something like this.

understand the catch, add `pycowsay==0.0.0.1` to the `tests_packages` files...
need to see all the tests pass

tests/test_install.py
```diff
@@ -264,6 +264,50 @@ def test_pip_args_with_windows_path(pipx_temp_env, capsys):
...
+    temp_file.touch()
+    temp_file.write_text(f"pycowsay=={pycowsay_version}")
+
+    assert run_pipx_cli(["install", f"--pip-args='{constraint_flag}{constraint_file_name}'", "pycowsay"]) == 0
+
+    assert f"{constraint_flag}{temp_file}" in caplog.text
+
+    subprocess_pycowsay_version = subprocess.run(["pycowsay", "--version"], capture_output=True, text=True, check=False)
+    subprocess_pycowsay_version_output = subprocess_pycowsay_version.stdout.strip()
+    assert subprocess_pycowsay_version_output == pycowsay_version
```
chrysle, line 295: I'd rather use `importlib.metadata` than run a subprocess.

tests/test_install.py
```diff
@@ -264,6 +264,50 @@ def test_pip_args_with_windows_path(pipx_temp_env, capsys):
...
+    output_lines = subprocess_pycowsay_available_versions_output.split("\n")
+    available_versions_line = next(line for line in output_lines if "Available versions:" in line)
+    versions_string = available_versions_line.split(":")[1].strip()
+    available_versions = versions_string.split(", ")
+    assert pycowsay_version in available_versions
+
+    os.chdir(tmp_path)
+    current_directory = Path(".").expanduser().resolve()
+
+    temp_file = current_directory / constraint_file_name
```
chrysle, line 285:
```suggestion
    constraints_file = current_directory / constraint_file_name
```

I'd name this a bit more meaningful, in case you have to debug the test output.

tests/test_install.py
```diff
@@ -264,6 +264,50 @@ def test_pip_args_with_windows_path(pipx_temp_env, capsys):
...
+    temp_file.touch()
+    temp_file.write_text(f"pycowsay=={pycowsay_version}")
```
chrysle, line 287: We have a small database for our test packages, which we use in tests.

```suggestion
    temp_file.write_text(PKG["pycowsay"]["spec"])
```

testdata/tests_packages/unix-python3.10.txt
```diff
@@ -148,6 +148,7 @@ pure_eval==0.2.2
 py==1.11.0
 pyasn1==0.6.0
 pycowsay==0.0.0.2
+pycowsay==0.0.0.1
```
chrysle, line 151: Why add _0.0.1_ to the packages, when _0.0.2_ is already present?

tests/test_install.py
```diff
@@ -264,6 +264,50 @@ def test_pip_args_with_windows_path(pipx_temp_env, capsys):
...
+    )
+    subprocess_pycowsay_available_versions_output = subprocess_pycowsay_available_versions.stdout
+    output_lines = subprocess_pycowsay_available_versions_output.split("\n")
+    available_versions_line = next(line for line in output_lines if "Available versions:" in line)
+    versions_string = available_versions_line.split(":")[1].strip()
+    available_versions = versions_string.split(", ")
+    assert pycowsay_version in available_versions
+
+    os.chdir(tmp_path)
+    current_directory = Path(".").expanduser().resolve()
```
chrysle, line 283: I don't think we need this. We can also access a temporary file via its full path:

```python
constraints_file = tmp_path / constraint_file_name
constraints_file.touch()
```

chrysle (approved):
Thanks for adding the tests.

tests/test_install.py
guysalt: sure done

tests/test_install.py
```diff
@@ -264,6 +264,50 @@ def test_pip_args_with_windows_path(pipx_temp_env, capsys):
...
+
+    subprocess_pycowsay_available_versions = subprocess.run(
+        [sys.executable, "-m", "pip", "index", "versions", "pycowsay"], capture_output=True, text=True, check=False
+    )
+    subprocess_pycowsay_available_versions_output = subprocess_pycowsay_available_versions.stdout
+    output_lines = subprocess_pycowsay_available_versions_output.split("\n")
+    available_versions_line = next(line for line in output_lines if "Available versions:" in line)
+    versions_string = available_versions_line.split(":")[1].strip()
+    available_versions = versions_string.split(", ")
+    assert pycowsay_version in available_versions
```
chrysle, line 280: Why do we need this? Besides, `pip index` is still experimental.

testdata/tests_packages/unix-python3.10.txt
guysalt: I thought `pycowsay` its the default package for testing and example of pipx.

I can change to `nox` but somehow with subprocess I get the output on `stderr` and not `stdout`...
And for `pylint` its more complicated as the output contains multiline check.

I can change to whatever you think its better just tell me :)

chrysle: > I thought `pycowsay` its the default package for testing and example of pipx.

Right, but I think we should make use of available ressources whenever possible.

> I can change to `nox` but somehow with subprocess I get the output on `stderr` and not `stdout`... And for `pylint` its more complicated as the output contains multiline check.

Confirmed, but that isn't a blocker, is it? Besides, have a look at our test packages lists, you could also use `tox` or something else.

src/pipx/package_specifier.py
```diff
@@ -161,7 +161,27 @@ def parse_specifier_for_install(package_spec: str, pip_args: List[str]) -> Tuple
         )
         pip_args.remove("--editable")
 
-    return (package_or_url, pip_args)
+    for index, option in enumerate(pip_args):
+        if not option.startswith(("-c", "--constraint")):
+            continue
+
+        if option == "-c" or option == "--constraint":
```
chrysle, line 168:
```suggestion
        if option in ("-c", "--constraint"):
```

src/pipx/package_specifier.py
```diff
@@ -161,7 +161,27 @@ def parse_specifier_for_install(package_spec: str, pip_args: List[str]) -> Tuple
...
+            next_option_index = index + 1
+            if next_option_index < len(pip_args):
+                next_option = pip_args[next_option_index]
```
chrysle, line 171: Let's rename this and also `next_option_index`.
```suggestion
                constraints_file = pip_args[argument_index]
```

testdata/tests_packages/unix-python3.10.txt
guysalt: done, use `ipython` package

tests/test_install.py
guysalt: Right, done.

tests/test_install.py
guysalt: For double check if the required package version is available and if not, will be more clearly to say why the test fails.

do you prefer to remove this block?

tests/test_install.py
```diff
@@ -264,6 +264,50 @@ def test_pip_args_with_windows_path(pipx_temp_env, capsys):
...
+
+    os.chdir(tmp_path)
+    current_directory = Path(".").expanduser().resolve()
+
+    constraints_file = current_directory / constraint_file_name
+    constraints_file.touch()
+    constraints_file.write_text(f"pycowsay=={pycowsay_version}")
+
+    assert run_pipx_cli(["install", f"--pip-args='{constraint_flag}{constraint_file_name}'", "pycowsay"]) == 0
```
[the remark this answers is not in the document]
guysalt: Don't you think the code will be less readable and less specific like this (specific with `== 1`)?

I think its better with the specific number written like that.

src/pipx/package_specifier.py
guysalt: haha fool me :sweat_smile:

done.

src/pipx/package_specifier.py
guysalt: done

tests/test_install.py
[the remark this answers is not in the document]
chrysle: It's only that the rest of the code base uses this form.

tests/test_install.py
```diff
@@ -264,6 +264,49 @@ def test_pip_args_with_windows_path(pipx_temp_env, capsys):
...
+        [sys.executable, "-m", "pip", "index", "versions", package_name], capture_output=True, text=True, check=False
+    )
+    subprocess_package_available_versions_output = subprocess_package_available_versions.stdout
+    output_lines = subprocess_package_available_versions_output.split("\n")
+    available_versions_line = next(line for line in output_lines if "Available versions:" in line)
+    versions_string = available_versions_line.split(":")[1].strip()
+    available_versions = versions_string.split(", ")
+    assert package_version in available_versions
+
+    os.chdir(tmp_path)
```
chrysle, line 283: This isn't necessary, now, or am I missing something?

```suggestion
```

tests/test_install.py
```diff
@@ -264,6 +264,49 @@ def test_pip_args_with_windows_path(pipx_temp_env, capsys):
     assert r"D:\\TEST\\DIR" in captured.err
 
 
+@pytest.mark.parametrize("constraint_flag", ["-c ", "--constraint ", "--constraint="])
+def test_pip_args_with_constraint_relative_path(constraint_flag, pipx_temp_env, tmp_path, caplog):
+    constraint_file_name = "constraint.txt"
```
chrysle, line 269: This is the preferred name, AFAIK.

```suggestion
    constraint_file_name = "constraints.txt"
```

tests/test_install.py
chrysle: @guysalt I think you overlooked this comment?

tests/test_install.py
chrysle: Hmm, if it wasn't available, there would be something wrong with our test packages setup. I'd tend not to over-complicate things.

tests/test_install.py
guysalt: Also here, if someone change to the newer version the test will not check if the older versions indeed installed.

**Changed**

tests/test_install.py
guysalt: Ok, just wanted to check the setup for the tests good, and we have two different versions available on the tests and see the constraint flag works and the older version is installed.

Now after I wrote this I also see I did not check there at lease newer version indeed...

Anyway, Im ok also with not checking this,

**Removed.**

tests/test_install.py
guysalt: sure its necessary, this is all the test..

we going to the dir of the file and give in the constraint flag the local path `constraint_file_name` and not the absolute path `constraints_file`

without this the test does not works as the installation fails on non-existing constraint file.

tests/test_install.py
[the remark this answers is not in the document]
guysalt: done

tests/test_install.py
```diff
@@ -264,6 +264,50 @@ def test_pip_args_with_windows_path(pipx_temp_env, capsys):
...
+    subprocess_pycowsay_version_output = subprocess_pycowsay_version.stdout.strip()
+    assert subprocess_pycowsay_version_output == pycowsay_version
+
+
+@pytest.mark.parametrize("constraint_flag", ["-c ", "--constraint ", "--constraint="])
+def test_pip_args_with_wrong_constraint_fail(constraint_flag, pipx_ultra_temp_env, tmp_path, capsys):
+    constraint_file_name = "constraint.txt"
+    os.chdir(tmp_path)
+
+    assert run_pipx_cli(["install", f"--pip-args='{constraint_flag}{constraint_file_name}'", "pycowsay"]) == 1
```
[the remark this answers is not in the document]
guysalt: done

tests/test_install.py
guysalt: haha ok, the flag on `pip` is  `constraint`, but it does not matter to me.

Changed

tests/test_install.py
chrysle: > Ok, just wanted to check the setup for the tests good, and we have two different versions available on the tests and see the constraint flag works and the older version is installed.

Although I think we can trust `pip` it handles the constraints files it is given correctly (if it does find them), creating a constraints file without a proper constraint doesn't seem good. Could you try specifying a version not equal (`!=`) to the spec? That way, we should be able to determine whether the mechanism works.

guysalt: Im not sure I understand you correctly, Im kind of confusing right now.

Because firstly I have take a const version string, and then changed it to the spec for `PKG[package_name]["spec"]` as you wish.

can you tell me exactly what you miss right now or how do you think it should be?

thank you :)

guysalt: Also, I can change it to be like this:

1. first installing the package without constraint flag
2. then checking the `PKG[package_name]["spec"]` is not the same version as the installed version already (another subprocess 😪).
3. then installing the package with the constraint flag and see if the version changed.

This is another approach for checking the constraint flag is working properlly.

chrysle: I mean, if the have two or more versions of the same package available for installation and specify in the constraints file that one of the versions should *not* be selected, it will logically take  another – thus we can verify specifying a constraint in a file works, by checking in the end that the version that was given as part of the spec was *not* installed.

guysalt: ok now I got you, to do so with the `PKG[package_name]["spec"]` or with const?

I prefer const as the `PKG[package_name]["spec"]` taking the older version, and with const I can put not equal `!=` to the newer version.

chrysle: > I prefer const as the `PKG[package_name]["spec"]` taking the older version, and with const I can put not equal `!=` to the newer version.

You could run a replace on that, but either way is fine.

guysalt: Now I see the newer version isn't the same in all the `tests_packages` constraint files 😪 ...

How do I run a replace? Can I change it manually?

What do you think about put in the constraint file `<` from `PKG[package_name]["spec"]` and then check the version is indeed less then the `PKG[package_name]["spec"]`.

For this I need that every test constraint file will contain at lease one version less the `PKG[package_name]["spec"]` version.

chrysle: > How do I run a replace? Can I change it manually?

I thought of just using the `str` types' `replace()` function?

As far as I'm aware, we don't even need to test whether the version after installing with constraints is less than the spec given; only if it is not equal? Sorry for the hassle here.

guysalt: Hey im sorry for the delay, yea im sorry for the hassle too lets finish it.

I wanted to check if it less than the spec given, to validate for sure the constraint file is working. as I mention before, even if I using not equal sign and the constraint file somehow does not working, the test still can pass if the pypi versions matching the constraint.

Anyway, if its ok for you its ok for me, so now I believe I did it as you asked for.

thanks for your time!

guysalt:
Its seems that the tests fails exactly on what I was trying to tell you on the pip args split...

that the splitting does now working on the workflow, as it does now working on my terminal (xonsh).

do you have a clue why does it happening?

guysalt:
Hey again @chrysle 

I was a little bit busy, I would like to keep going and finish this please.

We discussed about that thing with the `'` on the first thread here.
Now, after I add the last commit that strip the `'` char, all the tests are working...

Please review it, and let me know if you still don't want the striping on the pip args.
Thanks

epx5k (approved):

chrysle:
Looks good. Please add a comment to the code explaining why it is necessary, then I think this is mergeable.

guysalt:
> Looks good. Please add a comment to the code explaining why it is necessary, then I think this is mergeable.

Thanks, done.

chrysle (approved):

Gitznik (approved):

Gitznik (approved):

chrysle:
Thanks!

PR #1397, opened 2024-05-11
Move --global option into shared parser

dukecat0:

## Summary of changes
Move `--global` option into shared parser, such that it can be passed after the subcommand, for example `pipx ensurepath --global`.

## Test plan

Tested by running

```
pipx install --global nox
pipx ensurepath --global
```


chrysle (approved):

chrysle (approved):

Paiusco:
Dumb question:
will it still accept `pipx --global ensurepath`?

If not, then `--help` is misleading now IMO, if both positions are being considered, then forget what I've said

dukecat0:
Yes, both positions are accepted.

PR #1414, opened 2024-05-18
Apply ruff rule RUF012

### LINKED ISSUE #1408: Look into ruff rule RUF012

[`RUF012`](https://docs.astral.sh/ruff/rules/mutable-class-default/) checks for mutable default values in class attributes:
> Mutable default values share state across all instances of the class, while not being obvious. This can lead to bugs when the attributes are changed in one instance, as those changes will unexpectedly affect all other instances.

```console
$ ruff check --extend-select RUF --ignore RUF005
src/pipx/animate.py:24:20: RUF012 Mutable class attributes should be annotated with `typing.ClassVar`
src/pipx/pipx_metadata_file.py:40:28: RUF012 Mutable class attributes should be annotated with `typing.ClassVar`
src/pipx/pipx_metadata_file.py:41:29: RUF012 Mutable class attributes should be annotated with `typing.ClassVar`
src/pipx/pipx_metadata_file.py:42:44: RUF012 Mutable class attributes should be annotated with `typing.ClassVar`
src/pipx/pipx_metadata_file.py:43:56: RUF012 Mutable class attributes should be annotated with `typing.ClassVar`
Found 5 errors.
$ 
```

Introduced in #610:
https://github.com/pypa/pipx/blob/5eed47950ba97ea8e0ad292a1026cf87809f940e/src/pipx/animate.py#L23-L24

Introduced in #222:
https://github.com/pypa/pipx/blob/5eed47950ba97ea8e0ad292a1026cf87809f940e/src/pipx/pipx_metadata_file.py#L29-L44

As a first step, I'd like to understand whether this is on purpose or not.

I suspect the relevant classes are designed to have one instance, in which case this is not an issue in practice. Nevertheless, it's good practice to avoid mutable default values, so that future maintainers don't have to ask themselves again.

DimitriPapadopoulos:

## Summary of changes

Apply ruff rule `RUF012`
```
RUF012 Mutable class attributes should be annotated with `typing.ClassVar`
```

Gitr rid of mutable default values:
* `list` → `tuple`
* `list` → `field(default_factory=list)` (using [`dataclass`](https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass) instead of [`NamedTuple`](https://docs.python.org/3/library/typing.html#typing.NamedTuple))
* `dict` → `field(default_factory=dict)`(using [`dataclass`](https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass) instead of [`NamedTuple`](https://docs.python.org/3/library/typing.html#typing.NamedTuple))

Fixes #1408.

## Test plan

Tested by running
```console
$ ruff check --extend-select RUF012
All checks passed!
$ 
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -76,6 +76,7 @@ lint.extend-select = [
   "PLE",
   "PLW",
   "RSE",
+  "RUF012",
   "RUF100",
   "W",
 ]
--- a/src/pipx/animate.py
+++ b/src/pipx/animate.py
@@ -21,7 +21,7 @@ if WINDOWS:
     import ctypes
 
     class _CursorInfo(ctypes.Structure):
-        _fields_ = [("size", ctypes.c_int), ("visible", ctypes.c_byte)]
+        _fields_ = (("size", ctypes.c_int), ("visible", ctypes.c_byte))
 
 
 def _env_supports_animation() -> bool:
--- a/src/pipx/pipx_metadata_file.py
+++ b/src/pipx/pipx_metadata_file.py
@@ -1,7 +1,8 @@
 import json
 import logging
+from dataclasses import asdict, dataclass, field
 from pathlib import Path
-from typing import Any, Dict, List, NamedTuple, Optional, Union
+from typing import Any, Dict, List, Optional, Union
 
 from pipx.emojis import hazard
 from pipx.util import PipxError, pipx_wrap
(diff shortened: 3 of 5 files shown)


DimitriPapadopoulos:
Other classes inheriting `NamedTuple` can be changed to `dataclass` in a different PR.

chrysle (approved):

Issue #1429, opened 2024-05-28
`sudo pipx ensurepath --global` fails on MacOs due to --global flag order


openrory:
**Describe the bug**

When running `sudo pipx ensurepath --global` as part of the installation process for MacOS, it gives the following error: 
```
usage: pipx [-h] [--quiet] [--verbose] [--global] [--version]
            {install,uninject,inject,upgrade,upgrade-all,uninstall,uninstall-all,reinstall,reinstall-all,list,interpreter,run,runpip,ensurepath,environment,completions}
            ...
pipx: error: unrecognized arguments: --global
```

**How to reproduce**

run `sudo pipx ensurepath --global`

**Expected behavior**

```
/usr/local/bin is already in PATH.

⚠️  All pipx binary directories have been added to PATH. If you are sure you want to proceed, try again with the '--force'
flag.
```

Gitznik:
The fix for this has already been implemented in https://github.com/pypa/pipx/pull/1397, and will be in the next release.

dukecat0:
Probably it's time to create a new release?

Gitznik:
> Probably it's time to create a new release?

Whoever does this must fix the news fragments though. Multiple lines are not supported and break the formatting (see the current latest changelog). I can do it in the evening.

openrory:
Even better. I see you’ve already closed the issue, thanks! Looking forward to the new release ;-)On 28 May 2024, at 16:03, Robert ***@***.***> wrote:﻿
The fix for this has already been implemented in #1397, and will be in the next release.

PR #1430, opened 2024-05-28
fix order of --global flag for MacOS

openrory:

## Summary of changes
This fixes issue #1429, the last step in the process of installing pipx for MacOS where it fails due to the `--global` flag being in the wrong order. 

## Test plan

1. uninstall pipx
2. install pipx in MacOS
3. use `sudo pipx --global ensurepath` instead of `sudo pipx ensurepath --global`

Tested by running

```
sudo pipx --global ensurepath
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ _For comparison to other tools including pipsi, see
 ```
 brew install pipx
 pipx ensurepath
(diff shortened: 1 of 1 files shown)


chrysle:
Thanks, but this was already fixed in #1397, which makes the current flag order possible.

Issue #1400, opened 2024-05-12
Add ruff rule RUF100


DimitriPapadopoulos:
**How would this feature be useful?**

[`RUF100`](https://docs.astral.sh/ruff/rules/unused-noqa/) detects unused **`noqa`** directives.

```console
$ ruff check --extend-select RUF100
src/pipx/__main__.py:11:28: RUF100 [*] Unused blanket `noqa` directive
src/pipx/venv_inspect.py:111:37: RUF100 [*] Unused `noqa` directive (unknown: `T484`)
Found 2 errors.
[*] 2 fixable with the `--fix` option.
```

**Describe the solution you'd like**

Fix issues reported by `ruff check --extend-select RUF100` and add `RUF100` to `pyproject.toml`.

**Describe alternatives you've considered**

PR #1431, opened 2024-05-28
Prepare news fragments for 1.6.0

Gitznik:

## Summary of changes
Fixed the news fragments so they no longer break the formatting of the changelog.

Will create a release PR after this is merged.

## Test plan

Tested by running

```
nox -s watch_docs
```

### DIFF
--- a/changelog.d/1316.feature.md
+++ b/changelog.d/1316.feature.md
@@ -1,7 +1,1 @@
-Add `pipx upgrade-shared` command.
-
-This allows to create/upgrade shared libraries as a standalone command.
-
-It might be useful to initialize shared libraries with a pinned version of pip in order to create e.g. reproducible docker images using a constraint file.
-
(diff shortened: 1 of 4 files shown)


Gitznik:
Do we actually want to keep the news fragment for [#1400](https://github.com/pypa/pipx/issues/1400)? I don't think there is any reason to include it in the changelog.

huxuan:
> Do we actually want to keep the news fragment for [#1400](https://github.com/pypa/pipx/issues/1400)? I don't think there is any reason to include it in the changelog.

I agree on removing it.

Gitznik:
> > Do we actually want to keep the news fragment for [#1400](https://github.com/pypa/pipx/issues/1400)? I don't think there is any reason to include it in the changelog.
> 
> I agree on removing it.

Done

huxuan (approved):

PR #1437, opened 2024-06-02
Add quotes to the title in release action

dukecat0:


chrysle:
Whee, it [worked](https://github.com/pypa/pipx/actions/runs/9336567854)! Except for the zipapp, which we'll need to upload manually. It's because the [output](https://github.com/pypa/pipx/blob/273c558fb5f183e16f396dbeeba2a951cfce678a/.github/workflows/release.yml#L107)
comes now from the `create-tag` job (another oversight of mine). Could you amend that?

dukecat0:
>  Could you amend that?

Fixed!

chrysle:
Thanks! Don't we also need to specify the new dependency in the `needs` field?

dukecat0:
Looks like everthing should work now.

dukecat0:
Btw, I removed `pipx.pyz` from release since it doesn't provide the correct version when running `pipx --version`.

chrysle:
> Btw, I removed `pipx.pyz` from release since it doesn't provide the correct version when running `pipx --version`.

Ah, right. Uploaded a correctly generated one. I think we need to copy over the job to generate the zipapp then.

dukecat0:
Or maybe we can generate it by re-running failed jobs:
image

chrysle:
For this time, it's done anyway. For the future, it should be necessary to generate the zipapp directly before uploading to the release – as you pointed out, retrieving it from the preceding tests run results in incorrect version information. Can that be done in this PR?

dukecat0:
> generate the zipapp directly before uploading to the release

Done in the latest commit.

chrysle (approved):

PR #1440, opened 2024-06-04
ci: clean up auto changelog on releases

henryiii:
I noticed on the last release that the auto generated changelog included some bot entries. Those can be cleaned out with a config file like this.

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -3,3 +3,4 @@ changelog:
     authors:
       - dependabot
       - pre-commit-ci
+      - github-actions
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,5 @@
+changelog:
+  exclude:
+    authors:
+      - dependabot
+      - pre-commit-ci


.github/release.yml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  exclude:
+    authors:
+      - dependabot
+      - pre-commit-ci
```
Gitznik, line 5: I reckon we can also add the `github-actions` user

```suggestion
      - pre-commit-ci
      - github-actions
```

Gitznik (changes requested):
good call :+1:

henryiii:
I wonder if the [bot] is part of the name if if it's possible to match on that?

Gitznik (approved):

Gitznik:
> I wonder if the [bot] is part of the name if if it's possible to match on that?

Does not look like it from what I can see in the docs. I think this is fine :+1:

chrysle:
Hmm, I think this might be needed; similar problem at https://github.com/jazzband/pip-tools/pull/1976#issuecomment-1707321760, but we didn't come to test it.

henryiii:
I've been using this as is in a bunch of repos, and it works, you don't need the trailing `[bot]` for sure; it would be nice though if you could trigger on it.

Issue #1333, opened 2024-04-08
`pipx.util` circular import


adamjorr:
Importing from `pipx.util` fails due to circular dependency unless `pipx.paths` is imported first. It seems `pipx.util` became dependent on `pipx.paths` in #1281. `pipx.paths` then tries to import `pipx_wrap` from `pipx.util` and fails.

script to repro:

```py
import pipx.util

if __name__ == "__main__":
    print("Success!")

```

```
Traceback (most recent call last):
  File "/Users/aorr1/Library/Application Support/JetBrains/PyCharm2023.3/scratches/scratch_19.py", line 1, in <module>
    import pipx.util
  File "/Users/aorr1/Library/Caches/pypoetry/virtualenvs/project-dJ3m09UH-py3.11/lib/python3.11/site-packages/pipx/util.py", line 24, in <module>
    from pipx import paths
  File "/Users/aorr1/Library/Caches/pypoetry/virtualenvs/project-dJ3m09UH-py3.11/lib/python3.11/site-packages/pipx/paths.py", line 10, in <module>
    from pipx.util import pipx_wrap
ImportError: cannot import name 'pipx_wrap' from partially initialized module 'pipx.util' (most likely due to a circular import) (/Users/aorr1/Library/Caches/pypoetry/virtualenvs/project-dJ3m09UH-py3.11/lib/python3.11/site-packages/pipx/util.py)

```

**Expected behavior**

import succeeds

JamesParrott:
A circular import's not ideal, but removing it probably requires refactoring.  

In the mean time, is this work around not suitable?

```
import pipx
import pipx.util
```

adamjorr:
I think `import pipx` still runs into the issue. `import pipx.paths` before `import pipx.util` does help.

PR #1446, opened 2024-06-10
Introduce a new env variable to allow spaces in home path

### LINKED ISSUE #1320: Improve docs for resolving `space in home path` warning

The existing warning and docs are not enough for users to resolve it.

Dedicated docs on how to move the home path and a clear link to it in the warning, as well as specifying it's the `Pipx Home` that has the space in it.

Hope people that are confused just find the existing discussions.

### Discussed in https://github.com/pypa/pipx/discussions/1315

<div type='discussions-op-text'>

<sup>Originally posted by **homocomputeris** April  1, 2024</sup>
I get the message every time I run pipx.

- What is home path here?
- Where are the docs on this issue? Nothing can be found here: https://www.google.com/search?hl=en&q=%22home%20path%22%20site%3Ahttps%3A%2F%2Fpipx.pypa.io%2Flatest%2F
- How to fix or disable this warning?</div>

Gitznik:

## Summary of changes
Introduces an environment variable to silence the warning for having a space in the pipx home path.

Improves the docs on how to troubleshoot the warning and move the pipx installation location in general. This makes [the discussion](https://github.com/pypa/pipx/discussions/1330) I created a while ago redundant.

Resolves #1320

## Test plan

Tested by running

```
nox -s tests-3.8 -- -k test_environment
```

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -281,9 +281,13 @@ $home_dir = pipx environment --value PIPX_HOME
 
 Remove-Item -Recurse -Force -ErrorAction SilentlyContinue $cache_dir, $logs_dir, $trash_dir
 
-# Remove the destionation directory to ensure rename behavior of `Move-Item`
+# Remove the destination directory to ensure rename behavior of `Move-Item`
 Remove-Item -Recurse -Force -ErrorAction SilentlyContinue $NEW_LOCATION
 
 Move-Item -Path $home_dir -Destination $NEW_LOCATION
 pipx reinstall-all
 ```
+
+If you would prefer doing it in bash via git-bash/WSL, feel free to use
+the MacOS/Linux instructions, changing the `$NEW_LOCATION` to the Windows
+version.
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -270,7 +270,7 @@ pipx reinstall-all
 
 Requires bash (for example `git bash`)
 
-Current default location: `~`
+Current default location: `~/pipx`
 
 ```bash
 NEW_LOCATION=~
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -261,8 +261,8 @@ if [ -n "$XDG_DATA_HOME" ]; then
 else
     NEW_LOCATION="$HOME/.local/share"
 fi
-rm -rf $cache_dir $logs_dir $trash_dir
-mkdir -p $NEW_LOCATION && mv $home_dir $NEW_LOCATION
+rm -rf "$cache_dir" "$logs_dir" "$trash_dir"
+mkdir -p $NEW_LOCATION && mv "$home_dir" $NEW_LOCATION
 pipx reinstall-all
 ```
 
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -288,13 +288,9 @@ cache_dir=$(pipx environment --value PIPX_VENV_CACHEDIR)
 logs_dir=$(pipx environment --value PIPX_LOG_DIR)
 trash_dir=$(pipx environment --value PIPX_TRASH_DIR)
 home_dir=$(pipx environment --value PIPX_HOME)
-# If you wish another location, replace the if block below
+# If you wish another location, replace the expression below
 # and set `NEW_LOCATION` explicitly
-if [ -n "$XDG_DATA_HOME" ]; then
-    NEW_LOCATION="$XDG_DATA_HOME"
-else
-    NEW_LOCATION="$HOME/.local/share"
-fi
+NEW_LOCATION="${XDG_DATA_HOME:-$HOME/.local/share}"
 rm -rf "$cache_dir" "$logs_dir" "$trash_dir"
 mkdir -p $NEW_LOCATION && mv "$home_dir" $NEW_LOCATION
 pipx reinstall-all
@@ -302,7 +298,7 @@ pipx reinstall-all
 
 ### Windows
 
-Requires bash (for example `git bash`)
+Requires bash (for example `git bash`).
 
 Current default location: `~/pipx`
 
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -272,13 +272,18 @@ Requires bash (for example `git bash`)
 
 Current default location: `~/pipx`
 
-```bash
-NEW_LOCATION=~
-cache_dir=$(pipx environment --value PIPX_VENV_CACHEDIR)
-logs_dir=$(pipx environment --value PIPX_LOG_DIR)
-trash_dir=$(pipx environment --value PIPX_TRASH_DIR)
-home_dir=$(pipx environment --value PIPX_HOME)
-rm -rf $cache_dir $logs_dir $trash_dir
-mv $home_dir $NEW_LOCATION
+```powershell
+$NEW_LOCATION = Join-Path $HOME 'pipx'
+$cache_dir = pipx environment --value PIPX_VENV_CACHEDIR
+$logs_dir = pipx environment --value PIPX_LOG_DIR
+$trash_dir = pipx environment --value PIPX_TRASH_DIR
+$home_dir = pipx environment --value PIPX_HOME
+
+Remove-Item -Recurse -Force -ErrorAction SilentlyContinue $cache_dir, $logs_dir, $trash_dir
+
+# Remove the destionation directory to ensure rename behavior of `Move-Item`
+Remove-Item -Recurse -Force -ErrorAction SilentlyContinue $NEW_LOCATION
+
+Move-Item -Path $home_dir -Destination $NEW_LOCATION
 pipx reinstall-all
 ```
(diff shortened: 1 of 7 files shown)


huxuan:
> @huxuan if you have the time, could you please verify that the windows example for moving the pipx install location actually works? Or maybe even translate it to powershell?

Gotcha, will take a look later.

docs/installation.md
```diff
@@ -223,3 +223,62 @@ You can easily get your shell's tab completions working by following instruction
...
+```bash
+NEW_LOCATION=~
+cache_dir=$(pipx environment --value PIPX_VENV_CACHEDIR)
+logs_dir=$(pipx environment --value PIPX_LOG_DIR)
+trash_dir=$(pipx environment --value PIPX_TRASH_DIR)
+home_dir=$(pipx environment --value PIPX_HOME)
+rm -rf $cache_dir $logs_dir $trash_dir
+mv $home_dir $NEW_LOCATION
+pipx reinstall-all
+```
```
huxuan, line 284:
````suggestion
```powershell
$NEW_LOCATION = Join-Path $HOME '.local/pipx'
$cache_dir = pipx environment --value PIPX_VENV_CACHEDIR
$logs_dir = pipx environment --value PIPX_LOG_DIR
$trash_dir = pipx environment --value PIPX_TRASH_DIR
$home_dir = pipx environment --value PIPX_HOME

Remove-Item -Recurse -Force -ErrorAction SilentlyContinue $cache_dir, $logs_dir, $trash_dir

Move-Item -Path $home_dir -Destination $NEW_LOCATION
pipx reinstall-all
```
````

BTW, the process works fine on my PC.

File: docs/installation.md [hunk 2]
```diff
@@ -223,3 +223,62 @@ You can easily get your shell's tab completions working by following instruction
...
+rm -rf $cache_dir $logs_dir $trash_dir
+mkdir -p $NEW_LOCATION && mv $home_dir $NEW_LOCATION
+pipx reinstall-all
+```
+
+### Windows
+
+Requires bash (for example `git bash`)
+
+Current default location: `~`
```
huxuan, line 273:
```suggestion
Current default location: `~/pipx`
```

According to the implementation here [1], the default location should be "~/pipx".

[1] https://github.com/pypa/pipx/blob/dc9f9a849a918c68bc2cc99c72651022371fcf9f/src/pipx/paths.py#L16

huxuan:
> Could you please verify that the windows example for moving the pipx install location actually works? Or maybe even translate it to powershell?

Verified and translated! PTAL

docs/installation.md
Gitznik: In powershell, the `Move-Item` command moves the files in the specified paths into the `-Destination` path?

huxuan: Good catch! I asked ChatGPT and tested on my PC, it seems to depend on the existence of destination directory.

If the destination directory exists, the behavior is just like bash. When the destionation does not exists, it works more like a rename. That is why it works on my side previously.

So maybe we should remove the destionation directory before `Move-Item` to ensure the rename behavior. WDYT?

BTW, I updated the suggested changes accordingly.

Gitznik: Gotcha, yeah nice this looks good to me. Thanks for helping me out here, I'm useless when it comes to windows ;)

huxuan (dismissed):

Gitznik:
Sorry had to add just one small change to make sure spaces in the paths are not a problem when using bash.

huxuan:
> Sorry had to add just one small change to make sure spaces in the paths are not a problem when using bash.

looks like the double quotes are also needed for the dirs and $NEW_LOCATION in powershell.

huxuan (dismissed):

docs/installation.md
```diff
@@ -257,3 +257,71 @@ You can easily get your shell's tab completions working by following instruction
...
+logs_dir=$(pipx environment --value PIPX_LOG_DIR)
+trash_dir=$(pipx environment --value PIPX_TRASH_DIR)
+home_dir=$(pipx environment --value PIPX_HOME)
+# If you wish another location, replace the if block below
+# and set `NEW_LOCATION` explicitly
+if [ -n "$XDG_DATA_HOME" ]; then
+    NEW_LOCATION="$XDG_DATA_HOME"
+else
+    NEW_LOCATION="$HOME/.local/share"
+fi
```
chrysle, line 297: We might be able to simplify this:

```suggestion
# If you wish another location, replace the expression below
# and set `NEW_LOCATION` explicitly
NEW_LOCATION="${XDG_DATA_HOME:-$HOME/.local/share}"
```

docs/installation.md
```diff
@@ -257,3 +257,71 @@ You can easily get your shell's tab completions working by following instruction
...
+rm -rf "$cache_dir" "$logs_dir" "$trash_dir"
+mkdir -p $NEW_LOCATION && mv "$home_dir" $NEW_LOCATION
+pipx reinstall-all
+```
+
+### Windows
+
+Requires bash (for example `git bash`)
```
chrysle, line 305:
```suggestion
Requires bash (for example `git bash`).
```

docs/installation.md
```diff
@@ -257,3 +257,71 @@ You can easily get your shell's tab completions working by following instruction
...
+Remove-Item -Recurse -Force -ErrorAction SilentlyContinue "$cache_dir", "$logs_dir", "$trash_dir"
+
+# Remove the destination directory to ensure rename behavior of `Move-Item`
+Remove-Item -Recurse -Force -ErrorAction SilentlyContinue "$NEW_LOCATION"
+
+Move-Item -Path $home_dir -Destination "$NEW_LOCATION"
+pipx reinstall-all
+```
+
+If you would prefer doing it in bash via git-bash/WSL, feel free to use
```
chrysle, line 325: Don't the instructions say that git-bash is required, above?

docs/troubleshooting.md
```diff
@@ -148,13 +148,28 @@ In Pipx version 1.5.0, this was reverted for Windows and MacOS. It defaults agai
...
+automatically) and then reinstall all packages.
 
-```
-rm -rf ~/.local/pipx/{.cache,logs,trash}
-mkdir -p ~/.local/share && mv ~/.local/pipx ~/.local/share/
-pipx reinstall-all
-```
+Please refer to [Installation](installation.md#moving-your-pipx-installation) on how to move it.
+
+## Warning: Found a space in the pipx home path
```
chrysle, line 155: Since this is documented here, now, we should replace the link pointing to the GH discussion at [this paragraph](https://pipx--1446.org.readthedocs.build/1446/installation/#installation-options).

src/pipx/animate.py
```diff
@@ -4,8 +4,8 @@
 from threading import Event, Thread
 from typing import Generator, List
 
+import pipx.emojis as emojis
```
chrysle, line 7: This change seems unrelated... I prefer to not mix changes, as it gets unclear why they were introduced. Also, is it necessary?

src/pipx/emojis.py
```diff
@@ -1,15 +1,7 @@
 import os
 import sys
 
-
-def strtobool(val: str) -> bool:
```
chrysle, line 5: Same for this change. Could you bundle them in a separate PR?

docs/troubleshooting.md
```diff
@@ -148,13 +148,28 @@ In Pipx version 1.5.0, this was reverted for Windows and MacOS. It defaults agai
...
+
+If you want to use a script installed via pipx in a shebang itself (common for example for the aws cli), you run into a similar problem, as the path to the installed script will contain a space.
+
+### How to fix
+
+You can generally fix this by using our default locations, as long as your `$HOME` path does not contain spaces.
+Please refer to our [Installation](installation.md#moving-your-pipx-installation) docs on how to move the `pipx` installation.
 
-For moving the paths back after 1.5.0, you can perform the same steps, switching the paths around.
+If you're really sure you want to stick to your path with spaces, you can set the `PIPX_HOME_ALLOW_SPACE` environment variable to `true`.
```
chrysle, line 175:
```suggestion
If you're really sure you want to stick to your path with spaces, to suppress the warning set the `PIPX_HOME_ALLOW_SPACE` environment variable to `true`.
```

src/pipx/emojis.py
Gitznik: It felt wrong using `strtobool` from `emojis.py` in `paths.py` for something completely emoji unrelated, so I moved it to utils. This required the import changes to fix a circular import.
Do you think I should stick to importing it from emojis for this PR?

chrysle: Probably, yes; this and the circular import problem mentioned in #1333 could be resolved in one go.

docs/installation.md
Gitznik: This was a leftover from what I copied from the discussion. Removed it.

docs/troubleshooting.md
Gitznik: 100%

src/pipx/emojis.py
Gitznik: Removed it from this PR

chrysle (approved):

PR #1451, opened 2024-06-13
Add a `--prepend` option to the `pipx ensurepath` command

### LINKED ISSUE #1424: pipx ensurepath appends to PATH. Option to prepend ?

`pipx ensurepath` appends the pipx path to the user paths (https://github.com/pypa/pipx/blob/main/src/pipx/commands/ensure_path.py#L64). I understand that it is a reasonable default. 

However, if one wants to hide a system installed command with a pipx installed one, the pipx path needs to be prepended to `$PATH`. For example, if an old version of Mercurial is installed (`/usr/bin/hg`) and I want to use a newer Mercurial managed with pipx.

There could be an option to ensurepath so that the pipx path would be preprended. Something like:

```
pipx ensurepath --prepend
```

Note that there is `userpath.prepend` similar to `userpath.append` so the implementation would be straightforward.

agriyakhetarpal:

## Summary of changes

Closes #1424.

This PR adds an option to `pipx ensurepath`, i.e., `pipx ensurepath --prepend` to allow users to prepend `pipx`-installed applications to the `PATH` environment variable rather than appending to it. This is disabled and set to `False` by default; it is only enabled if the user does `pipx ensurepath --prepend` (or its `--force` counterparts such as `pipx ensurepath --force --prepend` or `pipx ensurepath --prepend --force`). The implementation is through `userpath.prepend`, which is similar to `userpath.append` (see https://github.com/ofek/userpath#api).

## Test plan

Tested by running

```bash
pipx ensurepath --prepend  # or pipx ensurepath --prepend --force if already in PATH
```

This modifies my `~/.zshrc` file on macOS accordingly:

```zsh
# Created by `pipx` on 2024-05-20 14:27:49
export PATH="$PATH:/Users/agriyakhetarpal/.local/bin"

# Created by `pipx` on 2024-06-13 14:24:36
export PATH="/Users/agriyakhetarpal/.local/bin:$PATH"
```

Therefore, the latter entry from these changes has prepended `pipx`'s installation location to `PATH`. The earlier entry is from an older `pipx` installation I had that had appended the location. I'm happy to add additional tests in an appropriate place in `tests/` as necessary (though I notice that `ensurepath` doesn't have tests yet), and update the documentation (not sure where a mention of this would be best appropriate to add, though).

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -19,8 +19,19 @@ pipx works on macOS, linux, and Windows.
 ```
 brew install pipx
 pipx ensurepath
-sudo pipx ensurepath --global # optional to allow pipx actions in global scope. See "Customising your installation" section below.
-sudo pipx ensurepath --prepend # optional to prepend the pipx bin directory to PATH instead of appending it. See "Customising your installation" section below.
+```
+
+#### Additional (optional) commands
+
+To allow pipx actions in global scope. See "Customising your installation" section below.
+```
+sudo pipx ensurepath --global
+```
+
+To prepend the pipx bin directory to PATH instead of appending it. See "Customising your installation" section below.
+
+```
+sudo pipx ensurepath --prepend
 ```
 
 ### On Linux:
@@ -31,8 +42,6 @@ sudo pipx ensurepath --prepend # optional to prepend the pipx bin directory to P
 sudo apt update
 sudo apt install pipx
 pipx ensurepath
-sudo pipx ensurepath --global # optional to allow pipx actions in global scope. See "Customising your installation" section below.
-sudo pipx ensurepath --prepend # optional to prepend the pipx bin directory to PATH instead of appending it. See "Customising your installation" section below.
 ```
 
 - Fedora:
@@ -40,8 +49,6 @@ sudo pipx ensurepath --prepend # optional to prepend the pipx bin directory to P
 ```
 sudo dnf install pipx
 pipx ensurepath
-sudo pipx ensurepath --global # optional to allow pipx actions in global scope. See "Customising your installation" section below.
-sudo pipx ensurepath --prepend # optional to prepend the pipx bin directory to PATH instead of appending it. See "Customising your installation" section below.
 ```
 
 - Using `pip` on other distributions:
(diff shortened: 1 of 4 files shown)


Gitznik (changes requested):
Thanks for this contribution. I think a quick note about this in the installation docs would be nice. Else lgtm.

agriyakhetarpal:
> Thanks for this contribution. I think a quick note about this in the installation docs would be nice. Else lgtm.

Thanks for the review – addressed in 435173f45cd7fb71cbdbfd12f5b42740dfb1d179; I added a note to `docs/installation.md`, renaming the "Global installation" heading to "Customising your installation" as I felt that it was the best place to mention something related to `pipx ensurepath`, similar to the `--global` argument.

docs/installation.md
```diff
@@ -19,7 +19,8 @@ pipx works on macOS, linux, and Windows.
 ```
 brew install pipx
 pipx ensurepath
-sudo pipx ensurepath --global # optional to allow pipx actions in global scope. See "Global installation" section below.
+sudo pipx ensurepath --global # optional to allow pipx actions in global scope. See "Customising your installation" section below.
+sudo pipx ensurepath --prepend # optional to prepend the pipx bin directory to PATH instead of appending it. See "Customising your installation" section below.
```
Gitznik, line 23: I'd prefer having a comment after line 22: `pipx ensurepath # optional with --prepend, to prepend the pipx bin directory to PATH instead of appending it. See "Customising your installation" section below.`, instead of having it as a separate line. 

So far the install instructions can just be executed top to bottom and everything is fine. If we add your line, and the user goes top to bottom, they would get a warning for prepending, as pipx is already in PATH. For a user that does not really read the description and just copies the instructions, this would be confusing IMO.

Same for the other install instructions.

Gitznik (changes requested):

docs/installation.md
agriyakhetarpal: Makes sense. I'll move both commands, `sudo pipx ensurepath --global` and `sudo pipx ensurepath --prepend` to a code block of their own so that they aren't inadvertently copied and run by a user on their system.

agriyakhetarpal: Addressed in 78c28930f9f62b0f26abf63fb2246bad34f8a146.

docs/installation.md
```diff
@@ -19,7 +19,19 @@ pipx works on macOS, linux, and Windows.
...
+To allow pipx actions in global scope. See "Customising your installation" section below.
+```
+sudo pipx ensurepath --global
+```
+
+To prepend the pipx bin directory to PATH instead of appending it. See "Customising your installation" section below.
+
+```
+sudo pipx ensurepath --prepend
 ```
```
Gitznik, line 35: I'd rephrase this to make it more concise, and link to the actual customization documentation:


#### Additional (optional) commands

To allow pipx actions in global scope.
```
sudo pipx ensurepath --global
```

To prepend the pipx bin directory to PATH instead of appending it.

```
sudo pipx ensurepath --prepend
```

For more details, refer to [Customising your installation](installation.md#customising-your-installation).

Gitznik (changes requested):

docs/installation.md
agriyakhetarpal: Makes sense, thanks! Addressed in 6fc30c07dcb7f567e9a827f071e0a92ff31c6833.

Gitznik (approved):

chrysle:
Please rebase.

agriyakhetarpal:
Done, @chrysle

chrysle:
Thanks!

PR #1457, opened 2024-06-18
Proposal: pipx run --inspect script.py

### LINKED ISSUE #1456: Feature request: inspect interactively after `pipx run script.py` like `python -i script.py`
Thank you very much for your awesome work.
`pipx` is a great tool for me to run python scripts with inline script metadata (PEP 723).

I'd like to inspect interactively after `pipx run script.py` like `python -i`.
https://docs.python.org/3/using/cmdline.html#cmdoption-i

```
% pipx run --inspect script.py
>>>

% pipx run --inspect --no-cache script.py
>>>
```

I will send pull request.

I read the source code(1) and tried to set the environment variable `PYTHONINSPECT` as a workaround.
https://docs.python.org/3/using/cmdline.html#envvar-PYTHONINSPECT
It often works fine, but when pipx creates a virtual environment(2), it causes for pipx for hanging.

(1): https://github.com/pypa/pipx/blob/1.6.0/src/pipx/util.py#L377-L390
(2): https://github.com/pypa/pipx/blob/1.6.0/src/pipx/venv.py#L163-L167

```
% PYTHONINSPECT=1 pipx run script.py  # this works
>>>

% PYTHONINSPECT=1 pipx run --no-cache script.py  # Hang
creating virtual environment
```

(FYI) References
I wrote my blog (Sorry, there are Japanese)

* https://nikkie-ftnext.hatenablog.com/entry/pipx-run-with-envvar-pythoninspect
* https://nikkie-ftnext.hatenablog.com/entry/pipx-run-hang-pythoninspect-and-venv-creation

ftnext:
Resolve #1456

## Test plan

Tested by running

```
```

### DIFF
--- a/tests/test_run.py
+++ b/tests/test_run.py
@@ -414,3 +414,13 @@ def test_run_local_path_entry_point(pipx_temp_env, caplog, root):
     run_pipx_cli_exit(["run", empty_project_path])
 
     assert "Using discovered entry point for 'pipx run'" in caplog.text
+
+
+@mock.patch("pipx.commands.run.exec_app")
+def test_run_inspect(exec_app, tmp_path):
+    script = tmp_path / "test.py"
+    script.write_text("a = 1 + 1")
+
+    run_pipx_cli(["run", "--inspect", str(script)])
+
+    exec_app.assert_called_once_with([mock.ANY, "-i", script])
--- a/src/pipx/commands/run.py
+++ b/src/pipx/commands/run.py
@@ -78,6 +78,7 @@ def run_script(
     venv_args: List[str],
     verbose: bool,
     use_cache: bool,
+    inspect: bool,
 ) -> NoReturn:
     requirements = _get_requirements_from_script(content)
     if requirements is None:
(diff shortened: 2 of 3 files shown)


tests/test_run.py
```diff
@@ -414,3 +414,13 @@ def test_run_local_path_entry_point(pipx_temp_env, caplog, root):
     run_pipx_cli_exit(["run", empty_project_path])
 
     assert "Using discovered entry point for 'pipx run'" in caplog.text
+
+
+@mock.patch("pipx.commands.run.exec_app")
+def test_run_inspect(exec_app, tmp_path):
```
ftnext, line 420: I think it is difficult to use [execvpe_mock](https://github.com/pypa/pipx/blob/1.6.0/tests/test_run.py#L26) in this (`python -i`) case.

I will appreciate your comment about this test implementation.

uranusjr: It’d be quite difficult to actually test this. I would be satisfied if this just mock out the actual process invocation and just chheck the right flags are passed to the child process.

gaborbernat:
Stalled, closing, if you pick it up again we can reopen.

PR #1458, opened 2024-06-23
Proposal: Add `--upgrade` argument for install command

jdknight:

## Proposal

The following is a proposal to support an upgrade option with pipx's install command. A practice I use for documenting various Python-based utilities is to have an installation section which has a single command to either install or upgrade a package:

```
pip install -U example
```

Now trying to promote the use of `pipx` over `pip`, this does not seem to be possible in a single command unless using the `--force` argument:

```
pipx install --force example
```

Using the "force" argument does not feel right, and the intention is not to have packages be re-installed if they are already in a good/upgraded state. Alternatively, I could update documentation to have two paragraphs to identify a method to install and a method to upgrade, but I would rather promote a single command that could do both.

## Summary of changes

When installing a package, both `pip install` and `pipx install` allow a new package to be installed. If a target package is already installed, no changes are made. Both pip and pipx also support a forced reinstall using `--force-reinstall` and `--force` respectively. Another popular installation more for pip is using the `--upgrade` option, which will only perform a package change if a new version is detected. Such a request is not possible in pipx.

This commit adds support for an upgrade method by adding the `--upgrade` option and forwarding it to pip when a package has been detected as already installed.

## Test plan

Tested by provided unit tests and running:

```
pipx install black==22.8.0
pipx install black
pipx install --upgrade black
```

<details><summary>Example output</summary>
<pre>
(venv) [jdknight@devel venv]$ pipx install black==22.8.0
  installed package black 22.8.0, installed using Python 3.12.4
    - black
    - blackd
done!
(venv) [jdknight@devel venv]$ pipx install black
'black' already seems to be installed. Not modifying existing installation in '.../.local/share/pipx/venvs/black'. Pass '--force' to
force installation.
(venv) [jdknight@devel venv]$ pipx install --upgrade black
  installed package black 24.4.2, installed using Python 3.12.4
    - black
    - blackd
done!
</pre>
</details>

### DIFF
--- a/src/pipx/commands/install.py
+++ b/src/pipx/commands/install.py
@@ -29,6 +29,7 @@ def install(
     verbose: bool,
     *,
     force: bool,
+    upgrade: bool,
     reinstall: bool,
     include_dependencies: bool,
     preinstall_packages: Optional[List[str]],
@@ -60,7 +61,7 @@ def install(
 
         venv = Venv(venv_dir, python=python, verbose=verbose)
         venv.check_upgrade_shared_libs(pip_args=pip_args, verbose=verbose)
-        if exists:
+        if exists and not upgrade:
             if not reinstall and force and python_flag_passed:
                 print(
                     pipx_wrap(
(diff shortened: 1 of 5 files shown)


chrysle:
Thanks, but this has already been proposed once (#953) and we agreed elsewhere on an `--install` option to the `--upgrade`  command.

PR #1462, opened 2024-06-26
Use multiple machine suffixes for standalone python

tusharsadhwani:

## Summary of changes

With the release of [Yen v0.5](https://github.com/tusharsadhwani/yen/tree/v0.5.0), `pipx` and `yen` are officially rivals (; But since it's very easy to port my bug fixes to `pipx`, I decided to do just that.

For now, this change only adds 1 new suffix compared to the previous logic: `x86_64-unknown-linux-gnu`, as a user reported this problem: https://github.com/tusharsadhwani/yen/issues/17

`python-build-standalone` ships various tiers of `unkown-linux-gnu` python builds. from `v1` to `v4`. `v4` is highly CPU specific and requires modern CPU features, only available on higher end desktop chips starting from early 2010s.

I defaulted to `v3` for a compromise of good speed, and good compatibility with most machines. But in the more recent releases, `v3` no longer ships `python3.8` wheels. Having a fallback to the `v1` machine suffix would fix this problem.

## Test plan

```
pipx run --python=python3.8 --fetch-missing-python meowsay hi
```

On a Linux machine, this would fail on main branch (as 3.8 is not available in for `x86_64-v3-unknown-linux-gnu`.

But with the current change it will work, as it is available in `x86_64-unknown-linux-gnu` variant.

### DIFF
--- a/src/pipx/standalone_python.py
+++ b/src/pipx/standalone_python.py
@@ -24,20 +24,23 @@ logger = logging.getLogger(__name__)
 
 MACHINE_SUFFIX: Dict[str, Dict[str, Any]] = {
     "Darwin": {
-        "arm64": "aarch64-apple-darwin-install_only.tar.gz",
-        "x86_64": "x86_64-apple-darwin-install_only.tar.gz",
+        "arm64": ["aarch64-apple-darwin-install_only.tar.gz"],
+        "x86_64": ["x86_64-apple-darwin-install_only.tar.gz"],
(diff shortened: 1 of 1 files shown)


chrysle:
Thanks! Healthy competition stimulates the business; we mustn't leave the field to `uv`. Don't expect us to port _our_ bug fixes though ;-) (I don't know Rust, anyway).

PR #1464, opened 2024-06-28
Fix install all example typo

agrvz:

n/a

## Summary of changes

Tiny typo fix in docs: `pipx instal-all pipx.json` -> `pipx install-all pipx.json`

## Test plan

n/a


chrysle (approved):

chrysle:
Good catch, thanks!

Issue #1460, opened 2024-06-25
"These apps are now globally available" is confusing now that --global is a thing


dechamps:
```
$ pipx --version
1.6.0
$ pipx install cowsay
  installed package cowsay 6.1, installed using Python 3.11.9
  These apps are now globally available
    - cowsay
done! ✨ 🌟 ✨
```

The phrasing "These apps are now globally available" seems to imply that I asked for a system-wide install using `--global`, but I didn't (and it didn't). This is confusing.

Alternatives include:

- Just writing "These apps are now available"
- Writing "These apps are now available user-wide" or "These apps are now available system-wide" depending on whether `--global` was passed

NerusSkyhigh:
Hi everyone,
I saw the tag "good first issue" and tried to help solve it.
I edited the file `commands/common.py` on line 352 with the first alternative suggested but while running the test I encountered some failures:

```
conda create --name pipx-test python pip
conda activate pipx-test
python -m pip install -e .
python -m pip install --user nox
nox -s tests-3.12
```

> ================================ short test summary info ================================
FAILED tests/test_environment.py::test_cli - AssertionError: assert False
FAILED tests/test_install.py::test_install_fetch_missing_python_invalid[3.1] - AssertionError: assert 0
= 2 failed, 344 passed, 8 skipped, 124 deselected, 1 xpassed, 7 warnings in 879.96s (0:14:39) =
nox > Command pytest --cov=pipx --cov-report= tests failed with exit code 1
nox > Session tests-3.12 failed.

As I don't think the line in itself is wrong (I mean, I only changed a string), I may need some help understanding the tests work so  that I can contribute correctly. Any suggestion is welcome.

As we are on the topic, @dechamps, do you think it would be better to also change line 358 from `These manual pages are now globally available` to `These manual pages are now available`?

huxuan:
> I saw the tag "good first issue" and tried to help solve it.

Thanks and welcome, @NerusSkyhigh!

> I may need some help understanding the tests work so that I can contribute correctly. Any suggestion is welcome.

I would suggest to create a pull request in case there is some local environment releate error. If not, it is still more convenient to deal with the problem  directly based on the pull request together.

> As we are on the topic, @dechamps, do you think it would be better to also change line 358 from `These manual pages are now globally available` to `These manual pages are now available`?

IMO, it is fine. The `globally` term is kind of ambiguous. Let us wait for other buddies' response.

NerusSkyhigh:
[...]

PR #1470, opened 2024-07-08
Iussue #1460

NerusSkyhigh:

## Summary of changes
Changed `commands/common.py`line 352 from `output.append("  These apps are now globally available")` to `output.append("  These apps are now available")` as suggested on [Iussue 1460](https://github.com/pypa/pipx/issues/1460).

## Test plan

Tested by running

```
conda create --name pipx-test python pip
conda activate pipx-test
python -m pip install -e .
python -m pip install --user nox
nox -s tests-3.12
```

### DIFF
--- a/changelog.d/1460.docs.md
+++ b/changelog.d/1460.docs.md
@@ -1,0 +1,1 @@
+provide an alternative output after an installation that is less confusing.
--- a/src/pipx/commands/common.py
+++ b/src/pipx/commands/common.py
@@ -349,7 +349,7 @@ def _get_list_output(
     )
 
     if new_install and (exposed_binary_names or unavailable_binary_names):
-        output.append("  These apps are now globally available")
+        output.append("  These apps are now available")
     output.extend(f"    - {name}" for name in exposed_binary_names)
     output.extend(
         f"    - {red(name)} (symlink missing or pointing to unexpected location)" for name in unavailable_binary_names


Gitznik:
Thanks for this contribution.

I personally am in favor or adding the `globally` in the output, if the `--global` flag was passed, and dropping it if it was not.

NerusSkyhigh:
Hi,
I feel like this issue is solved but is not getting accepted or rejected. Is there something I can do to improve it or speed up the reviewing project?
Thanks in advance for any indication.

huxuan:
> Is there something I can do to improve it or speed up the reviewing project?

As @Gitznik mentioned, we should still keep `globally` when `--global` is passed.

NerusSkyhigh:
Hi,
sorry for the late answer but I wasn't aware of the package [argparse](https://docs.python.org/3/library/argparse.html) and it took me a while to understand how arguments are parsed.
Anyway, I don't think there is an easy way to do it. The corresponding function has signature
`def _get_list_output(
    python_version: str,
    python_is_standalone: bool,
    package_version: str,
    package_name: str,
    new_install: bool,
    exposed_binary_names: List[str],
    unavailable_binary_names: List[str],
    exposed_man_pages: List[str],
    unavailable_man_pages: List[str],
    injected_packages: Optional[Dict[str, PackageInfo]] = None,
    suffix: str = "",
)`
and, as far as i know, none of these contains any reference to `argparse` to check wheter the option "--global" was passed.
I think there are two possible paths:
1. Rewrite the signature of the function, with a default value and backtrace its call up to a parent function where `argparse` is available
2. Ignore all possible guidelines and just do something like
```
for arg in sys.argv:
    if "arg" == "--global":
          # print one string
    else:
          # print the other
```
but I'm not a big fan of this solution.

Gitznik:
> I think there are two possible paths:
> 1. Rewrite the signature of the function, with a default value and backtrace its call up to a parent function where `argparse` is available
> 2. Ignore all possible guidelines and just do something like

Definitely option 1. Somewhere up the call stack it must be known whether the install is global, so you should be able to pass that right through. There should be no need for you to interact with argparse directly.

Issue #1443, opened 2024-06-05
HIGH: pipx install --global order broken in 1.6.0


wdoekes:
**Describe the bug**

Doing `pipx install --global` behaves differently from `pipx --global install`. The latter command does not do global but local.

**How to reproduce**

```
root@05515ef1bff7:/# pipx --global install dutree
  installed package dutree 1.8, installed using Python 3.9.2
  These apps are now globally available
    - dutree
⚠️  Note: '/root/.local/bin' is not on your PATH environment variable. These apps will not be globally accessible until your PATH is updated. Run `pipx ensurepath` to automatically
    add it, or manually modify your PATH in your shell's config file (e.g. ~/.bashrc).
done! ✨ 🌟 ✨
root@05515ef1bff7:/# which dutree
root@05515ef1bff7:/# ls /root/.local/bin
dutree
```
^- not global
```
root@05515ef1bff7:/# pipx uninstall dutree
uninstalled dutree! ✨ 🌟 ✨
```
```
root@05515ef1bff7:/# pipx install --global dutree
  installed package dutree 1.8, installed using Python 3.9.2
  These apps are now globally available
    - dutree
done! ✨ 🌟 ✨
root@05515ef1bff7:/# which dutree
/usr/local/bin/dutree
```
^- yes, global

```
root@05515ef1bff7:/# pipx --version
1.6.0
```

**Expected behavior**

Both `install --global` and `--global install` should install in the same place: /usr/local/bin.

**Cause?**

Likely the code from #1397 has something to do with this.

1.5 works fine:
```
root@05515ef1bff7:/# pipx uninstall --global dutree
uninstalled dutree! ✨ 🌟 ✨
```
```
root@05515ef1bff7:/# pip3 install pipx==1.5.0
Collecting pipx==1.5.0
  Downloading pipx-1.5.0-py3-none-any.whl (72 kB)
     |████████████████████████████████| 72 kB 653 kB/s 
...
Successfully installed pipx-1.5.0
```
```
root@05515ef1bff7:/# pipx --global install dutree
  installed package dutree 1.8, installed using Python 3.9.2
  These apps are now globally available
    - dutree
done! ✨ 🌟 ✨
```
```
root@05515ef1bff7:/# which dutree
/usr/local/bin/dutree
```
In 1.5 it complains when doing things "the new way":
```
root@05515ef1bff7:/# pipx install --global dutree
usage: pipx [-h] [--quiet] [--verbose] [--global] [--version]
[...]

PR #1476, opened 2024-07-12
Fix #1475 (`--global` installs to `~/.local`)

### LINKED ISSUE #1475: `pipx install --global` partially installs in `~/.local`
**Describe the bug**

When executing `pipx install --global`:

- It correctly installs the venvs to `/opt/pipx/venvs` and the binaries and manual pages to `/usr/local/bin` and `/usr/local/man`.
- The shared-libs are mistakenly installed to `~/.local/share/pipx`.
- Empty and unused `~/.local/bin` and `~/.local/share/man` directories are created. (This can be a problem if there are no write permissions to `~/.local`, e.g. in a docker container without a proper user home directory.)

**How to reproduce**

```sh
$ docker run -it --rm python:3 bash
$$ pip install git+https://github.com/pypa/pipx.git
$$ pipx install --global cowsay
$$ which cowsay
/usr/local/bin/cowsay   # --> good
$$ ls /opt/pipx/venvs
cowsay                  # --> good
$$ ls /opt/pipx/
venvs                   # --> bad: bin and share should be here
$$ ls ~/.local/
bin  share  state       # --> bad: bin and share should NOT exist here; good: state/pipx/log is ok here
$ ls ~/.local/bin/
                        # --> bad: ~/.local/share/bin is empty
$ ls ~/.local/share/
man  pipx               # --> bad: ~/.local/share/pipx is actually created and used here
```

**Expected behavior**

```sh
$ docker run -it --rm python:3 bash
$$ pip install git+https://github.com/pypa/pipx.git
$$ pipx install --global cowsay
$$ which cowsay
/usr/local/bin/cowsay
$$ ls /opt/pipx/venvs
cowsay
$$ ls /opt/pipx/
bin  share  venvs
$$ ls ~/.local/
state
```

viscruocco:
Fixes #1475 

_Does not fix the other --global related issue https://github.com/pypa/pipx/issues/1443 which is a completely different problem._

## Summary of changes

- Changed setup/initialization order of pipx.paths and pipx.shared_libs to correctly account for the global flag.
- ~~Added `PIPX_GLOBAL_[HOME|BIN_DIR|MAN_DIR]` the output of `pipx environment`.~~ Split to #1493 
- ~~Extended and updated documentation related to `--global`.~~ Split to #1493 

## Test plan

Tested global install by running:
```sh
$ docker run -it --rm python:3 bash
$$ mkdir -p ~/.local
$$ pip install git+https://github.com/viscruocco/pipx.git@fix-global-path-handling
$$ pipx install --global --verbose pycowsay
$$ which pycowsay
/usr/local/bin/pycowsay
$$ ls /opt/pipx/
logs  py  shared  venvs
$$ ls ~/.local
  # --> empty
```

Tested local install by running:
```
$ docker run -it --rm python:3 bash
$$ pip install git+https://github.com/viscruocco/pipx.git@fix-global-path-handling
$$ pipx install --verbose pycowsay
$$ pipx ensurepath --verbose
$$ source ~/.bashrc
$$ which pycowsay
/root/.local/bin/pycowsay
$$ ls ~/.local
bin  share  state
```


src/pipx/main.py
```diff
@@ -1087,6 +1100,9 @@ def setup(args: argparse.Namespace) -> None:
         print_version()
         sys.exit(0)
 
+    if not constants.WINDOWS and args.is_global:
```
uranusjr, line 1103: I don’t remember the context… why is Windows excluded here?

viscruocco: I don't know the background for that decision, but its explicitly mentioned in the docs that the `--global` flag is not available on Windows.

Quoting from https://pipx.pypa.io/stable/how-pipx-works/:
> * adding --global flag to any pipx command will execute the action in global scope which will expose app to all users - [reference](https://pipx.pypa.io/stable/installation/#global-installation). **Note that this is not available on Windows.**

tests/conftest.py
viscruocco, line 1: I'm running out of ideas why most of the `test_install.py` tests are failing. Any hint in the right direction would be much appreciated.

image

The common error looks like this:
image

https://github.com/pypa/pipx/actions/runs/9987614443/job/27602334951?pr=1476

src/pipx/paths.py
```diff
@@ -83,24 +96,40 @@ def home(self) -> Path:
...
         return Path(self._base_shared_libs or self.home / "shared").resolve()
 
     def make_local(self) -> None:
-        self._base_home = get_expanded_environ("PIPX_HOME")
-        self._base_bin = get_expanded_environ("PIPX_BIN_DIR")
-        self._base_man = get_expanded_environ("PIPX_MAN_DIR")
-        self._home_exists = self._base_home is not None or any(fallback.exists() for fallback in self._fallback_homes)
+        self._base_home = OVERRIDE_PIPX_HOME or get_expanded_environ("PIPX_HOME")
+        self._default_home = DEFAULT_PIPX_HOME
+        self._base_bin = OVERRIDE_PIPX_GLOBAL_BIN_DIR or get_expanded_environ("PIPX_BIN_DIR")
```
chrysle, line 109: This might cause the failure? It's hard to spot..

```suggestion
        self._base_bin = OVERRIDE_PIPX_BIN_DIR or get_expanded_environ("PIPX_BIN_DIR")
```

viscruocco: Fixed in [11dd6a6](https://github.com/pypa/pipx/pull/1476/commits/11dd6a6395a550d8703516a51b96e7060addd613)

changelog.d/1475.feature.md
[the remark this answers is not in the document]
viscruocco: Split to #1493

src/pipx/paths.py
```diff
@@ -83,24 +96,40 @@ def home(self) -> Path:
         elif self._fallback_home:
             home = self._fallback_home
         else:
-            home = Path(DEFAULT_PIPX_HOME)
+            home = Path(self._default_home)
```
chrysle, line 99: This should already be a `Path` object, IIUC.

```suggestion
            home = self._default_home
```

src/pipx/paths.py
```diff
@@ -54,27 +67,27 @@ def venvs(self) -> Path:
...
 
     @property
     def bin_dir(self) -> Path:
-        return Path(self._base_bin or DEFAULT_PIPX_BIN_DIR).resolve()
+        return Path(self._base_bin or self._default_bin).resolve()
 
     @property
     def man_dir(self) -> Path:
-        return Path(self._base_man or DEFAULT_PIPX_MAN_DIR).resolve()
+        return Path(self._base_man or self._default_man).resolve()
```
chrysle, line 90:
```suggestion
        return (self._base_man or self._default_man).resolve()
```

src/pipx/paths.py
```diff
@@ -54,27 +67,27 @@ def venvs(self) -> Path:
...
     def venv_cache(self) -> Path:
         if self._home_exists or not LINUX:
             return self.home / ".cache"
-        return user_cache_path("pipx")
+        return self._default_cache
 
     @property
     def bin_dir(self) -> Path:
-        return Path(self._base_bin or DEFAULT_PIPX_BIN_DIR).resolve()
+        return Path(self._base_bin or self._default_bin).resolve()
```
chrysle, line 86:
```suggestion
        return (self._base_bin or self._default_bin).resolve()
```

src/pipx/shared_libs.py
```diff
@@ -23,21 +23,39 @@
 
 class _SharedLibs:
     def __init__(self) -> None:
-        self.root = paths.ctx.shared_libs
```
chrysle, line 26: Why did you make this a property?

tests/conftest.py
```diff
@@ -201,6 +202,7 @@ def pipx_temp_env(tmp_path, monkeypatch, pipx_session_shared_dir, request, utils
     seamless.
     """
     pipx_temp_env_helper(pipx_session_shared_dir, tmp_path, monkeypatch, request, utils_temp_dir, pipx_local_pypiserver)
+    paths.ctx.make_local()
```
chrysle, line 205: Why did you add this?

src/pipx/paths.py
```diff
@@ -83,24 +96,40 @@ def home(self) -> Path:
         elif self._fallback_home:
             home = self._fallback_home
         else:
-            home = Path(DEFAULT_PIPX_HOME)
+            home = Path(self._default_home)
         return home.resolve()
 
     @property
     def shared_libs(self) -> Path:
         return Path(self._base_shared_libs or self.home / "shared").resolve()
```
viscruocco, line 104:
```suggestion
        return (self._base_shared_libs or self.home / "shared").resolve()
```

src/pipx/shared_libs.py
viscruocco: This class is only instantiated exactly once: during loading of this module. (See https://github.com/pypa/pipx/blob/62aa16301edc70fb503adf5e54eaf48f955441d9/src/pipx/shared_libs.py#L151)

This is before any command line arguments are parsed and thus before e.g. the `--global` argument can have any effect. (Also before e.g. the monkeypatching in the test is happening.) This problem has previously already been (insufficiently) addressed [here](https://github.com/pypa/pipx/commit/863a3ec600da6b8dd5155d8b3a92a991f0244e50#diff-78cbc9ee1ca185b4d739dc2e7c8b263156522c3fe8f73ae4511b73755ed73139L49).

By converting all attributes to properties this problem is circumvented completely, because the paths are always recomputed based on the current `paths.ctx.shared_libs`. The necessary computations do not involve any file system operations and should therefore hopefully not be too expensive to repeat on every attribute-read.

PS: I do not particularly like that the `paths` and `shared_libs` use "statically" initialized global objects for exactly this reason. IMHO passing around immutable `paths` and `shared`_libs` objects as function parameters would simplify the whole thing and improve testability a lot; but it would pose a much bigger change than the present bugfix.

tests/conftest.py
viscruocco: The `OVERRIDE_PIPX_*` variables are only read and used during `make_local()` and `make_global()` calls. Simply monkeypatching these variables above does in itself not change anything on the `paths.ctx` object. Instead these changes are only "committed" during this `make_local()` call here.

Similarly to the [comment on `shared_libs`](https://github.com/pypa/pipx/pull/1476/files#r1694114472) this in mainly caused by the fact that `paths` uses a mutable global object that is initialized during module load (before any testing code can interfere) and then modified afterwards through `make_local()` and `make_global()` calls.

src/pipx/shared_libs.py
chrysle: Thanks for clarifying!

> PS: I do not particularly like that the `paths` and `shared_libs` use "statically" initialized global objects for exactly this reason. IMHO passing around immutable `paths` and `shared_libs` objects as function parameters would simplify the whole thing and improve testability a lot; but it would pose a much bigger change than the present bugfix.

Makes sense, let's put that on the todo list.

tests/conftest.py
chrysle: That's true, and regarding my above comment, this invocation should even be _inside_ `pipx_temp_env_helper`.

tests/conftest.py
viscruocco: Amazing! Thank you for spotting this! 🥳

tests/conftest.py
viscruocco: Oh yeah! Fixed in 4609f25.

src/pipx/main.py
chrysle: It's because the submitter did not have access to a Windows system and there was no consensus (or even a discussion) about file system locations on Windows that `pipx` acting globally should install to. Once someone needing this feature steps up, we'll be happy to take it further.

chrysle (approved):

PR #1487, opened 2024-07-22
Support different backends

### LINKED ISSUE #1392: Support different backends for pipx?

There are several open issues about supporting different "backends", such as #1370 for `uv` and #853 for `virtualenv`. Moreover, there is already a project [`pipxu`](https://github.com/bulletmark/pipxu) that claims to have reimplemented pipx while using uv.

1. Abstract all the operations related to managed virtual environments.
2. Implement venv, virtualenv and uv for these abstracted operations.
3. Add virtualenv and uv as optional dependencies and use them with higher priority than venv when pipx is initialized.

There might be two other approaches to implement the same feature:
1. Create a backend plugin system and implement the backend in a separate repo.
2. Create a standalone and universal backend library that can be used directly in pipx.

dukecat0:

## Summary of changes
Closes #1392
This allows users to choose from different backends and also speeds up the installation process.

Screenshot 2024-07-21 at 11 11 55 AM

## Test plan

Tested by running

```
pipx install --backend uv --install uv black
pipx run --backend uv --installer uv black
```


src/pipx/venv.py
```diff
@@ -244,26 +276,40 @@ def install_package(
         # check syntax and clean up spec and pip_args
         (package_or_url, pip_args) = parse_specifier_for_install(package_or_url, pip_args)
 
-        logger.info("Installing %s", package_descr := full_package_description(package_name, package_or_url))
+        logger.info(
+            "Installing %s",
+            package_descr := full_package_description(package_name, package_or_url),
+        )
```
chrysle, line 282: I'd keep these style changes out.

src/pipx/venv.py
```diff
@@ -439,10 +504,16 @@ def has_package(self, package_name: str) -> bool:
...
+            "Upgrading %s",
+            package_descr := full_package_description(package_name, package_name),
+        )
         with animate(f"upgrading {package_descr}", self.do_animation):
-            pip_process = self._run_pip(["--no-input", "install"] + pip_args + ["--upgrade", package_name])
-        subprocess_post_check(pip_process)
+            if self.installer != "uv":
+                upgrade_process = self._run_pip(["--no-input", "install"] + pip_args + ["--upgrade", package_name])
+            else:
+                upgrade_process = self._run_uv(["install", "--upgrade", package_name])
```
chrysle, line 515: Creating a generic function `Venv._run_installer()` selecting the right command according to the installer value might simplify this and other logic?

chrysle (commented):
Thank you very much for working on this feature! Some initial thoughts.

pyproject.toml
```diff
@@ -43,6 +43,12 @@ dependencies = [
   "tomli; python_version<'3.11'",
   "userpath!=1.9,>=1.6",
 ]
+optional-dependencies.uv = [
+  "uv>=0.2.27",
+]
+optional-dependencies.virtualenv = [
+  "virtualenv>=20.26.3",
+]
```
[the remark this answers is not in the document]
dukecat0: Looks like it's auto-reformatted by `pyproject-fmt`.

src/pipx/commands/upgrade.py
```diff
@@ -137,6 +137,9 @@ def _upgrade_venv(
                 include_dependencies=False,
                 preinstall_packages=None,
                 python_flag_passed=python_flag_passed,
+                # What to deal with this? Should we allow cli option?
+                backend=DEFAULT_BACKEND,
+                installer=DEFAULT_INSTALLER,
```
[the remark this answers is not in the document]
dukecat0: Since there is an option `--install` in upgrade command, perhaps we should allow them to pass `--backend` and `--installer` as well? But this would be quite confusing when people are upgrading packages only instead of installing. How about we add a conditional check to see whether these arguments are used together with `--install` and fall back with a warning if the package exists?

src/pipx/pipx_metadata_file.py
```diff
@@ -100,6 +105,9 @@ def to_dict(self) -> Dict[str, Any]:
     def _convert_legacy_metadata(self, metadata_dict: Dict[str, Any]) -> Dict[str, Any]:
         if metadata_dict["pipx_metadata_version"] in (self.__METADATA_VERSION__):
             pass
+        elif metadata_dict["pipx_metadata_version"] == "0.5":
+            metadata_dict["backend"] = "venv"
+            metadata_dict["installer"] = "pip"
```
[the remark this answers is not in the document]
dukecat0: Since we only supported venv and pip before this feature, so I think it's reasonable to set them to these values?

src/pipx/commands/upgrade.py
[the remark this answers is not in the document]
chrysle: Sounds good.

pyproject.toml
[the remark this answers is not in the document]
uranusjr: Makes me wonder if we should have `optional-dependencies.pip` too (just installs pip, which is no-op most of the time, but explictness is not a bad thing here.)

src/pipx/venv.py
```diff
@@ -469,12 +492,38 @@ def upgrade_package(
...
+        log_stderr: bool = True,
+    ) -> "CompletedProcess[str]":
+        # do not use -q with `pip install` so subprocess_post_check_pip_errors
+        #   has more information to analyze in case of failure.
+        if installer != "uv":
+            return self._run_pip(
+                ["--no-input", "install"] + (["--no-dependencies"] if no_deps else []) + cmd, quiet=quiet
+            )
+        else:
+            return self._run_uv(["pip", "install"] + (["--no-deps"] if no_deps else []) + cmd, quiet=quiet)
```
chrysle, line 511: Would thid work? I believe `uv pip` supports the `--no-dependencies` option?

```suggestion
        install_cmd = ["install"] + (["--no-dependencies"] if no_deps else []) + cmd
        if installer == "pip":
            return self._run_pip(["--no-input"] + install_cmd, quiet=quiet)
        else:
            return self._run_uv(["pip"]  + install_cmd, quiet=quiet)
```

src/pipx/backend.py
```diff
@@ -0,0 +1,21 @@
...
+import shutil
+
+SUPPORTED_VENV_BACKENDS = ("uv", "venv", "virtualenv")
+SUPPORTED_INSTALLERS = ("uv", "pip")
+
+logger = logging.getLogger(__name__)
+
+
+def path_to_exec(executable: str, installer: bool = False) -> str:
+    if executable == "venv" or "pip":
```
chrysle, line 11:
```suggestion
    if executable in ("venv", "pip"):
```

src/pipx/backend.py
```diff
@@ -0,0 +1,21 @@
...
+        return executable
+    path = shutil.which(executable)
+    if path:
+        return path
+    elif installer:
+        logger.warning(f"{executable} not found on PATH. Falling back to pip.")
+        return ""
+    else:
+        logger.warning(f"{executable} not found on PATH. Falling back to venv.")
+        return ""
```
chrysle, line 21: We might as well return the program names here, to cut down on conditions below?

```suggestion
    elif installer:
        logger.warning(f"'{executable}' not found on PATH. Falling back to 'pip'.")
        return "pip"
    else:
        logger.warning(f"'{executable}' not found on PATH. Falling back to 'venv'.")
        return "venv"
```

src/pipx/venv.py
dukecat0: Not really, this option exists as `--no-deps` in uv.
```
error: unexpected argument '--no-dependencies' found

  tip: a similar argument exists: '--no-deps'
```

src/pipx/backend.py
dukecat0: The empty string is used to determine the backend and installer to use here:
```py
        if not path_to_exec(self.backend):
            self.backend = "venv"
        if not path_to_exec(self.installer, installer=True):
            self.installer = "pip"
```

chrysle: I know, I thought we could assign directly:

```python
self.backend = path_to_exec(self.backend)
self.installer = path_to_exec(self.installer, installer=True)
```

But I guess that's not doable, as in case of `uv` or `virtualenv`  a path would be returned.

dukecat0: > But I guess that's not doable, as in case of uv or virtualenv a path would be returned.

Yes exactly.

src/pipx/venv.py
chrysle: Ah, but `pip` itself [supports `--no-deps`](https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-no-deps). Edited the suggestion.

dukecat0: Ah I still haven't pushed the changes.

src/pipx/venv.py
```diff
@@ -469,12 +492,36 @@ def upgrade_package(
...
+        # do not use -q with `pip install` so subprocess_post_check_pip_errors
+        #   has more information to analyze in case of failure.
+        install_cmd = ["install"] + (["--no-deps"] if no_deps else []) + cmd
+        if self.installer == "pip":
+            return self._run_pip(["--no-input"] + install_cmd, quiet=quiet)
+        else:
+            return self._run_uv(["pip"] + install_cmd, quiet=quiet)
+
+    def _run_uv(self, cmd: List[str], quiet: bool = True) -> "CompletedProcess[str]":
+        cmd = [path_to_exec("uv", installer=True)] + cmd + ["--python", str(self.python_path)]
```
chrysle, line 512: I think we should always differentiate between `uv` and `pip`.  That way, we also don't need to pass in `pip` as a `uv` command when calling the function.

```suggestion
        if not uv_path := path_to_exec("uv", installer=True):
            self._run_pip(cmd=cmd, quiet=quiet)
        else:
            cmd = [uv_path] + cmd + ["--python", str(self.python_path)]
```

src/pipx/venv.py
```diff
@@ -212,14 +248,19 @@ def upgrade_packaging_libraries(self, pip_args: List[str]) -> None:
...
+            if self.backend != "uv":
+                self.upgrade_package_no_metadata("pip", pip_args)
 
     def uninstall_package(self, package: str, was_injected: bool = False):
         try:
             logger.info("Uninstalling %s", package)
             with animate(f"uninstalling {package}", self.do_animation):
-                cmd = ["uninstall", "-y"] + [package]
-                self._run_pip(cmd)
+                if self.installer != "uv":
```
chrysle, line 258: I'd say, explicit is better than implicit.

```suggestion
                if self.installer == "pip":
```

chrysle, line 258: I'd say, explicit is better than implicit.

```suggestion
                if self.installer == "pip":
```

chrysle, line 258: I'd say, explicit is better than implicit.

```suggestion
                if self.installer == "pip":
```

chrysle, line 258: I'd say, explicit is better than implicit.

```suggestion
                if self.installer == "pip":
```

src/pipx/venv.py
```diff
@@ -157,20 +180,31 @@ def main_package_name(self) -> str:
...
         override_shared -- Override installing shared libraries to the pipx shared directory (default False)
         """
         logger.info("Creating virtual environment")
-        with animate("creating virtual environment", self.do_animation):
-            cmd = [self.python, "-m", "venv"]
-            if not override_shared:
-                cmd.append("--without-pip")
+        with animate(f"creating virtual environment using {self.backend}", self.do_animation):
+            if self.backend == "venv":
+                cmd = self.default_create_venv_cmd(override_shared=override_shared)
```
chrysle, line 196: If this doesn't look too bulky, we can rationalise the function away; it's used only once.

```suggestion
                cmd = [self.python, "-m", "venv"]
                if not override_shared:
                    cmd.append("--without-pip")
```

src/pipx/venv.py
dukecat0: Why is this logic needed here? This function is `_run_uv()` and we already have the check before calling this function.

chrysle: Oh, sorry, I forgot already the `Venv` class's `__init__` method contains a condition, please disregard this.

pyproject.toml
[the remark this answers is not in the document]
dukecat0: > Makes me wonder if we should have `optional-dependencies.pip` too (just installs pip, which is no-op most of the time, but explictness is not a bad thing here.)

I am not sure whether there will be downsides for it, since it can be quite confusing to users?

src/pipx/backend.py
```diff
@@ -0,0 +1,24 @@
...
+        return executable
+    path = shutil.which(executable)
+    if path:
+        return path
+    elif installer:
+        logger.warning(f"'{executable}' not found on PATH. Falling back to 'pip'.")
+        return ""
+    else:
+        logger.warning(f"'{executable}' not found on PATH. Falling back to 'venv'.")
+        return ""
```
Gitznik, line 24: I find the `installer` argument confusing here. Everywhere else, `installer` is a `str` of the installer, not a boolean. IMO having an `ExecType` enum here with `INSTALLER` and `BACKEND` would make this function much clearer.

Gitznik (changes requested):

chrysle:
Per https://github.com/pypa/pipx/issues/1539#issuecomment-2335187541, we should consider adding the [`--no-seed` option](https://virtualenv.pypa.io/en/latest/cli_interface.html#no-seed) to the `virtualenv` (not sure concerning the `uv` mechanism) backend invocation.

src/pipx/backend.py
dukecat0: Or maybe we can simply rename `installer` argument to `is_installer`?

dukecat0:
It seems like uv doesn't have this issue since it doesn't carry extra seed packages by default:
```
activate      activate.fish  activate_this.py  python
activate.bat  activate.nu    deactivate.bat    python3
activate.csh  activate.ps1   pydoc.bat	       python3.9
```

src/pipx/venv.py
```diff
@@ -162,15 +185,22 @@ def create_venv(self, venv_args: List[str], pip_args: List[str], override_shared
...
+                    cmd.append("--no-pip")
             venv_process = run_subprocess(cmd + venv_args + [str(self.root)], run_dir=str(self.root))
         subprocess_post_check(venv_process)
-
-        shared_libs.create(verbose=self.verbose, pip_args=pip_args)
-        if not override_shared:
+        if self.backend != "uv" or self.installer != "uv":
+            shared_libs.create(verbose=self.verbose, pip_args=pip_args)
+        if not override_shared and (self.backend != "uv" or self.installer != "uv"):
             pipx_pth = get_site_packages(self.python_path) / PIPX_SHARED_PTH
```
mayeut, line 204: Should the `shared` environment still be injected/exist after this or should it only be used (and thus maybe renamed) as the environment holding `pip` and allowing `"{shared_libs.python_path}" -m pip --python "{self.python_path}"` being used as an installer (rather than `"{self.python_path}" -m pip` in `_run_pip`)

pyproject.toml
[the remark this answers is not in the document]
mayeut: > > Makes me wonder if we should have optional-dependencies.pip too (just installs pip, which is no-op most of the time, but explictness is not a bad thing here.)

> I am not sure whether there will be downsides for it, since it can be quite confusing to users?

Quite confusing indeed IMHO. I wouldn't add a dependency that never gets used. I would expect only `pip` installed via the specified optional-dependency would be used as the installer. Instead, pip in sharedlibs is the one used.

That being said, the same is true for `uv` & `virtualenv` optional dependencies. In the current state of the PR, the first version present on `PATH` (c.f. `path_to_exec`) gets used. So even if pipx gets installed with those optional dependencies, pipx might fail to use those versions or just fail to find them.
Some precedence should probably be implemented in `path_to_exec` e.g. prefer a version found in the same environment as pipx then find on `PATH`.

If the added complexity in `path_to_exec` is not wanted, then I'd suggest to drop the optional dependencies and to require the user to provide them on `PATH` (which could be documented as `pipx install virtualenv`/`pipx install uv` to bootstrap the tool installation rather than asking for optional dependencies)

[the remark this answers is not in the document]
dukecat0: > If the added complexity in path_to_exec is not wanted, then I'd suggest to drop the optional dependencies and to require the user to provide them on PATH (which could be documented as `pipx install virtualenv`/`pipx install uv` to bootstrap the tool installation rather than asking for optional dependencies)

+1 on this, especially we encourage users to install pipx through their system package managers instead of `pip`, so extra dependencies won't work in this case as well.

src/pipx/venv.py
dukecat0: Using `--python` option sounds reasonable, but would there be any scenarios where users would prefer to keep the `.pth` file?

mayeut: The only thing that comes to mind would be if they were directly running python in an app venv to run pip instead of using `pipx runpip`.

gaborbernat (commented):
Needs rebase and fixing the CI.

PR #1497, opened 2024-08-04
Assorted repo-review suggestions

DimitriPapadopoulos:

## Summary of changes

Changes suggested by repo-review:
https://learn.scientific-python.org/development/guides/repo-review/?repo=pypa%2Fpipx&branch=main

## Test plan

Pass CI tests.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,7 @@ repos:
   rev: v0.5.5
   hooks:
   - id: ruff
-    args: [ "--fix", "--unsafe-fixes", "--exit-non-zero-on-fix"]
+    args: [ "--fix", "--unsafe-fixes", "--show-fixes", "--exit-non-zero-on-fix"]
   - id: ruff-format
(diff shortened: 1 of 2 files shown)


src/pipx/standalone_python.py
```diff
@@ -119,7 +119,7 @@ def _unpack(full_version, download_link, archive: Path, download_dir: Path):
             )
 
         with tarfile.open(archive, mode="r:gz") as tar:
-            tar.extractall(download_dir)
+            tar.extractall(download_dir, filter="data")
```
uranusjr, line 122: I think `filter` is only available in 3.12+? We need a compatibility layer here.

DimitriPapadopoulos: Indeed, but that doesn't seem to be the cause of the CI errors.

DimitriPapadopoulos:
I've left out this `pytest` option for now:
```
xfail_strict = true
```
I'll try to add it in a different PR.

uranusjr (approved):

PR #1510, opened 2024-08-17
Fixes #1509 short-circuit of installing packages when already installed

### LINKED ISSUE #1509: `pipx install pkg1 pkg2` doesn't install `pkg2` if `pkg1` already installed
**Describe the bug**
If I have already installed `black` by running `pipx install black` and I attempt to run `pipx install black isort`, `black` is detected as having been already installed, and `isort` is not installed.

**How to reproduce**
`pipx install black`
`pipx install black isort`
\# _isort not installed!_

**Expected behavior**

`black` should be detected as already installed, and `isort` should be installed by pipx

```[tasklist]
### Tasks
- [ ] https://github.com/pypa/pipx/pull/1510
```

ayemiller:

## Summary of changes

Updates logic of `install.py` to continue installing packages if one is already installed.

## Test plan

Added tests to exercise the bug and demonstrate fix:
`test_install_three_packages_with_first_already_installed`
`test_install_two_packages_with_first_two_already_installed`

Tested by running

```
nox -s tests-3.12 -- -k test_install
```


chrysle:
Thanks for catching! Probably, we don't need two tests, though. Besides, the test function names don't really correspond to the actual test procedure..

ayemiller:
> Thanks for catching! Probably, we don't need two tests, though. Besides, the test function names don't really correspond to the actual test procedure..

Whoops! Updated :)

src/pipx/commands/install.py
```diff
@@ -84,7 +83,9 @@ def install(
                         """
                     )
                 )
-                return EXIT_CODE_INSTALL_VENV_EXISTS
```
chrysle, line 87: Shouldn't this still be returned if only one package was given as argument?

chrysle (approved):

changelog.d/1509.bugfix.md
```diff
@@ -0,0 +1 @@
+Installs multiple packages regardless if one or more already installed.
```
chrysle, line 1: This should be more descriptive.

```suggestion
Fix installation abortion on multiple packages when one or more are already installed.
```

chrysle (approved):

Gitznik (approved):

PR #1519, opened 2024-08-24
Apply repo-review suggestions

DimitriPapadopoulos:

## Test plan

Tested by running

```
```


src/pipx/commands/environment.py
```diff
@@ -38,7 +38,7 @@ def environment(value: str) -> ExitCode:
         "PIPX_HOME_ALLOW_SPACE": str(paths.ctx.allow_spaces_in_home_path).lower(),
     }
     if value is None:
-        print("Environment variables (set by user):")
+        print("Environment variables (set by user):")  # type: ignore[unreachable]
```
DimitriPapadopoulos, line 41: `value` is not `Optional`, so in theory it cannot be `None`.

Do we still want to defend against an actual `None`?

src/pipx/commands/inject.py
```diff
@@ -32,7 +32,7 @@ def inject_dep(
 ) -> bool:
     logger.debug("Injecting package %s", package_spec)
 
-    if not venv_dir.exists() or not next(venv_dir.iterdir()):
+    if not venv_dir.exists() or next(venv_dir.iterdir(), None) is None:
```
DimitriPapadopoulos, line 35: `venv_dir.iterdir()` yields `Path` which does not implement `__bool__` or `__len__` so it will always be true in boolean context. We might want to defend against `StopIteration` exceptions raised by `next()`.

src/pipx/commands/uninject.py
```diff
@@ -125,7 +125,7 @@ def uninject(
 ) -> ExitCode:
     """Returns pipx exit code"""
 
-    if not venv_dir.exists() or not next(venv_dir.iterdir()):
+    if not venv_dir.exists() or next(venv_dir.iterdir(), None) is None:
```
DimitriPapadopoulos, line 128: `venv_dir.iterdir()` yields `Path` which does not implement `__bool__` or `__len__` so it will always be true in boolean context. We might want to defend against `StopIteration` exceptions raised by `next()`.

src/pipx/commands/run.py
```diff
@@ -198,11 +198,11 @@ def run(
         # we can't parse this as a package
         package_name = app
 
-    content = None if spec is not None else maybe_script_content(app, is_path)
+    content = None if spec is not None else maybe_script_content(app, is_path)  # type: ignore[redundant-expr]
```
DimitriPapadopoulos, line 201: `spec` is not `Optional`, so in theory it cannot be `None`.

Do we still want to defend against an actual `None`?

src/pipx/commands/run.py
```diff
@@ -198,11 +198,11 @@ def run(
...
     if content is not None:
         run_script(content, app_args, python, pip_args, venv_args, verbose, use_cache)
     else:
-        package_or_url = spec if spec is not None else app
+        package_or_url = spec if spec is not None else app  # type: ignore[redundant-expr]
```
DimitriPapadopoulos, line 205: Same.

src/pipx/main.py
```diff
@@ -409,7 +409,7 @@ def run_pipx_command(args: argparse.Namespace, subparsers: Dict[str, argparse.Ar
             python_flag_passed=python_flag_passed,
         )
     elif args.command == "runpip":
-        if not venv_dir:
+        if not venv_dir:  # type: ignore[truthy-bool]
```
DimitriPapadopoulos, line 412: In theory, `venv_dir` cannot be `None`.

Do we still want to defend against an actual `None`?

src/pipx/shared_libs.py
```diff
@@ -114,7 +114,7 @@ def upgrade(self, *, pip_args: List[str], verbose: bool = False, raises: bool =
             return
 
         if pip_args is None:
-            pip_args = []
+            pip_args = []  # type: ignore[unreachable]
```
DimitriPapadopoulos, line 117: `pip_args` is not `Optional`, so in theory it cannot be `None`.

Do we still want to defend against an actual `None`?

PR #1556, opened 2024-10-07
ci: test on 3.13

henryiii:
Adding testing for 3.13 in CI.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -23,7 +23,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: ["3.12", "3.11", "3.10", "3.9"]
+        python-version: ["3.13", "3.12", "3.11", "3.10", "3.9"]
         include:
           - os: windows-latest
             python-version: "3.12"
@@ -37,6 +37,7 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
           cache: "pip"
+          allow-prereleases: true
       - name: Persistent .pipx_tests/package_cache
         uses: actions/cache@v4
         with:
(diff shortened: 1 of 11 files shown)


henryiii:
Stuck on `cloudtoken`, which is a Poetry project and puts tight upper caps on everything, so `Collecting lxml==4.9.2 (from cloudtoken==2.1.18)` breaks trying to compile old lxml.

Gitznik:
`cloudtoken` has released `2.1.19`, which removes the hard cap for the `lxml` version. Maybe that fixes your issue? I couldn't replicate the behavior of the pipeline locally for some reason, so I can't verify the fix.

huxuan:
Maybe we also need to delete the github action cache: https://github.com/pypa/pipx/actions/caches

henryiii:
I've cleared the cache and retrying.

henryiii:
Ahh, forgot to bump. :facepalm:

henryiii:
Don't see it? https://pypi.org/project/cloudtoken/

huxuan:
Maybe we should skip the tests related to cloudtoken temporarily? Any other suggestions?

dukecat0:
+1 on this. We can add it back later if a new version is released on PyPI.

uranusjr:
The entire cloudtoken repo is now gone. Maybe we should switch to another package instead.

henryiii:
Sounds good to me. :)

henryiii:
I don't think modern cloud token actually has the issue that 0.1.84 seems to have been added for from the changelog (infinite recursion) https://inspector.pypi.io/project/cloudtoken/0.1.84/packages/1c/62/ae6d69f1f2bd624e459d115883d0f5f882a716a876b89bbfb6eb190fc641/cloudtoken-0.1.84-py3-none-any.whl/cloudtoken-0.1.84.dist-info/METADATA. I'm trying that version, as it was a lot simpler, might pass on newer Pythons.

dukecat0:
Closing in favour of #1647

PR #1561, opened 2024-10-11
Amend global command example for installs using `pip`

### LINKED ISSUE #1560: `sudo pipx` does not work when `pipx` is installed with `pip --user`
```bash
$ sudo pipx
sudo: pipx: command not found
```

Lordfirespeed:
Closes #1560


Lordfirespeed:
Hmm, no. This doesn't work either because (obviously) the user doesn't have root privileges.

PR #1626, opened 2025-05-04
Upgrade ruff and rename ruleset: `TCH` → `TC`

DimitriPapadopoulos:

## Summary of changes

Ruff 0.8.0 changes the error codes for the rules in the [`flake8-type-checking`](https://docs.astral.sh/ruff/rules/#flake8-type-checking-tc) category from `TCH` to `TC`. This matches changes that were made in the original `flake8-type-checking` plugin from which these rules were originally adapted.

https://astral.sh/blog/ruff-v0.8.0#new-error-codes-for-flake8-type-checking-rules

Not sure whether to apply or ignore rule [TC006](https://docs.astral.sh/ruff/rules/runtime-cast-value/). See discussions here:
* https://github.com/astral-sh/ruff/issues/14676
* https://github.com/astral-sh/ruff/discussions/17127

## Test plan

Just CI.

The issues with Python 3.8 seem unrelated.

### DIFF
--- a/scripts/generate_man.py
+++ b/scripts/generate_man.py
@@ -13,7 +13,7 @@ from pipx.main import get_command_parser
 def main():
     sys.argv[0] = "pipx"
     parser, _ = get_command_parser()
-    parser.man_short_description = cast(str, parser.description).splitlines()[1]  # type: ignore[attr-defined]
+    parser.man_short_description = cast("str", parser.description).splitlines()[1]  # type: ignore[attr-defined]
 
     manpage = Manpage(parser)
     body = str(manpage)
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,7 +11,7 @@ repos:
   hooks:
     - id: pyproject-fmt
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.6.9
+  rev: v0.11.8
   hooks:
   - id: ruff
     args: [ "--fix", "--unsafe-fixes", "--show-fixes", "--exit-non-zero-on-fix"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,7 +79,7 @@ lint.extend-select = [
   "RSE",
   "RUF012",
   "RUF100",
-  "TCH",
+  "TC",
   "W",
 ]
 lint.ignore = [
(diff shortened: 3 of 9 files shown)


scripts/generate_man.py
```diff
@@ -13,7 +13,7 @@
 def main():
     sys.argv[0] = "pipx"
     parser, _ = get_command_parser()
-    parser.man_short_description = cast(str, parser.description).splitlines()[1]  # type: ignore[attr-defined]
+    parser.man_short_description = cast("str", parser.description).splitlines()[1]  # type: ignore[attr-defined]
```
uranusjr, line 16: Why is this needed?

uranusjr (approved):

scripts/generate_man.py
DimitriPapadopoulos: That's the result of [TC006](https://docs.astral.sh/ruff/rules/runtime-cast-value/) kicking in, probably after upgrading _ruff_.

The purpose of [TC006](https://docs.astral.sh/ruff/rules/runtime-cast-value/) is primarily to enforce a **consistent style**:
> This rule helps enforce a consistent style across your codebase.
> 
> It's often necessary to quote the first argument passed to `cast()`, as type expressions can involve forward references, or references to symbols which are only imported in `typing.TYPE_CHECKING` blocks. This can lead to a visual inconsistency across different `cast()` calls, where some type expressions are quoted but others are not. By enabling this rule, you ensure that all type expressions passed to `cast()` are quoted, enforcing stylistic consistency across all of your `cast()` calls.

So while applying the rule to `str` might not be a clear improvement, it's part of a plan to enforce consistency. This is an opinionated rule, which is why I asked whether you'd rather enforce or ignore it. From https://github.com/astral-sh/ruff/issues/14676#issuecomment-2507873721:
> Everyone will have to make their own trade-offs based on their use-cases, `flake8-type-checking` is definitely one of the more opinionated plugins, so it doesn't come as a surprise to me, that there are some rules or decisions a significant number of people disagree with, but it's also worth recognizing, that you may also find a significant number of people in the opposite camp.
> 
> But I can at least tell you that while, when this rule was first introduced in `flake8-type-checking`, there was one issue opened, that complained about it (mostly due to the churn it caused, since there are no automatic fixes in `flake8`), there has been no mention of it in the two years since then. So it appears that people are generally happy to either live with the rule in its current state or disable it, we haven't had any requests to change it, or make it more configurable.

PR #1631, opened 2025-05-07
Update the logic of finding python interpreter

### LINKED ISSUE #1521: `--fetch-missing-python` not working
The `--fetch-missing-python` option appears to have no effect.

Following the [example](https://pipx.pypa.io/stable/examples/):

`pipx install --fetch-missing-python --python 3.12 pycowsay`

I see

> ⚠️ No executable for the provided Python version '3.12' found in py launcher. Please make sure the provided version is listed when running `py --list`.

It is true that I do not have Python 3.12 installed. The suggested `py --list`  gives

```
py --list
 -V:3.11 *        Python 3.11 (64-bit)
 -V:3.10          Python 3.10 (64-bit)
```

But the [documentation](https://pipx.pypa.io/stable/docs/) says that `--fetch-missing-python` instructs Pipx "to fetch a standalone python build from GitHub if the specified python version is not found locally on the system."

I expected Pipx to automatically download Python 3.12 and create the virtual environment with it.

Verbose output of `pipx install --fetch-missing-python --python 3.12 pycowsay --verbose`:

```
pipx >(setup:1110): pipx version is 1.7.1
pipx >(setup:1111): Default python interpreter is 'C:\Program Files\Python310\python.exe'
⚠️ No executable for the provided Python version '3.12' found in py launcher. Please make sure the provided version is listed when running `py --list`.
```

My system is Windows 11 64-bit.

dukecat0:

## Summary of changes
Update the logic of finding python interpreter
Closes #1521 

## Test plan

Tested by running

```
pipx install black --python 3.11 --fetch-missing-python
```


uranusjr:
I think this would cause versions to be incorrectly downloaded when an installed version is available, but only visible via registry (and `py.exe`)?

dukecat0:
Nice catch! I will update the logic later.

uranusjr (approved):

src/pipx/interpreter.py
```diff
@@ -97,7 +97,8 @@ def find_python_interpreter(python_version: str, fetch_missing_python: bool = Fa
         if py_executable:
             return py_executable
     except (subprocess.CalledProcessError, FileNotFoundError) as e:
-        raise InterpreterResolutionError(source="py launcher", version=python_version) from e
+        if not fetch_missing_python and not FETCH_MISSING_PYTHON:
+            raise InterpreterResolutionError(source="py launcher", version=python_version) from e
```
uranusjr, line 101:
```suggestion
            raise
```

This should be enough?

dukecat0: This will result in a very detailed traceback, which is not user friendly here.

dukecat0: Both of them will provide the same output like this:
> $ pipx install black --python 3.7
⚠️ No executable for the provided Python version '3.7' found in py launcher. Please make sure the provided version is
    listed when running `py --list`.

PR #1632, opened 2025-05-07
Update the archive name of build of Python for Windows

### LINKED ISSUE #1630: URLs used by `--fetch-missing-python` to find Windows Python builds is incorrect.
**Describe the bug**

The URL suffix defined here for Windows is incorrect:

https://github.com/pypa/pipx/blob/2146ab6d3c4fb2fe21e30f9df0b8343539b0b98c/src/pipx/standalone_python.py#L43

According to the [documentation for python-build-standalone](https://gregoryszorc.com/docs/python-build-standalone/main/running.html) the word "`shared`" has been dropped from the suffix:

> *-windows-msvc-shared
This is a standard build of Python for Windows. There are DLLs for Python and extensions. These builds behave like the official Python for Windows distributions.
>
> These builds are now published **without the -shared suffix** [emphasis added].

**How to reproduce**

`pipx install ... --fetch-missing-python --python 3.12` fails with a `PipxError: Unable to acquire a standalone python build matching 3.12.`

The same thing happens for Python 3.13.

**Expected behavior**

`pipx` should have installed the requested version of python.

dukecat0:

## Summary of changes
Closes #1630 

## Test plan

Tested by running

```
pipx install black --python 3.11 --fetch-missing-python
```


uranusjr (approved):

huxuan (approved):

tmpho079-pixel (commented):
`[065a189b248f8c663646e5dc05150be32eb63ca9``](`url`)` #1629 **honour [os**](url)

Issue #1069, opened 2023-10-01
Show pip download progress bar


tae-soo-kim:
**How would this feature be useful?**

`pip install` shows a progress bar which is helpful when the package has a large size and downloading may take a long time. But `pipx install` omits that progress bar even with `--verbose`.

**Describe the solution you'd like**

Give users an option to show the progress bar. An easy way to do this is to pipe `pip` output into `pipx`.

**Describe alternatives you've considered**

gaborbernat:
PR welcome.

kofalt:
Does the [linked branch above](https://github.com/pypa/pipx/compare/main...inoc603:pipx:feat/pipe-pip-output) sufficiently address the issue?
Looks reasonable at a glance, wondering if they simply forgot to file a PR.

Stopping by to mention this issue hit me pretty hard, installing ansible and after trying `--verbose --verbose --verbose` was convinced `pipx` was broken... ended up checking `slurm` to see that the package download is just very slow tonight.

fluzzywuzzy:
Would this be a valid solution? (Work in progress right now, but works with pipx install)

I found it difficult to pipe the progress bars themselves, but just showing the actual pip output during animation is possible, and might be considered a fix for this issue?

Please see for a draft of the solution: [fluzzywuzzy:1069-pipe-pip-progress-bar](https://github.com/fluzzywuzzy/pipx/tree/1069-pipe-pip-progress-bar)

If it looks like a good start, I will finnish it up and open a PR.

I assume it might as well be integrated with all calls to animate for clarifying progress?

NonVideri:
This is still very relevant.

dukecat0:
I am trying to stream the pip output if `--verbose` is passed, but the main drawback is that pipx won't be able to capture the stdout/stderr and analyze the error. Folks may have check the traceback by themselves.
[![asciicast](https://asciinema.org/a/SAxmmZQr1OPowFHB3HktVgmYn.svg)](https://asciinema.org/a/SAxmmZQr1OPowFHB3HktVgmYn)

PR #1640, opened 2025-05-18
#1069 Pipe pip progress to pipx

fluzzywuzzy:

## Summary of changes
Refactor `pipx.util.run_subprocess` to use `subprocess.Popen` instead, to let us stream terminal output.
Send captured output stream to `pipx.animate.animate` and forward to `pipx.animate.print_animate` where it is included in the printed animation output.

**HELP**
I am not sucessfull in getting the progress-bars themselves from the pip output. I would appreciate som input on if that is necessary, or if this added information is enough to consider #1069 solved? (It does not pipe progress bars, but the spirit of the issue is that it is dfficult to understand if the installation is stuck or just slow).

If this is sufficient, it should just be a matter of deciding on which `with animate` calls it should be included.

## Test plan
Run pipx install to verify output manually, and make sure all test suites pass.

To test, run:

```
# pipx install {PACKAGE_NAME}
```


fluzzywuzzy:
*Force pushed a commit to overwrite previous with the same name, as replaced commit accidentally deleted file: `test_run.py`

gaborbernat (changes requested):
Can you rebase and fix conflicts? Thanks!

gaborbernat:

Issue #1652, opened 2025-07-21
Checksum validation failing when fetching missing python


thmahe:
**Describe the bug**

pipx is unable to validate archives from https://github.com/astral-sh/python-build-standalone since [20250708](https://github.com/astral-sh/python-build-standalone/releases/tag/20250708) release.

**Note from `python-build-standalone` release**
> Remove .sha256 files: these artifacts are no longer published to GitHub Releases; unfortunately, they put us above the newly enforced 1000 file limit for GitHub Release artifacts. A single SHA256SUMS file with all of the checksums is still published.

**How to reproduce**

```
$ pipx install xxx --global --python 3.11 --fetch-missing-python
pipx >(setup:1110): pipx version is 1.7.1
pipx >(setup:1111): Default python interpreter is '/usr/bin/python3.13'
pipx >(find_unix_command_python:74): Command `python3.11` was not found on the system
Traceback (most recent call last):
  File "/usr/local/bin/pipx", line 8, in <module>
    sys.exit(cli())
             ~~~^^
  File "/usr/local/lib/python3.13/dist-packages/pipx/main.py", line 1176, in cli
    return run_pipx_command(parsed_pipx_args, subparsers)
  File "/usr/local/lib/python3.13/dist-packages/pipx/main.py", line 252, in run_pipx_command
    interpreter = find_python_interpreter(
        args.python or DEFAULT_PYTHON, fetch_missing_python=fetch_missing_python
    )
  File "/usr/local/lib/python3.13/dist-packages/pipx/interpreter.py", line 104, in find_python_interpreter
    return download_python_build_standalone(python_version)
  File "/usr/local/lib/python3.13/dist-packages/pipx/standalone_python.py", line 84, in download_python_build_standalone
    _unpack(full_version, download_link, archive, download_dir)
    ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/dist-packages/pipx/standalone_python.py", line 115, in _unpack
    expected_checksum = urlopen(checksum_link).read().decode().rstrip("\n")
                        ~~~~~~~^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/urllib/request.py", line 189, in urlopen
    return opener.open(url, data, timeout)
           ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
[...]

PR #1655, opened 2025-07-25
Use `SHA256SUMS` file for standalone python checksum validation

thmahe:
**Bugfix for  #1652**

* Support new release content from python-build-standalone project

No entry in `changelog.d/` since this patch is a bugfix.

## Test plan

Tested by running

```
$ pipx install --python 3.11 --fetch-missing-python
```

### DIFF
--- a/src/pipx/standalone_python.py
+++ b/src/pipx/standalone_python.py
@@ -112,19 +112,17 @@ def _unpack(full_version, download_link, archive: Path, download_dir: Path):
             checksum = hashlib.sha256(python_zip.read()).hexdigest()
 
         # Validate checksum
-        checksum_link = "/".join(download_link.split('/')[:-1] + ["SHA256SUMS"])
+        checksum_link = "/".join(download_link.split("/")[:-1] + ["SHA256SUMS"])
         expected_checksums = urlopen(checksum_link).read().decode().rstrip("\n")
-        release_file = unquote(download_link.split('/')[-1])
-        
-        match = re.findall(rf"(\S+)\s+{re.escape(release_file)}" , expected_checksums, re.MULTILINE)
-        
+        release_file = unquote(download_link.split("/")[-1])
+
+        match = re.findall(rf"(\S+)\s+{re.escape(release_file)}", expected_checksums, re.MULTILINE)
+
         if match is None or len(match) != 1:
-            raise PipxError(
-                f"Unable to retrieve checksum from python-build-standalone for python {full_version}"
-            )
-        
+            raise PipxError(f"Unable to retrieve checksum from python-build-standalone for python {full_version}")
+
         expected_checksum = match[0]
-        
+
         if checksum != expected_checksum:
             raise PipxError(
                 f"Checksum mismatch for python {full_version} build. Expected {expected_checksum}, got {checksum}."


13steinj:
:wave: I was just bit by the issue that this PR resolves.

Just a comment: I don't know what would be best, to use `SHA256SUMS` or, since the GitHub API [already provides the digests as of June 3](https://github.blog/changelog/2025-06-03-releases-now-expose-digests-for-release-assets/) it might instead be best to save the digest with the initial API call that populates the local index, and just check the digest against that, instead of separately downloading the `SHA256SUMS` file each time?

See something like my change on my branch: https://github.com/13steinj/pipx/commit/4bb24ed8fddbaf846ee5398cc4810fe0ebd2c0db (I did this quick and dirty trying to figure out what was going on, before I stumbled upon this).

thmahe:
Hi @13steinj,

I initially started my work by considering using checksum computed by Github and available as you said in the initial API call that `pipx` is making.

Requires more changes and in the end, should we trust checksum provided by GitHub or the one provided by the project itself ?
Personally made my choice: small change-set & checksum from the people releasing python standalone.

13steinj:
Yeah just bringing up as an option in case maintainers prefer it. I just want _something_ merged and a hotfix released sooner rather than later.

> should we trust checksum provided by GitHub or the one provided by the project itself ?

Me personally, I don't know. I mean, if you trust automated checksums by the hosting platform you can argue you're more secure than the alternative. It would solve (similar, not the same) kind of issues that happened with the (relatively recent) xz supply chain attack.

dukecat0:
@thmahe @13steinj  Thanks for your contributions! Personally I would prefer the solution provided by @13steinj as it's more efficient and an extra file is not required.
Also, with https://github.com/13steinj/pipx/commit/4bb24ed8fddbaf846ee5398cc4810fe0ebd2c0db, I think it's a small change as well so it shouldn't be an issue. 👍

13steinj:
@dukecat0 I don't want to step on any toes here, am I good to clean my branch up a bit (mainly the whole `link[0]` thing which is odd instead of properly unpacking it to a descriptive var name) and make a PR?

I'd also like to:

* confirm that older downloads than June 3 contain the digests as well (unclear to me) otherwise keep downloading the files separately as a fallback mechanism if github isn't providing a digest
* tag the index/cache file downloaded with a simple incrementing schema number; force a re-download on either no schema number found or an old one found, re-download as well

dukecat0:
> am I good to clean my branch up a bit (mainly the whole link[0] thing which is odd instead of properly unpacking it to a descriptive var name) and make a PR?

Yes sure, feel free to work on it! @thmahe Still thanks for your work on this and raising the issue!

> confirm that older downloads than June 3 contain the digests as well (unclear to me) otherwise keep downloading the files separately as a fallback mechanism if github isn't providing a digest

Don't worry, it's not an issue. The code guarantees that the freshness of index is always within 30 days:
https://github.com/pypa/pipx/blob/849d3f40baf881fa908226faf64aa85b8f0c2c0d/src/pipx/standalone_python.py#L125-L142

thmahe:
Superseded by #1662

Issue #1607, opened 2025-02-17
pipx run --with=extra_dependency my_tool


cclauss:
**How would this feature be useful?**

Streamline three commands into a single intuitive command that is similar to the `uvx` syntax.

**Describe the solution you'd like**

Desired:
```bash
pipx run --with=dataclass_wizard pytest
```
Current:
```bash
pipx install pytest
pipx inject pytest dataclass_wizard
pytest
```

**Describe alternatives you've considered**
```
uvx run --with=dataclass_wizard pytest  # But `pipx` is pre-installed in GitHub Actions
```

PR #1666, opened 2025-09-09
feat: add --with option to "pipx run" command

pabloyoyoista:

## Summary of changes

Added the option in argparse for `pipx run`, and used `inject_dep`

## Test plan

I've used test-driven development, and there is a test for a simple case that should prove this feature is working

TODOs: 
- [x] If this approach is considered acceptable, refactor `_download_and_run` to just download, so that we can inject the dependency also in the cases where the `if venv.has_app == True` branch  in `run_package`
- [x] Given this approach is considered acceptable, add parametrized tests

Closes https://github.com/pypa/pipx/issues/1607

### DIFF
--- a/changelog.d/1607.feature.md
+++ b/changelog.d/1607.feature.md
@@ -1,0 +1,1 @@
+Add `--with` flag to `pipx run` to allow injecting dependencies
--- a/src/pipx/commands/run.py
+++ b/src/pipx/commands/run.py
@@ -8,12 +8,13 @@ import urllib.parse
 import urllib.request
 from pathlib import Path
 from shutil import which
-from typing import List, NoReturn, Optional, Union
+from typing import List, NoReturn, Optional, Tuple, Union
 
 from packaging.requirements import InvalidRequirement, Requirement
 
 from pipx import paths
 from pipx.commands.common import package_name_from_spec
+from pipx.commands.inject import inject_dep
 from pipx.constants import TEMP_VENV_EXPIRATION_THRESHOLD_DAYS, WINDOWS
 from pipx.emojis import hazard
 from pipx.util import (
@@ -111,6 +112,7 @@ def run_script(
 def run_package(
     app: str,
     package_or_url: str,
+    dependencies: List[str],
     app_args: List[str],
     python: str,
     pip_args: List[str],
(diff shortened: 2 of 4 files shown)


changelog.d/1607.feature.md
```diff
@@ -0,0 +1 @@
+Add `--with` flat to `pipx run` to allow injecting dependencies
```
dukecat0, line 1:
```suggestion
Add `--with` flag to `pipx run` to allow injecting dependencies
```

dukecat0 (changes requested):
Thanks for the contribution!

changelog.d/1607.feature.md
pabloyoyoista: fixed!

pabloyoyoista:
Thanks a lot for running the tests! I'm a bit confused about the macos failure though (https://github.com/pypa/pipx/actions/runs/17588256767/job/50365801078?pr=1666). It seems related to the code here, but I fail to see why it fails there. Any tips you might be able to have?

dukecat0:
> It seems related to the code here

It's not related to your code here, it also failed in every other PRs. Don't worry! :D

src/pipx/commands/run.py
```diff
@@ -156,6 +158,7 @@ def run_package(
     _prepare_venv_cache(venv, bin_path, use_cache)
 
     if venv.has_app(app, app_filename):
+        # TODO: Should refactor and consider injecting here
```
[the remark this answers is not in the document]
dukecat0: >  I'll go ahead and add tests to try have the most code coverage possible

The new logic looks good to me :D Please go ahead!

pabloyoyoista:
CI green :partying_face:

tests/test_run.py
```diff
@@ -443,3 +443,26 @@ def test_run_local_path_entry_point(pipx_temp_env, caplog, root):
...
+    run_pipx_cli_exit(["run", "--with", "black", "pycowsay", "--help"])
+    captured = capsys.readouterr()
+    assert "injected package black into venv pycowsay" in captured.out
+
+
+@mock.patch("os.execvpe", new=execvpe_mock)
+def test_run_with_cache(capsys, caplog):
+    # Maybe there's a better way to remove the previous venv cache?
+    run_pipx_cli_exit(["run", "--no-cache", "pycowsay", "cowsay", "args"])
+    run_pipx_cli_exit(["run", "pycowsay", "cowsay", "args"])
```
dukecat0, line 459: Are there any reasons that we need this? 👀 I think the one below performs the same action?

PR #1682, opened 2025-10-07
Add PIPX_DISABLE_SHARED_LIBS_AUTO_UPGRADE environment variable to disable automatic shared library upgrades

### LINKED ISSUE #1650: Disable auto-upgrading shared libraries
I am trying to update a Pipx install that has to stay on Python 3.8.  Given 3.8 is end of life, pip has dropped support for it and has introduced code that requires >Python 3.9.  That means when I try to upgrade this install, I get errors from pip (specifically `TypeError: 'ABCMeta' object is not subscriptable`).

I have tried forcing the shared library version of pip back to a version that supports Python 3.8 [per the docs](https://pipx.pypa.io/stable/examples/#pipx-upgrade-shared-examples):

```
pipx upgrade-shared --pip-args 'pip<25'
```

However, when I run install or upgrade, the shared libraries are always immediately upgraded before running the pip install, producing the same error.

The docs mention something about how this should pin the pip version 'temporarily, until the next automatic upgrade, if that is not explicitly turned off'.

How do I explicity turn off automatic upgrades?

G26karthik:
## Summary
Fixes #1650

This PR implements the feature requested in #1650 by adding a new environment   
variable `PIPX_DISABLE_SHARED_LIBS_AUTO_UPGRADE` that allows users to disable   
automatic upgrades of shared libraries (pip, setuptools, wheel).

## Motivation
Users who need to maintain compatibility with older Python versions (e.g., Python 3.8)
can encounter errors when pip automatically upgrades to versions that no longer support
their Python version. This environment variable provides a way to disable the automatic
upgrade behavior while still allowing manual upgrades via `pipx upgrade-shared`.

## Changes Made
1. Added `PIPX_DISABLE_SHARED_LIBS_AUTO_UPGRADE` constant to `src/pipx/constants.py`
2. Modified `src/pipx/venv.py` to check the environment variable before auto-upgrading
3. Added documentation in `docs/examples.md` explaining the new feature
4. Added test in `tests/test_shared_libs.py` to verify the behavior

## Usage
```bash
# Linux/MacOS
export PIPX_DISABLE_SHARED_LIBS_AUTO_UPGRADE=1
pipx install <package>  # Will NOT auto-upgrade shared libraries

# Windows (PowerShell)
$env:PIPX_DISABLE_SHARED_LIBS_AUTO_UPGRADE=1
pipx install <package>  # Will NOT auto-upgrade shared libraries
```


G26karthik:
@Maintainers Pls Review and Approve for Tests

gaborbernat (changes requested):
Please rebase against main.

gaborbernat:

PR #1684, opened 2025-10-18
TST: add explicit support for Python 3.14

neutrinoceros:

## Test plan

Tested by running

```
nox -s tests
```


dukecat0:
Thanks for the contribution, but it looks like some of the test dependencies haven't supported Python 3.14 yet. 👀

neutrinoceros:
Indeed. I didn't identify this issue locally because apparently I ran the test suite on the only platform where everything builds fine (macOS). Let me list the issues I'm parsing from log here:
- [`cmarkgfm`](https://github.com/theacodes/cmarkgfm) fails to build, at least on Linux
- `test_run_with_windows_python_version` fails (not sure where the bug is coming from)

neutrinoceros:
the relevant PR upstream is already merged https://github.com/theacodes/cmarkgfm/pull/77
what's lacking is a release

neutrinoceros:
The remaining failure looks genuine to me. I even think it was probably already discovered once, but hidden away by updating the test itself instead of fixing the bug (#1647). Of course, this is just my conclusion from a *very* quick review, and I could be entirely mistaken.

uranusjr:
Looks like issues are essentially caused by pycowsay not working on Windows with newer Python. Probably a good idea to fix the tests to use a different package instead.

neutrinoceros:
Hum, that's really surprising for a package that has a pure-python wheel.
I'm inclined to go and dig a little deeper before we resort to ditching it entirely, since it's used everywhere in pipx's documentation

The log from CI is not particularly useful however
```
ERROR    pipx.util:util.py:347 Fatal error from pip prevented installation. Full pip output in file:
    C:\Users\runneradmin\AppData\Local\Temp\pytest-of-runneradmin\pytest-0\test_do_not_wait_for_input0\subdir\pipxhome\logs\cmd_2025-11-03_02.34.38_pip_errors.log
```

is there any way I could retrieve this file or see its content ?

neutrinoceros:
fixed

cclauss:
Cannot replicate the `pycowsay` problem on GitHub Actions...
* cs01/pycowsay#5

gaborbernat (changes requested):
CI is failing 🤔

neutrinoceros:
I'm aware. This is already being discussed above. I don't plan to dig further myself at the moment, so if that's what drafting the PR signals, you might as well just close it.

cclauss:
https://github.com/pypa/pipx/blob/5a45b9aba1dc919a2286048d5eb5fd073a6596a0/tests/test_run.py#L480-L481
Change this hard-coded value?
```diff
- run_pipx_cli_exit(["run", script.as_uri(), "--python", "3.13"]) 
- assert "3.13" in out.read_text() 
+ run_pipx_cli_exit(["run", script.as_uri(), "--python", "3.14"]) 
+ assert "3.14" in out.read_text()
```

gaborbernat:

PR #1689, opened 2025-11-10
chore: pytest log_level is better than log_cli_level

henryiii:
Committed via https://github.com/asottile/all-repos

See https://github.com/scientific-python/cookie/issues/679


gaborbernat (approved):

gaborbernat (changes requested):
Can you rebase and fix the conflict?

gaborbernat:
@henryiii seems the CI is unhappy with your change here 🤔

gaborbernat (approved):

PR #1690, opened 2025-11-17
feat: add --upgrade and --upgrade-strategy options

### LINKED ISSUE #1572: pipx install --update
`pipx install`:
- install if missing.
- if available and in the right or wrong version (spec not respected), does nothing

`pipx install --force`:
- always reinstall, even if available and if spec respected

`pipx update`:
- always update, even when spec is respected
- if not installed, fail

`pipx update --install`:
- always update, even when spec is respected
- if not installed, install.

What I need is a command that will:
- if not install, install respecting the provided spec
- if installed but the spec is not respected, update to the spec, allowing downgrade
- if installed and spec is respected, does nothing, not even a network connectivity

Example:
```bash
$ pipx install --upgrade pylint<7
```

- if the current pylint venv respects `<7`, does nothing.
- If the current pylint venv does not respect `<7`, update
- if no pylint installed, install to version `<7`

j-krl:
## Summary of changes

Resolves #1572

- Adds the `--upgrade` option to `inject` and `install`, mirroring the pip install option.
- Adds `--upgrade-strategy` to `upgrade` as well as all subcommands where `--upgrade` has been added above. Upgrade strategy options are the same as pip's where `eager` upgrades all dependencies of the main package regardless of the upgrade being required, and `only-if-needed` upgrades only dependencies that require an upgrade.

## Motivation

Provides a way to conditionally upgrade only if the currently installed version does not match the passed spec. Currently, `pipx upgrade` upgrades to the latest available version and does not allow an installation spec. `pipx install --upgrade` allows this flexibility. This also adds parity with underlying pip commands which is a usability benefit.

For example, if I have `black` v24.1.0 installed and I want to upgrade to no higher than the latest minor version of v24 I can now run `pipx install --upgrade 'black<25'`, which was not possible previously.

## Test plan

PENDING


j-krl:
Maintainers: This PR is not fully complete as tests & docs have not been updated. Seeking approval on my general approach and that this feature is still desired before implementing remaining requirements.

j-krl:
Hey maintainers, still looking to get this PR reviewed, maybe @dukecat0 do you have any input?

gaborbernat (changes requested):
Can we add tests here please?

gaborbernat:

PR #1694, opened 2025-11-19
Fix uninject removing shared dependencies from other injected packages

CorentinDurand:
## Summary

When multiple injected packages share a dependency in the same pipx venv,
`pipx uninject` could uninstall that shared dependency, even if it is still
required by other injected packages. This could break the remaining packages.

This PR makes `uninject` more conservative when cleaning up dependencies.

## Changes

- In `uninject_dep`, after uninstalling the injected package, we still compute
  the set of packages that became "not required" using
  `venv.list_installed_packages(not_required=True)`, but before uninstalling
  them we:
  - Build a set of dependencies still required by:
    - the main package of the venv, and
    - all other injected packages (excluding the one being uninjected).
  - Only uninstall dependencies that are **not** in this "protected" set.

- The user-facing behaviour is unchanged when there are no shared dependencies,
  but shared dependencies used by other packages in the venv are now preserved.

- Added a regression test `test_uninject_keeps_shared_dependencies` to ensure
  that uninjection of one injected package does not remove another injected
  package from the same venv.

- Added a changelog fragment describing the bugfix.

## Testing

Local (Windows 11):

- `python -m nox -s tests-3.10 -- -k "uninject"`

Result:

- All `uninject`-related tests pass:
  - 4 passed, 1 skipped (the global test skipped on Windows).
- The `cover` nox session fails locally because only a subset of tests is run
  (global coverage 56% < fail-under=70). I expect full coverage to be validated
  by the CI when running the complete test suite.

### DIFF
--- a/changelog.d/1672.bugfix.md
+++ b/changelog.d/1672.bugfix.md
@@ -1,0 +1,1 @@
+Prevent uninject from removing dependencies still required by other packages.
--- a/src/pipx/commands/uninject.py
+++ b/src/pipx/commands/uninject.py
@@ -78,6 +78,8 @@ def uninject_dep(
 
     new_resource_paths = get_include_resource_paths(package_name, venv, local_bin_dir, local_man_dir)
 
+    deps_of_uninstalled: Set[str] = set()
+
     if not leave_deps:
         orig_not_required_packages = venv.list_installed_packages(not_required=True)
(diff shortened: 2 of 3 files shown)


gaborbernat (changes requested):
Can you fix the CI please?

gaborbernat:

PR #1699, opened 2025-12-12
Make list_installed_packages handle pip errors (#1698)

### LINKED ISSUE #1698: Misleading exception when pip subprocess crashes
**Describe the bug**

When pipx install on a misconfigured system, pipx raised an exception about a JSON decode error. The cause of the exception turned out to be an unhandled `pip list` subprocess crash.

Apparently, running `pip list` in a broken or improperly configured environment can result in pip crashing without writing to stdout. The `pipx.util.run_subprocess` function can silently eat that subprocess error, relying on the caller checking the subprocess return code. When `pipx.venv.list_installed_packages` runs into this, the call to `json.loads` (<https://github.com/pypa/pipx/blob/2146ab6d3c4fb2fe21e30f9df0b8343539b0b98c/src/pipx/venv.py#L398>) decodes an empty string, resulting in pipx crashing with a misleading error about malformed JSON:

```
```

This is the same error message as #393, but in this case the traceback pointed to venv.py so it's a different issue altogether.

In normal execution, `pip list --format=json` will print `[]` to stdout in the event that there aren't any installed packages, so an empty stdout only happens when pip crashes.

**How to reproduce**

Unfortunately I don't have a traceback or a reproduction for this issue. The original bug was intermittent, and I lost the original traceback before I was able to track down the cause.

**Expected behavior**

Pipx should raise a useful error message informing the user that there was an issue running the `pip list` command.

absolutely-vivid:

closes #1698

## Summary of changes

Running `pip list` in a broken or improperly configured environment can result in pip crashing without providing output. `run_subprocess` can eat that subprocess error leading to `json.loads` decoding an empty string, resulting in pipx crashing with a misleading error about malformed JSON.

This commit explicitly checks the return code of the pip process, providing a friendlier error message for easier debugging.

## Test plan

I don't think testing is needed here, the commit just changes which exception is raised in this edge case.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,6 +63,7 @@ test = [
   "pytest>=8",
   "pytest-cov>=4",
   "pytest-mock>=3.11.1",
+  "pytest-subprocess>=1.5.4",
   "pytest-xdist>=3.5",
   "setuptools>=68; python_version>='3.12'",
   "watchdog>=4",
--- a/tests/test_list.py
+++ b/tests/test_list.py
@@ -245,29 +245,14 @@ def test_list_pinned_packages_include_injected(pipx_temp_env, monkeypatch, capsy
     assert "black 22.8.0 (injected in venv pylint)" in captured.out
 
 
-def test_list_installed_packages_error(pipx_temp_env, monkeypatch, tmp_path):
-    called_with = None
+def test_list_installed_packages_error(monkeypatch, tmp_path, fake_process):
+    fake_venv = venv.Venv(tmp_path / "fake_venv")
+    pip_list_args = [str(fake_venv.python_path), "-m", "pip", "list", "--format=json"]
 
-    def mockreturn(args):
-        nonlocal called_with
-        called_with = args
-
-        class MockPopen:
-            def __init__(self, args):
-                self.args = args
-                self.returncode = 1
-                self.stderr = "unit test stderr"
-
-        return MockPopen(args)
-
-    # we don't want to actually spawn a subprocess, we just want to test the
-    # error handling for a failed pip list command
-    monkeypatch.setattr(venv, "run_subprocess", mockreturn)
+    fake_process.register(pip_list_args, returncode=1, stderr="unit test stderr")
 
     # don't reformat the error message, so we can compare it to a known string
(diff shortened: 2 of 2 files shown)


src/pipx/venv.py
```diff
@@ -395,6 +395,17 @@ def list_installed_packages(self, not_required=False) -> Set[str]:
...
             [str(self.python_path), "-m", "pip", "list", "--format=json"] + (["--not-required"] if not_required else [])
         )
+
+        if cmd_run.returncode != 0:
+            raise PipxError(
+                f"""
+                Failed to execute pip list - {cmd_run.args}. Process exited with return code
+                {cmd_run.returncode}.\n
+                stderr:\n
+                {"\t".join(cmd_run.stderr.split("\n"))}
```
absolutely-vivid: The specifics of the exception message might need to be changed, particularly around this line. I wasn't sure what the preferred style is, so I erred on the side of caution and just added as much context to the exception message as I could

changelog.d/1698.bugfix.md
```diff
@@ -0,0 +1 @@
+Make list_installed_packages handle pip subprocess errors
```
dukecat0:
```suggestion
Added error handling for `pip list` command failures
```

src/pipx/venv.py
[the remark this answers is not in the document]
absolutely-vivid: Oh that's way better, thank you!

dukecat0 (approved):
Thanks!

absolutely-vivid:
@gaborbernat Sure, I'd be happy to. What about it would you want to test? I could jut add a test that runs `pip list` in a way that fails and then check the stderr stream, but I'm not actually sure how to make pip list fail, I never actually managed to reproduce the original crash that lead to this bug report haha.

I guess I could just mock the `run_subprocess` function to return a fake failed process object, would that be sufficient?

absolutely-vivid:
I tried my hand at a test, it turned out to be easier than I thought. It's a little hacky though, so it might need some fixing up or style changes.

absolutely-vivid:
I found the source of the failing tests, and fixed it up with a less hacky string comparison. This version just monkeypatches PipxError to stop the error message from being wrapped in the first place.

PR #1701, opened 2025-12-23
Update README.md with installation instructions on Debian

jeremy-bryant:
Short and simple addition to README.md to clarify installation for Debian users


dukecat0:
Hello, personally I think this installation instruction is similar to the Ubuntu one? Perhaps it is not necessary to add an extra instruction here.

gaborbernat:
> Hello, personally I think this installation instruction is similar to the Ubuntu one? Perhaps it is not necessary to add an extra instruction here.

Agreed.

PR #1710, opened 2026-01-15
Stop `pipx run` from leaving bad temporary venvs

### LINKED ISSUE #1709: Bad temporary venvs are left by `pipx run` under certain conditions
If the dependency installation step fails when `pipx run` is invoked, subsequent calls to `pipx run` will fail, even if the dependencies can still potentially be successfully installed. This is because the inline script metadata specifying dependencies will not have changed, and so the temporary venv from the first attempt is reused.

**How to reproduce**

An example instance where the package installation step fails on a first `pipx run`, but could succeed on a second (if it weren't for this bug) is when a dependency in a private git repository is specified, but SSH keys have not yet been correctly set up.

```python
# /// script
# dependencies = [ "failure @ git+ssh://0.0.0.0/failure.git" ]
# ///

import failure
```

Invoking `pipx run` on the above script initially causes a package installation error (as expected):

```
    [...]

pip seemed to fail to build package:
    'failure'

    error: subprocess-exited-with-error
    ERROR: Failed to build 'failure' when git clone --filter=blob:none --quiet ssh://0.0.0.0/failure.git [...]
Error installing failure@ git+ssh://0.0.0.0/failure.git.
```

Invoking `pipx run` on the above script for a second time raises a `ModuleNotFoundError`, instead of attempting package installation again. (In this case, the address `0.0.0.0` will again not respond. However, consider the case where permissions to a private git repository were changed.)

```
  File "[...]", line 5, in <module>
    import failure
ModuleNotFoundError: No module named 'failure'
```

**Expected behavior**

If the git repository access were fixed, then re-running `pipx run` on the above script should instead attempt to re-install the packages, and proceed to run the script correctly. This can be fixed by `pipx run` proactively removing the temporary venv if package installation fails.

sam-sw:
If the dependency installation step fails when `pipx run` is invoked, subsequent calls to `pipx run` will fail, even if the dependencies can still potentially be successfully installed. This is because the inline script metadata specifying dependencies will not have changed, and so the temporary venv from the first attempt is reused.

This commit adds logic to check the dependency installation step and, if it failed, mark the temporary venv as out of date and for deletion. This means that subsequent calls to `pipx run` will recreate the venv, and attempt to reinstall the dependencies. This commit also adds a unit test.

Fixes #1709.

### DIFF
--- a/changelog.d/1709.bugfix.md
+++ b/changelog.d/1709.bugfix.md
@@ -1,1 +1,1 @@
-Stop `pipx run` from leaving bad temporary venvs.
+Stop `pipx run` from leaving bad temporary venvs when first installation was unsuccessful.


sam-sw:
I would be grateful for any comments on this! Testing locally, it has solved a number of issues I encountered with changing private git repository access permissions. It is helpful not to have to manually remove the temporary venv directories.

dukecat0:
Thanks for the contribution! The logic looks good to me, and perhaps a changelog entry should be added for this change?

sam-sw:
I have added a short changelog entry now. Thank you for reviewing this!

changelog.d/1709.bugfix.md
```diff
@@ -0,0 +1 @@
+Stop `pipx run` from leaving bad temporary venvs.
```
dukecat0:
```suggestion
Stop `pipx run` from leaving bad temporary venvs when first installation was unsuccessful.
```

PR #1715, opened 2026-02-13
feat(logs): make max log count configurable

### LINKED ISSUE #1570: Have max_logs able to be set from env variable with a default value of 10
Executing more than 10 python scripts (cicd jobs) concurrently will lead to pipx failing 

The current default value to `max_logs` is 10, this solution involves reading from an environment variable, `MAX_PIPX_LOGS`. If the variable is not present the current values is set to 10. 

`max_logs = int(os.getenv("MAX_PIPX_LOGS", 10))`

Possibly allowing the value to be set by cmdline arg, however an environment variable is simpler for automated environments.

LouisLau-art:
Fixes #1570.

- Read MAX_PIPX_LOGS from the environment to control how many pipx command log files are kept (default: 10).
- Document the variable in the CLI help and in `pipx environment`.
- Add unit tests for env parsing.

Test:
- .venv/bin/python -m pytest --net-pypiserver tests/test_main.py tests/test_environment.py -k "setup_log_file_max_logs or test_cli"


src/pipx/commands/environment.py
```diff
@@ -18,6 +18,7 @@
     "PIPX_FETCH_MISSING_PYTHON",
     "PIPX_USE_EMOJI",
     "PIPX_HOME_ALLOW_SPACE",
+    "MAX_PIPX_LOGS",
```
dukecat0: IMO `PIPX_MAX_LOGS` is better here? So it is consistent with other existing environment variables.

gaborbernat: > IMO `PIPX_MAX_LOGS` is better here? So it is consistent with other existing environment variables.

Agreed.

gaborbernat:

PR #1724, opened 2026-03-14
Add MAX_PIPX_LOGS environment variable support

### LINKED ISSUE #1570: Have max_logs able to be set from env variable with a default value of 10

luminexo:
## Summary

This PR adds support for the `MAX_PIPX_LOGS` environment variable to configure the maximum number of log files to keep, as requested in #1570.

## Changes

- Modified `_setup_log_file()` in `src/pipx/main.py` to read `MAX_PIPX_LOGS` from environment
- Default value remains 10 when the variable is not set

## Usage

```bash
export MAX_PIPX_LOGS=20
pipx install some-package
```

## Testing

- Syntax check passed: `python3 -m py_compile`
- The change is minimal (1 line) and follows the pattern used elsewhere in the codebase

Fixes #1570


dukecat0:
Btw, maybe it's more consistent with other existing environment variables if it's named as `PIPX_MAX_LOGS`?

PR #1732, opened 2026-03-17
Fix CLI parameter parsing for inject command (Python < 3.12)

### LINKED ISSUE #1444: cli parameter parsing is corrupted by strict positional
In version 1.6.0, the following command cannot be used

**Describe the bug**

In version 1.6.0, the following command cannot be used.

```
pipx inject ansible-core --force ansible
```
```
error: unrecognized arguments: ansible
```

While adjusting the order can work

```
pipx inject ansible-core ansible --force
```

But this will create compatibility issues, the current way used in devcontainers is

[features/src/pipx-package/install.sh at 9a1d24b27b2d1ea8916ebe49c9ce674375dced27 · devcontainers-contrib/features](https://github.com/devcontainers-contrib/features/blob/9a1d24b27b2d1ea8916ebe49c9ce674375dced27/src/pipx-package/install.sh#L152-L152)

[[Bug]: Failed to build ansible feature · Issue #607 · devcontainers-contrib/features (github.com)](https://github.com/devcontainers-contrib/features/issues/607)

This is related to [HIGH: pipx install --global order broken in 1.6.0 · Issue #1443 · pypa/pipx](https://github.com/pypa/pipx/issues/1443)

J926L:
### Description

This PR fixes a bug in CLI parameter parsing for the inject command that occurs in Python versions older than 3.12.

#### The Problem
In pipx inject, the dependencies argument is defined with nargs='*'. In Python < 3.12, argparse stops parsing positional arguments as soon as it encounters an optional argument (like --force or --verbose). This leads to unrecognized arguments errors when a user provides dependencies after an option.

#### The Fix
Introduced an intermixed parsing strategy in the cli() entry point for the inject command.

Fixes #1444

### DIFF
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -1021,7 +1021,7 @@ def get_command_parser() -> Tuple[argparse.ArgumentParser, Dict[str, argparse.Ar
         description="Print instructions on enabling shell completions for pipx",
         parents=[shared_parser],
     )
-    return parser, subparsers_with_subcommands
+    return parser, subparsers.choices
 
 
 def delete_oldest_logs(file_list: List[Path], keep_number: int) -> None:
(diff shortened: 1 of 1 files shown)


gaborbernat (changes requested):
Must come with tests and changelog.

gaborbernat:

PR #1762, opened 2026-03-21
Implement the command "exec" which executes a given app in the venv of a package

mardukbp:
This PR implements the feature described in #1726. 


uranusjr:
I wonder if this can be implemented as a flag on `pipx run`. They are mentally similar, and tt seems to me the arguments don’t conflict. (Just my personal opinion.)

mardukbp:
> I wonder if this can be implemented as a flag on `pipx run`. They are mentally similar, and tt seems to me the arguments don’t conflict. (Just my personal opinion.)

You are absolutely right! In my opinion, the flag --no-cache should execute an app within a venv created by pipx install. Which is not cached, but permanent. However, this flag just rebuilds the cached venv created by pipx run, which is not what --no-cache means :) Maybe a new flag is needed for this, like --rebuild-cache.

gaborbernat:

PR #1779, opened 2026-04-02
package_is_path: do not check if path exists

### LINKED ISSUE #1778: Do not check if the package is a path in the current working directory
**Issue**

**Environment**

Provide at least:

- OS: Arch Linux
- Shell: Bash

```
```

StaticRocket:
Having separators is the real indicator that the user is doing something wrong. We shouldn't unnecessarily error out in the off chance the user has a directory that has the same name as a package in the current working directory.

- [ ] ensured there are test(s) validating the fix

Not entirely sure if this needs a test or documentation update considering it's so small and there aren't many tests for specific functions outlined in the `main.py`.

Fixes: #1778

### DIFF
--- a/changelog.d/1778.bugfix.md
+++ b/changelog.d/1778.bugfix.md
@@ -1,0 +1,1 @@
+Do not check if the package name exists as path.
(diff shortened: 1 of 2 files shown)


StaticRocket:
Guess this is a dupe now.

PR #1788, opened 2026-04-17
First attempt

clin1234:
#  Summary
                                                                                                            
  - `PIPX_DEFAULT_PYTHON` previously only accepted a full path or a PATH-resolvable name; version strings like
   3.12 were rejected at import time with a cryptic error                                                   
  - _get_absolute_python_interpreter now delegates to find_python_interpreter (which handles version-style
  lookup via pythonX.Y naming convention) instead of shutil.which                                           
  - When resolution fails at module import time, the raw spec is stored and resolution is deferred to
  command execution, so pipx doesn't crash on startup when a standalone Python hasn't been downloaded yet   
  - shared_libs.create now re-resolves DEFAULT_PYTHON at call time, catching any deferred resolution errors
  with a clean PipxError                                                                                    
                  
  Test plan                                                                                                 
                  
  - test_env_python_by_version — verifies X.Y version strings resolve correctly on Unix                     
  - test_default_python_env_var_version_not_found — verifies an unavailable version defers gracefully
  instead of crashing at import                                                                             
  - test_default_python_env_var_full_path — verifies a full path still resolves correctly
  - Run pytest tests/test_interpreter.py -v locally to confirm all three new tests pass

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,97 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## What is pipx
+
+pipx installs Python CLI applications into isolated virtual environments and exposes their entry points on PATH. Each
+app gets its own venv under `PIPX_HOME/venvs/`, preventing dependency conflicts. `pipx run` additionally supports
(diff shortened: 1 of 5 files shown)


gaborbernat:
Closing in favor of https://github.com/pypa/pipx/pull/1791

PR #1790, opened 2026-04-17
fix: remove exists() check from package_is_path validation

### LINKED ISSUE #1778: Do not check if the package is a path in the current working directory

bobo-xxx:
## Bug

The `package_is_path()` function in `main.py` checks if a package name looks like a path by:
1. Checking for path separators (`/` or `\`)
2. Checking if a file/directory with that name exists in the current working directory

The second check is too aggressive - if a user has a directory named `commit-check` in their current working directory, pipx incorrectly rejects any package operation with `commit-check` as a package name:

```
Error: 'commit-check' looks like a path.
Expected the name of an installed package.
```

## Fix

Remove the `Path(package).exists()` check from `package_is_path()`. Only check for path separators, which are unambiguous indicators of a path:

```python
# Before
if os.path.sep in package or Path(package).exists():

# After  
if os.path.sep in package or (os.path.altsep and os.path.altsep in package):
```

A valid package name like `commit-check` will never contain a path separator, so the separator check alone is sufficient.

Fixes #1778

### DIFF
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -214,7 +214,7 @@ def package_is_url(package: str, raise_error: bool = True) -> bool:
 
 
 def package_is_path(package: str):
-    if os.path.sep in package or Path(package).exists():
+    if os.path.sep in package or (os.path.altsep and os.path.altsep in package):
         raise PipxError(
             pipx_wrap(
                 f"""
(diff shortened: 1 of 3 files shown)


bobo-xxx:
This PR has the same checksum bug as #1789 — stripping the sha256: prefix breaks _unpack() validation. The original code is correct. Please close or revert.

bobo-xxx:
Closing — duplicate of #1789 which I'm closing. The digest fix is correct, my earlier comment was wrong.

PR #1797, opened 2026-04-23
fix(main): don't treat existing CWD dir as package path

### LINKED ISSUE #1778: Do not check if the package is a path in the current working directory

Bojun-Vvibe:
Closes #1778

## Repo
pypa/pipx

## Issue
https://github.com/pypa/pipx/issues/1778

## Root cause
`package_is_path()` in `src/pipx/main.py` rejected any package argument that either contained a path separator OR happened to exist as a file/directory in the current working directory (`Path(package).exists()`). This caused legitimate package names (e.g. `commit-check`) to be flagged as paths whenever a same-named directory existed alongside the user's shell.

## Fix
Dropped the `Path(package).exists()` clause; only the path-separator check remains, since a real PyPI package name can never contain `os.path.sep`.

## Regression test
`tests/test_main.py::test_package_is_path_ignores_existing_directory` — chdirs into a tmp_path, creates a `commit-check/` directory, then asserts that `main.package_is_path("commit-check")` does not raise.

## Risk
trivial

## Verification
skipped: project test deps (`userpath`, etc.) not installed in sandbox; manual reasoning confirms the new branch only triggers on `os.path.sep in package`, matching the issue author's proposed fix.

### DIFF
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -214,7 +214,7 @@ def package_is_url(package: str, raise_error: bool = True) -> bool:
 
 
 def package_is_path(package: str):
-    if os.path.sep in package or Path(package).exists():
+    if os.path.sep in package:
         raise PipxError(
             pipx_wrap(
                 f"""
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -40,3 +40,12 @@ def test_prog_name(monkeypatch, argv, executable, expected):
 def test_limit_verbosity():
     assert not run_pipx_cli(["list", "-qqq"])
     assert not run_pipx_cli(["list", "-vvvv"])
+
+
+def test_package_is_path_ignores_existing_directory(tmp_path, monkeypatch):
+    # Regression for #1778: a directory in CWD with the same name as a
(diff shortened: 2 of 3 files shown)


gaborbernat (dismissed):

gaborbernat:
@Bojun-Vvibe can you fix the CI?

Bojun-Vvibe:
Thanks for the review @gaborbernat! CI was failing on `test_uninstall_rejects_existing_directory_name` (added in #1753), which encodes the exact CWD-as-path behavior this PR removes. I've removed that test in 23ac8c6 since its expectation is now incompatible with the new behavior — let me know if you'd prefer it kept and re-shaped instead.

Bojun-Vvibe:
Hi @gaborbernat — quick note: the GitHub Actions checks here are empty because the workflow run is gated on "Approve and run workflows" (first-time contributor). If you click that on the latest commit, CI should run and we can confirm green. Thanks!

PR #1812, opened 2026-05-14
ci: declare workflow-level `contents: read` on 3 remaining workflows

arpitjain099:
Adds explicit `permissions: contents: read` at workflow scope to the 3 workflows that were still relying on the repo default. The repo's existing `pre-release.yml` / `release.yml` already declare per-job `permissions:`, so this brings the rest of `.github/workflows/` in line.

| Workflow | What it does |
|---|---|
| `create_tests_package_lists.yml` | generates offline test fixtures |
| `exhaustive_package_test.yml` | slow matrix of installs across OS + Python versions |
| `tests.yml` | regular CI test matrix |

None of the three calls a GitHub API beyond the initial checkout, so `contents: read` is sufficient.

## Why

CVE-2025-30066 (the March 2025 `tj-actions/changed-files` supply-chain compromise) exfiltrated `GITHUB_TOKEN` from workflow logs and the leaked token retained whatever scope was issued at the workflow level. Pinning per workflow caps that runtime authority irrespective of the repo or org default, gives drift protection if the default ever widens, and is credited per-file by the OpenSSF Scorecard `Token-Permissions` check.

YAML validated locally with `yaml.safe_load` on each touched file.


gaborbernat (changes requested):
failing checks
[pre-commit.ci - pr](https://results.pre-commit.ci/run/github/151871286/1778795999.7T_9YIWpT8aiXUbfRLFj4Q)
pre-commit.ci - pr — checks completed with failures

arpitjain099:
> failing checks
> [pre-commit.ci - pr](https://results.pre-commit.ci/run/github/151871286/1778795999.7T_9YIWpT8aiXUbfRLFj4Q)
> pre-commit.ci - pr — checks completed with failures

@gaborbernat I'll check. Thanks!

arpitjain099:
Reopened as #1822. While addressing the pre-commit (yamlfmt) failure I force-pushed from a stale local base, which orphaned this branch's history (no common ancestor with `main`) and left this PR's head unable to re-sync, so I could not push the fix here. #1822 is the same change rebuilt cleanly on current `main` as a single signed commit, with the `permissions:` block placed so yamlfmt passes. Sorry for the noise, @gaborbernat.

PR #1825, opened 2026-06-03
Add --no-path-check to skip the on-PATH check in pipx run

### LINKED ISSUE #1824: Allow suppressing "is already on your PATH and installed at ... Downloading and running anyway." warning in `pipx run`
**What's the problem this feature will solve?**

I want to provide a shim script for pipx-installable tool that runs `pipx run ... "$@"` transparently.
For users convenience, the shim has the same name as the app it downloads and runs.
This works great except that running the shim triggers a warning

```
⚠️  app_name is already on your PATH and installed at /path/to/shim/app_name. Downloading and running anyway.
```

Option to disable the PATH check.

**Alternative Solutions**

Using `--quiet` to suppress this warning, and all other warnings.
I feel like this is legitimate use-case and would prefer not to disable all other warnings.

**Additional context**

None

findepi:
For shim scripts that wrap `pipx run <app>` under the same name as the app, the on-PATH check fires on every invocation and is expected to print a warning.  `--quiet` silences it, but also drops every other warning.  `--no-path-check` skips just this check (and therefore its warning).

- fixes https://github.com/pypa/pipx/issues/1824

### DIFF
--- a/changelog.d/1824.feature.md
+++ b/changelog.d/1824.feature.md
@@ -1,0 +1,2 @@
+Add `--no-path-check` to `pipx run` to skip the warning when the app is already on `PATH`. Useful for shim scripts that
+wrap `pipx run <app>` under the same name as the app.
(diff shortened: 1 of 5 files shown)


gaborbernat (approved):

findepi:
thank you!

PR #1842, opened 2026-06-18
Add --dry-run flag to pipx ensurepath

### LINKED ISSUE #1014: Provide `ensurepath` with a check option or create an `init` command

Currently, `pipx ensurepath` does not tell me what changes will it make.
I would like to stay in control: see what changes it would make, and then decide if it's okay with me.

- Add a `--check` option that checks for necessary changes and outputs them.
- Add an `init` command that outputs the shell config changes - like `pyenv` does it.

somaz94:
Adds a `--dry-run` flag to `pipx ensurepath` that reports which directories
would be added to `PATH` without modifying `PATH` or any shell configuration
file, so users can preview the change before running it for real (closes #1014).

The flag threads a `dry_run` parameter through `ensure_pipx_paths` →
`ensure_path`; in dry-run mode `userpath.append`/`prepend` are never called and
a "Would append/prepend ..." line plus a "no changes were made" footer are
printed instead.

- [x] added news fragment in `changelog.d/` folder (`feature`)
- [x] updated/extended the documentation

### Validation
- New `tests/test_ensurepath.py` (4 tests — first test file for this command):
  asserts dry-run never calls `userpath.append`/`prepend`, never mutates config,
  and checks the footer + `EXIT_CODE_OK`.
- `ruff check` + `ruff format --check` clean on all changed files.
- End-to-end `pipx ensurepath --dry-run --force` prints `Would append ...` and
  the no-changes footer without touching PATH.

### DIFF
--- a/changelog.d/1014.feature.md
+++ b/changelog.d/1014.feature.md
@@ -1,0 +1,1 @@
+Add a `--dry-run` flag to `pipx ensurepath` that reports which directories would be added to `PATH` without modifying `PATH` or any shell configuration file.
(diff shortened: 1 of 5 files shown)


gaborbernat (approved):
LGTM

PR #1848, opened 2026-06-26
Keep trash cleanup non-fatal for locked files

### LINKED ISSUE #1847: Windows PermissionError when clearing trash folder
**Issue**

If a running service has a file in the trash folder locked, subsequent attempts to run `pipx` always fail with `PermissionError`.

**Environment**

Provide at least:

- OS: Windows
- Shell: Powershell, bash
- Python version and path: 3.10
- `pipx --version` output: 1.15.0

The failure is on a different machine, but the traceback points to https://github.com/pypa/pipx/blob/9ba7965d4cb964ec58d3231cde437083f9e8cb8b/src/pipx/main.py#L1374 as the offending line.

Specifically, what's happening:

* try to `pipx uninstall` a command backing a running Windows service
* the tree removal fails due to the service having some DLLs and the wrapper executable open, so the offending files get moved to the trash folder instead of deleted immediately
* subsequent `pipx` commands attempt to clear the trash folder, but fail, because the service is still running and still has the affected files locked. They won't be unlocked until the machine is restarted or the service is explicitly stopped.

I can avoid the failures by actually shutting down the service or by moving the pending trash out of the way, but it would be more convenient if `pipx` remained usable until the next machine restart even in the face of poor service management practices.

**Suggested resolution**

Change the logic in https://github.com/pypa/pipx/blob/main/src/pipx/util.py#L48 to always pass `ignore_errors=True` to `shutil.rmtree`. The `safe_rm` flag would then just control whether the dir was moved to the trash folder or merely skipped with a warning if it couldn't be deleted entirely.

The `safe_rm` flag could even be dropped entirely and replaced by a "deleting trash" check that checks to see if the given path is already a subdirectory of `paths.ctx.trash`.

cyphercodes:
### Summary

- Keep `rmdir(..., safe_rm=False)` non-fatal when files remain locked during trash cleanup.
- Preserve the existing manual-cleanup warning when the trash directory cannot be removed.
- Add regression coverage and a towncrier news fragment.

Fixes #1847

### Test plan

- `uvx --from 'tox>=4.45' tox run -e 3.11 -- tests/test_util.py -k rmdir`
- `uvx --from 'tox>=4.45' tox run -e 3.11 -- tests/test_util.py`
- `uvx --from 'tox>=4.45' tox run -e lint`

### Checklist

- [x] ran the linter to address style issues (`pre-commit run --all-files` via `tox run -e lint`)
- [x] updated/extended the documentation (not applicable beyond the news fragment for this internal cleanup fix)

### DIFF
--- a/changelog.d/1847.bugfix.md
+++ b/changelog.d/1847.bugfix.md
@@ -1,0 +1,1 @@
+Keep pipx startup usable when locked files prevent trash cleanup.
--- a/src/pipx/util.py
+++ b/src/pipx/util.py
@@ -53,8 +53,10 @@ def rmdir(path: Path, safe_rm: bool = True) -> None:
     # Windows doesn't let us delete or overwrite files that are being run
     # But it does let us rename/move it. To get around this issue, we can move
     # the file to a temporary folder (to be deleted at a later time)
-    # So, if safe_rm is True, we ignore any errors and move the file to the trash with below code
-    shutil.rmtree(path, ignore_errors=safe_rm)
+    # Always ignore errors so locked files don't make cleanup fatal. If the
+    # directory is still present afterwards, safe_rm controls whether it is
+    # moved to the trash or left in place with a warning.
+    shutil.rmtree(path, ignore_errors=True)
 
     # move it to be deleted later if it still exists
     if path.is_dir():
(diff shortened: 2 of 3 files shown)


gaborbernat (approved):
LGTM

Issue #1853, opened 2026-06-29
`pipx uninject` leaves man page symlinks behind


sronix:
**Issue**

Injecting a package with `--include-apps` and then uninjecting it removes the
app symlinks but not the man page symlinks. The man page symlink stays in
`PIPX_MAN_DIR` and is now broken, since the venv file it points to is gone.
`pipx uninstall` removes the same symlink correctly.

    pipx install black
    pipx inject black pycowsay --include-apps
    ls "$(pipx environment --value PIPX_MAN_DIR)/man6"   # pycowsay.6
    pipx uninject black pycowsay
    ls "$(pipx environment --value PIPX_MAN_DIR)/man6"   # pycowsay.6 still here, now dangling

Expected: uninject removes the man page symlinks it created, like uninstall does.

**Environment**

- OS: macOS 26.5.1
- Shell: zsh 5.9
- Python version and path: 3.14.2, /opt/homebrew/bin/python3
- `pipx --version` output: 1.8.0

Also reproduces on current main.

**Output of the failing command**

    $ pipx uninject black pycowsay
    uninstalling pycowsay...
    Uninjected package pycowsay and its dependencies from venv black

    $ ls -l "$(pipx environment --value PIPX_MAN_DIR)/man6/pycowsay.6"
    pycowsay.6 -> .../venvs/black/share/man/man6/pycowsay.6   # target no longer exists

sronix:
Fixed by #1854

PR #1854, opened 2026-06-30
Fix uninject leaving man page symlinks behind

sronix:
Refs #1853.

`uninject` removed an injected package's app symlinks but not its man page
symlinks. In `get_include_resource_paths` the man branch added the relative
path (`man6/pycowsay.6`) to the removal set, while the app branch added the
absolute path. `os.unlink` then gets the relative path, fails with
`FileNotFoundError`, and the error is ignored, so the symlink stays.
`uninstall` handles the same file correctly because it uses the absolute path.
The man branch has stored the relative path since it was added in #1767.

Fix: add the absolute `man_path`; the relative path stays for the `man_pages`
membership check. Adds `test_uninject_man_page`.

### DIFF
--- a/changelog.d/1853.bugfix.md
+++ b/changelog.d/1853.bugfix.md
@@ -1,0 +1,1 @@
+Fix `pipx uninject` leaving the injected package's man page symlinks behind.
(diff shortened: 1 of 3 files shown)


gaborbernat (approved):
LGTM

PR #1855, opened 2026-07-01
Refresh metadata after runpip installs

### LINKED ISSUE #1473: `package_or_url` remains local path even after installed package is no longer editable
**Describe the bug**

When installing a package with `--editable`, the `package_or_url` field in `pipx_metadata.json` is updated to the local path of the package.

Even after forcing the installation of a published version of the package, the `package_or_url` field is not updated back to the name of the package, creating a mostly hidden and confusing coupling to the local source code. An example is that running `pipx upgrade` to try to get the latest published version can incorrectly say that the package is already up-to-date, when really pipx has only checked that the installed version is at least as new as the spec from the local source code.

I've tried a few different ways to install the published package, but none seem to have the effect of updating `package_or_url`. One way around this is to `pipx reinstall`, but this has at least the caveat that any other package that's been `pipx inject`ed has to be reinjected.

**How to reproduce**

1. Install a package from local source code using `pipx install --editable /path/to/<package>`
2. Confirm using `pipx runpip <venv> list` to confirm it's the editable package that's installed
3. Install a published version of the package using `pipx runpip <venv> install <package>==X.Y.Z --force`
4. Confirm using `pipx runpip <venv> list` that the package is now the published version
5. Observe that the `package_or_url` field in `~/.local/pipx/venvs/<venv>/pipx_metadata.json` is still set to `/path/to/package`

**Expected behavior**

Expect that the `package_or_url` field can get set back to `<package>` when the installed version of the package is the published one rather than the local one.

itscloud0:
## Summary
Fixes #1473.

When `pipx runpip <venv> install ...` replaces the main package, `pipx_metadata.json` keeps the old source path and editable flag. That leaves later `upgrade` and `reinstall` commands coupled to the stale install spec instead of the package that is actually installed now.

This change:
- inspects direct `pip install` specs passed through `runpip`
- refreshes main-package metadata when one of those specs replaces the main package
- keeps the stored editable flag in sync with the replacement install
- adds a regression test that swaps an editable local install to a wheel install and checks the metadata update
- adds a changelog fragment for the bugfix

## Testing
- `uv run --group test pytest tests/test_runpip.py -q -k 'runpip'`
- `uv run --group lint pre-commit run --files src/pipx/commands/run_pip.py tests/test_runpip.py changelog.d/1473.bugfix.md`

### Checklist
- [x] ran the linter to address style issues (`pre-commit run --all-files` equivalent on changed files)
- [x] added news fragment in `changelog.d/`
- [ ] updated/extended the documentation (not needed for this internal metadata sync)

### DIFF
--- a/changelog.d/1473.bugfix.md
+++ b/changelog.d/1473.bugfix.md
@@ -1,0 +1,2 @@
+Refresh `pipx_metadata.json` after `pipx runpip ... install` replaces the main package, so future `upgrade` and
+`reinstall` commands stop following a stale editable source path.
--- a/src/pipx/commands/run_pip.py
+++ b/src/pipx/commands/run_pip.py
@@ -1,8 +1,115 @@
 from pathlib import Path
 
+from packaging.utils import canonicalize_name
+
+from pipx.commands.common import package_name_from_spec
 from pipx.constants import ExitCode
 from pipx.util import PipxError
 from pipx.venv import Venv
+
+
+def _iter_install_specs(pip_args: list[str]) -> list[tuple[str, bool]]:
+    if not pip_args or pip_args[0] != "install":
+        return []
+
+    specs: list[tuple[str, bool]] = []
+    options_with_value = {
+        "-c",
+        "--constraint",
+        "--config-settings",
+        "-f",
+        "--find-links",
+        "-i",
+        "--index-url",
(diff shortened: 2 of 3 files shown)


gaborbernat (approved):
LGTM

Issue #149, opened 2019-04-29
Feature request: Option to list available upgrades without performing them


Phidica:
I would like to see this as an option to `upgrade` and `upgrade-all`, where in the case of the former it would tell you if an upgrade is available for the given package, and in the latter it would list the packages for which an upgrade is available

cs01:
Maybe an option like `--dry-run`?

Phidica:
Yes, that would work

cs01:
I started implementing this using `pip search PACKAGE`, but it does not do exact matching, nor can it (https://github.com/pypa/pip/issues/3354). If there is an API pipx can use, I am happy to add it. 

Another way to implement this would be to install to a temporary directory, get the version, then erase the temporary directory, but I am going to hold off on that.

Phidica:
Hmm, I didn't realise `pip search` was so fraught with troubles. And it seems they may even consider disabling it in the future (pypa/pip#5216) without a better API powering it

It seems that running `pip install <name>==` will prompt an error message (because a version number is missing after the `==`) containing a list of versions which *are* available. Eg, 
```
$ pip3 install pipx==
Collecting pipx==
  Cache entry deserialization failed, entry ignored
  Could not find a version that satisfies the requirement pipx== (from versions: 0.1.0, 0.12.0.0, 0.12.0.1, 0.12.0.2, 0.12.0.3, 0.12.0.4, 0.12.1.0, 0.12.2.0, 0.12.3.0, 0.12.3.1, 0.12.3.2, 0.12.3.3, 0.13.0.0b1, 0.13.0.0, 0.13.0.1, 0.13.1.0)
No matching distribution found for pipx==
```
Would parsing this output be acceptable for finding the latest available version of a package?

Phidica:
Studying the debug output of `pip`, it appears this is internally implemented by analysing the links on the page `https://pypi.python.org/simple/<name>/` eg https://pypi.python.org/simple/pipx/

cs01:
Looks like this url will work:

https://pypi.org/pypi/PACKAGE/json

StaticPH:
Not sure if it will be of much use to others, but I wrote a proof-of-concept bash function that does the job well enough, if not very elegantly.
I use this on a machine with Bash4.4 and GNU coreutils, but I don't think I used anything that wont also work with the BSD counterparts.
```bash
pipx-outdated() {
	echo "OUTDATED PACKAGES:"
[...]

PR #1916, opened 2026-07-12
✨ feat(outdated): list available upgrades

gaborbernat:
For [#149](https://github.com/pypa/pipx/issues/149), `pipx list --outdated` queries the pip or uv backend of each environment for newer releases without changing the environment. The text view prints installed and available versions; `--json` exposes the same data for scripts.

pipx forwards index settings from installation metadata and filters backend output to managed packages. It records skip reasons for editable and non-index sources.

The lock spans each query, blocking concurrent metadata changes.

### DIFF
--- a/changelog.d/149.feature.md
+++ b/changelog.d/149.feature.md
@@ -1,0 +1,1 @@
+Add `pipx list --outdated` and JSON output.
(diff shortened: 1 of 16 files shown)


justinmayer:
That's exciting, Bernát! Thank you for working on this. One question about the following…

> Pass `--json` to read the installed and available versions in a script.

Does that mean that in the current implementation, `pipx list --outdated` will *not* show both the installed and available version numbers? I ask because it seems like that would be useful information to have in the non-JSON view. For example, it seems Pip shows this information when running `pip list --outdated`:

```
~ ➤ python -m pip list --outdated

Package Version Latest Type
------- ------- ------ -----
pip     26.0.1  26.1.2 wheel
```

Perhaps I'm misunderstanding the documentation in the PR, and if so, my apologies! Otherwise, is there some specific reason not to include the version number information in the non-JSON view?

PR #1951, opened 2026-07-14
✨ feat(reset): add pipx reset to start over from a fresh install

### LINKED ISSUE #1606: Proposal: pipx clean/reset

Pipx environment can become corrupted due to aborted installations (packages with missing metadata) or aborted shared upgrades.

Example error:

```
ModuleNotFoundError: No module named 'pip._internal.cli'
```

For this and other cases with corrupted metadata that are hard to recover from, it would be useful if pipx could reset itself to a first-installation state.

Completely reset pipx to first installation state

Alternative is to manually delete pipx state. This is cumbersome however, as the location depends on the platform and each user needs to find out where it is located.

gaborbernat:
An interrupted install or a shared upgrade that dies halfway can leave pipx in a state `pipx repair` cannot mend, the shared environment whose pip stops importing being the case [issue #1606](https://github.com/pypa/pipx/issues/1606) opens with. Recovering from that meant deleting the pipx directories by hand, and where those sit depends on the platform, so each user had to go and find them. 🧹

`pipx reset` uninstalls every package pipx manages and removes the shared libraries, the caches, the standalone interpreters, the logs and the trash, which leaves pipx as it stood the day you installed it. Uninstalling runs first and carries its weight: pipx links the apps and the man pages into the bin and man directories, which sit outside the pipx home, so removing the home on its own would leave those links behind pointing at nothing.

The command asks before it removes anything. `--force` answers for a script, `--dry-run` reports what would go and stops, and `--output json` hands the packages and the removed locations to a caller that wants to read them. ✨

One wrinkle shapes the code: pipx opens a log file for every command it runs, including this one, and Windows declines to unlink a file it still holds open. The logs directory therefore keeps the log of the reset itself and loses the rest.

`pipx cache purge` still clears the run cache on its own, for anyone who wants that and nothing more.

Closes #1606

### DIFF
--- a/changelog.d/1606.feature.md
+++ b/changelog.d/1606.feature.md
@@ -1,0 +1,3 @@
+Add `pipx reset`, which uninstalls every pipx-managed package and removes the shared libraries, the caches, the
+standalone interpreters, the logs and the trash, so pipx starts over from a fresh install. It asks before it removes
+anything, `--force` answers for a script, and `--dry-run` lists what would go.
--- a/docs/how-to/troubleshoot.md
+++ b/docs/how-to/troubleshoot.md
@@ -58,6 +58,25 @@ pipx uninstall <mypackage>
 pipx install <mypackage>
 ```
 
+## Start over from a fresh install
+
+An interrupted install or shared-library upgrade can leave state that `pipx repair` cannot mend, such as a shared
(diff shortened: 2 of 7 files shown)


sigma67:
Thank you 🤩

PR #1966, opened 2026-07-17
fix: don't crash on non-UTF8 shebang in foreign bin-dir executables

### LINKED ISSUE #1965: Crash (UnicodeDecodeError) on install/list when a foreign executable is in pipx's bin dir
## Describe the bug

`pipx install` (and `pipx list`) crashes with an unhandled `UnicodeDecodeError` when
the pipx bin directory (`~\.local\bin` on Windows) contains an executable that was
**not created by pipx** and is not a distlib/pip launcher.

pipx scans every file in the bin directory to determine resource ownership
(`_venv_owned_resource_paths` → `get_exposed_paths_for_package` →
`_resource_belongs_to_venv` → `_copy_belongs_to_venv` →
`_copy_launcher_targets_venv`), and unconditionally calls `os.fsdecode()` on bytes
read from the file assuming it is a distlib-style launcher with a UTF-8-encoded
interpreter path embedded in it. If the file is a large, unrelated binary (in my
case, a ~250 MB standalone `claude.exe`, the Claude Code CLI, placed in the same
directory by an unrelated installer), the byte sequence read is not valid UTF-8 and
the call raises, aborting the entire `install`/`list` operation with no
recoverable error message.

## Expected behavior

pipx should not crash outright when it encounters a file in its bin directory
that isn't a launcher it created. At minimum, `_copy_launcher_targets_venv` should
catch `UnicodeDecodeError` (and other malformed-launcher failures) and treat the
file as "not owned by this venv" instead of propagating the exception.

## How to reproduce

1. On Windows, install pipx into a dedicated venv (not via Scoop's shiv zipapp).
2. Ensure `~\.local\bin` (or whatever `--local-bin-dir`/`PIPX_BIN_DIR` resolves to)
   also contains an unrelated large executable placed there by another tool
   (in my case, Claude Code's standalone `claude.exe`).
3. Run `pipx install <any package>` or `pipx list`.
4. Observe the traceback below.

Moving the unrelated executable out of the pipx bin directory immediately resolves
the issue, confirming the file itself (not its content/purpose) is the trigger.

## Traceback
```
File "<frozen runpy>", line 203, in _run_module_as_main
File "<frozen runpy>", line 88, in run_code
File "C:\Users<user>\venv\pipx\Scripts\pipx.exe_main.py", line 5, in <module>
sys.exit(cli())
File "C:\Users<user>\venv\pipx\Lib\site-packages\pipx\main.py", line 2023, in cli
return _dispatch(sys.argv[1:])
File "C:\Users<user>\venv\pipx\Lib\site-packages\pipx\main.py", line 2016, in _dispatch
File "C:\Users<user>\venv\pipx\Lib\site-packages\pipx\main.py", line 344, in run_pipx_command
result = args.func(args, ctx)
File "C:\Users<user>\venv\pipx\Lib\site-packages\pipx\main.py", line 596, in _cmd_install
return commands.install(
None, ... , emit_output=False,
)
File "C:\Users<user>\venv\pipx\Lib\site-packages\pipx\commands\install.py", line 247, in install
run_post_install_actions(
File "C:\Users<user>\venv\pipx\Lib\site-packages\pipx\commands\common.py", line 587, in run_post_install_actions
_remove_stale_venv_resources(previous_resource_paths, venv, local_bin_dir, local_man_dir)
File "C:\Users<user>\venv\pipx\Lib\site-packages\pipx\commands\common.py", line 608, in _remove_stale_venv_resources
owned: Final[set[Path]] = _venv_owned_resource_paths(venv, local_bin_dir, local_man_dir)
File "C:\Users<user>\venv\pipx\Lib\site-packages\pipx\commands\common.py", line 619, in _venv_owned_resource_paths
owned: set[Path] = get_exposed_paths_for_package(venv.bin_path, local_bin_dir, app_paths)
File "C:\Users<user>\venv\pipx\Lib\site-packages\pipx\commands\common.py", line 370, in get_exposed_paths_for_package
if _resource_belongs_to_venv(resource_path, venv_resource_path, local_resource_dir, package_resource_paths)
File "C:\Users<user>\venv\pipx\Lib\site-packages\pipx\commands\common.py", line 384, in _resource_belongs_to_venv
return _copy_belongs_to_venv(resource_path, venv_resource_path, package_resource_paths)
File "C:\Users<user>\venv\pipx\Lib\site-packages\pipx\commands\common.py", line 398, in _copy_belongs_to_venv
if _copy_launcher_targets_venv(resource_path, venv_resource_path):
File "C:\Users<user>\venv\pipx\Lib\site-packages\pipx\commands\common.py", line 426, in _copy_launcher_targets_venv
if interpreter and Path(os.fsdecode(interpreter)).parent.samefile(venv_resource_path):
File "<frozen os>", line 864, in fsdecode
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xeb in position 1: invalid continuation byte
```

## Environment

- pipx version: 1.16.0
- Install method: pipx installed both via Scoop (`main` bucket, shiv zipapp) **and**
  standalone in a dedicated venv — same crash reproduced in both.
- OS: Windows 11
- Python: installed via the `py` launcher (python.org)
- `~\.local\bin` also contained: `claude.exe` (Claude Code CLI, ~250 MB standalone
  binary, unrelated to pipx), placed there by a separate installer that reuses the
  conventional `~/.local/bin` path.

## Suggested fix

Wrap the launcher-parsing logic in `_copy_launcher_targets_venv` in a
try/except for `UnicodeDecodeError` (and ideally any other malformed-binary
condition), returning `False` (not owned by this venv) instead of propagating.
This makes resource-ownership scanning robust against arbitrary third-party
executables coexisting in the same bin directory — a realistic scenario on
Windows where several CLI tools default to `~/.local/bin`.

## Workaround

Point pipx to a dedicated bin directory not shared with other tools:

```powershell
pipx install <package> --local-bin-dir "C:\path\to\dedicated\pipx\bin"
```
or permanently via the `PIPX_BIN_DIR` environment variable.

alberto743:
Fixes #1965

## Summary

`pipx install` and `pipx list` crash with an unhandled `UnicodeDecodeError`
when the pipx bin directory contains a file that isn't a launcher pipx
created — for example another CLI tool's large standalone `.exe` sharing
`~/.local/bin` on Windows.

## Root cause

While determining which bin-dir resources are owned by a venv,
`_copy_launcher_targets_venv` (`src/pipx/commands/common.py`) scans a
file's *entire* byte content for a `#!` marker and decodes whatever follows
as the launcher's interpreter path via `os.fsdecode()`. For a large
unrelated binary, a stray `#!` byte pair can occur anywhere in the file,
and the bytes that happen to follow it are effectively random — not valid
UTF-8.

On Windows, `os.fsdecode()` uses the `utf-8`/`surrogatepass` codec, which
(unlike POSIX's `surrogateescape`) does not rescue genuinely invalid byte
sequences, so it raises `UnicodeDecodeError`. That exception wasn't in the
existing `except OSError:` guard, so it propagated and aborted the whole
`install`/`list` command instead of just being treated as "not a pipx
launcher."

## Fix

Widen the `except OSError:` to also catch `UnicodeDecodeError`, so a
malformed/foreign file at that stage is simply treated as not owned by the
venv, matching the handling already in place for other malformed-launcher
conditions.

## Testing

- Added `test_copy_launcher_targets_venv_ignores_invalid_utf8_shebang` in
  `tests/test_common.py`, which reproduces the failure deterministically
  (independent of host OS) by forcing `os.fsdecode` to raise
  `UnicodeDecodeError`, and asserts the function now returns `False`
  instead of propagating.
- Verified against the pre-fix code that the test fails with the exact
  traceback reported in #1965, and passes with the fix applied.
- Full `tests/test_common.py` suite passes locally.
- Ran the "tests" GitHub Actions workflow on a fork branch: all jobs green
  across Python 3.10–3.14 on Ubuntu, macOS, and **Windows** (the platform
  the bug is specific to).

## Use of AI tools

This PR — diagnosis, the fix itself, the regression test, and this
description — was developed with the assistance of Claude (Anthropic's
Claude Code), working from the traceback and repro steps in #1965. The
change has been reviewed by me, although I do not have a specific
competence of this codebase.

## Checklist

- [x] added news fragment in `changelog.d/` folder (`changelog.d/1965.bugfix.md`)
- [ ] updated/extended the documentation *(not applicable — internal
      resource-ownership logic, no user-facing behavior/API change)*

### DIFF
--- a/changelog.d/1965.bugfix.md
+++ b/changelog.d/1965.bugfix.md
@@ -1,0 +1,3 @@
+Stop crashing `install`/`list` with an unhandled `UnicodeDecodeError` when the bin directory contains a
+foreign binary that isn't a launcher pipx created (for example another CLI tool's `.exe` sharing
+`~/.local/bin` on Windows); such files are now treated as not owned by the venv.
--- a/changelog.d/1965.bugfix.md
+++ b/changelog.d/1965.bugfix.md
@@ -1,3 +1,3 @@
-Stop crashing `install`/`list` with an unhandled `UnicodeDecodeError` when the bin directory contains a
-foreign binary that isn't a launcher pipx created (for example another CLI tool's `.exe` sharing
-`~/.local/bin` on Windows); such files are now treated as not owned by the venv.
+Stop crashing `install`/`list` with an unhandled `UnicodeDecodeError` when the bin directory contains a foreign binary
+that isn't a launcher pipx created (for example another CLI tool's `.exe` sharing `~/.local/bin` on Windows); such files
+are now treated as not owned by the venv.
--- a/changelog.d/1965.bugfix.md
+++ b/changelog.d/1965.bugfix.md
@@ -1,3 +1,4 @@
-Stop crashing `install`/`list` with an unhandled `UnicodeDecodeError` when the bin directory contains a foreign binary
-that isn't a launcher pipx created (for example another CLI tool's `.exe` sharing `~/.local/bin` on Windows); such files
-are now treated as not owned by the venv.
+Stop `install` and `list` from crashing when the pipx bin directory contains an executable pipx did not create.
+Scanning for launcher ownership decoded arbitrary binary content as an interpreter path, which raised
+`UnicodeDecodeError` on Windows, or `ValueError` for a path holding a NUL byte. Such a file is now treated as not
+owned by the venv.
--- a/changelog.d/1965.bugfix.md
+++ b/changelog.d/1965.bugfix.md
@@ -1,4 +1,3 @@
-Stop `install` and `list` from crashing when the pipx bin directory contains an executable pipx did not create.
-Scanning for launcher ownership decoded arbitrary binary content as an interpreter path, which raised
-`UnicodeDecodeError` on Windows, or `ValueError` for a path holding a NUL byte. Such a file is now treated as not
-owned by the venv.
+Stop `install` and `list` from crashing when the pipx bin directory contains an executable pipx did not create. Scanning
+for launcher ownership decoded arbitrary binary content as an interpreter path, which raised `UnicodeDecodeError` on
+Windows, or `ValueError` for a path holding a NUL byte. Such a file is now treated as not owned by the venv.
--- a/src/pipx/commands/common.py
+++ b/src/pipx/commands/common.py
@@ -425,7 +425,10 @@ def _copy_launcher_targets_venv(resource_path: Path, venv_resource_path: Path) -
         try:
             if interpreter and Path(os.fsdecode(interpreter)).parent.samefile(venv_resource_path):
                 return True
-        except (OSError, UnicodeDecodeError):
+        except (OSError, ValueError):
+            # The scan reads every file in the bin directory, so a stray "#!" in a foreign binary leads here with
+            # arbitrary bytes rather than a path: os.fsdecode rejects undecodable ones under the surrogatepass
+            # codec Windows uses, and stat rejects an embedded NUL. Neither makes the copy ours.
             continue
     return False
 
--- a/src/pipx/commands/common.py
+++ b/src/pipx/commands/common.py
@@ -425,7 +425,7 @@ def _copy_launcher_targets_venv(resource_path: Path, venv_resource_path: Path) -
         try:
             if interpreter and Path(os.fsdecode(interpreter)).parent.samefile(venv_resource_path):
                 return True
-        except OSError:
+        except (OSError, UnicodeDecodeError):
             continue
     return False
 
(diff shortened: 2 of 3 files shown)


alberto743:
here is the outcome of the tests: https://github.com/alberto743/pipx/actions/runs/29617910282

gaborbernat:
Thank you for this, and for the write-up in particular. Pinning the crash on Windows decoding filesystem bytes with `surrogatepass` where POSIX uses `surrogateescape` is the right root cause, and the final fix rests on it.

Three PRs arrived for #1965, so I have folded them into #1970 and am closing this one in favour of that. #1970 widens the guard to `ValueError` rather than `UnicodeDecodeError`, which also covers a second crash the report does not mention: off Windows `os.fsdecode` escapes the bad bytes instead of raising, but a decoded path holding a NUL still reaches `stat`.

The commit credits you as a co-author. Thanks again for the careful diagnosis. 🙏

PR #1968, opened 2026-07-18
fix: handle UnicodeDecodeError in _copy_launcher_targets_venv

Dodothereal:
# fix: handle UnicodeDecodeError in _copy_launcher_targets_venv

## Summary

Fixes a crash (UnicodeDecodeError) that occurs when `pipx install` or `pipx list` encounters a non-UTF-8 executable file in the application's binary directory. The function `_copy_launcher_targets_venv` in `src/pipx/commands/common.py` was calling `os.fsdecode` on arbitrary bytes without handling decoding errors, causing an uncaught exception.

The fix wraps the `os.fsdecode` call in a try/except block that catches `UnicodeDecodeError` (and `OSError` for completeness) and returns `False`, indicating the resource does not belong to the venv, which allows the function to continue processing other files.

## Changes

- Modified `src/pipx/commands/common.py`: Wrapped the `os.fsdecode(interpreter)` call in a try/except block to catch `UnicodeDecodeError` and `OSError`, returning `False` on exception.
- Added unit tests in `tests/test_common.py`:
  - `test_copy_launcher_targets_venv_unicode_decode_error`: verifies that a non-UTF-8 file returns `False` without raising an exception.
  - `test_copy_launcher_targets_venv_valid_launcher`: ensures a valid launcher with a shebang pointing to the venv's python returns `True`.
  - `test_copy_launcher_targets_venv_launcher_pointing_outside`: ensures a launcher pointing outside the venv returns `False`.

## Testing

- Syntax check: `python3 -m py_compile src/pipx/commands/common.py` passed.
- Unit tests: Attempted to run the new and existing tests via `sandbox-run python -m pytest tests/test_common.py` but the sandbox environment encountered permission errors (`bwrap: setting up uid map: Permission denied`). Due to this environment issue, the tests were not executed in the sandbox. However, the changes are minimal and focused, and the new unit tests are designed to validate the fix.
- The existing tests in `tests/test_common.py` were not run due to the sandbox issue, but the change only adds exception handling and does not alter the existing success path.

---
### AI assistance disclosure

This contribution was produced by an autonomous AI coding agent (Claude Code) that @Dodothereal operates and monitors. @Dodothereal is accountable for it, will address review feedback promptly, and will close this PR immediately if this kind of contribution is unwelcome in this project. Commits carry an `Assisted-by: Claude Code` trailer.

### DIFF
--- a/changelog.d/1965.bugfix.md
+++ b/changelog.d/1965.bugfix.md
@@ -1,0 +1,4 @@
+Stop `install` and `list` from crashing when the pipx bin directory contains an executable pipx did not create.
+Scanning for launcher ownership decoded arbitrary binary content as an interpreter path, which raised
+`UnicodeDecodeError` on Windows, or `ValueError` for a path holding a NUL byte. Such a file is now treated as not
+owned by the venv.
--- a/changelog.d/1965.bugfix.md
+++ b/changelog.d/1965.bugfix.md
@@ -1,4 +1,3 @@
-Stop `install` and `list` from crashing when the pipx bin directory contains an executable pipx did not create.
-Scanning for launcher ownership decoded arbitrary binary content as an interpreter path, which raised
-`UnicodeDecodeError` on Windows, or `ValueError` for a path holding a NUL byte. Such a file is now treated as not
-owned by the venv.
+Stop `install` and `list` from crashing when the pipx bin directory contains an executable pipx did not create. Scanning
+for launcher ownership decoded arbitrary binary content as an interpreter path, which raised `UnicodeDecodeError` on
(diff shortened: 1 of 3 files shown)


gaborbernat:
Thanks for picking this up. Three PRs arrived for #1965, so I have folded them into #1970 and am closing this one in favour of that. The commit credits you as a co-author.

One note that might help next time. The `b"\xff\xfe"` payload holds no `#!`, so the marker search found nothing and the assertion passed against unpatched `main`, which means the test went green without the fix in place. A regression test earns its keep by failing first, so it is worth reverting `src/` and watching it go red before opening the PR. #1970 drives the same cases through the public scan on payloads that carry a marker, and patches `can_symlink` so the parse runs off Windows too.

Thanks again for the time you put in. 🙏

PR #1985, opened 2026-07-29
🐛 fix(zipapp): bundle platform-gated dependencies in the artifact

### LINKED ISSUE #1984: Released pipx.pyz omits colorama and tomli, so it fails to start on Windows and on Python 3.10
## Description

The published `pipx.pyz` does not contain the dependencies that are gated behind environment markers in `pyproject.toml`:

- `colorama>=0.4.4; sys_platform=='win32'`
- `tomli; python_version<'3.11'`

`tox run -e zipapp` builds the artifact with `shiv -c pipx -o pipx.pyz .`. shiv delegates to pip, and pip evaluates those markers against the interpreter doing the build. `.github/workflows/release.yml` runs that job on `ubuntu-latest` with a modern Python, so both markers are false and both distributions are resolved away — but the single artifact is then published for every platform and every supported Python.

## Reproduction

Against the current release asset:

```console
$ curl -sLO https://github.com/pypa/pipx/releases/latest/download/pipx.pyz
$ python -c "import zipfile; n=zipfile.ZipFile('pipx.pyz').namelist(); print([x for x in n if 'colorama' in x or 'tomli' in x])"
[]
```

For comparison, the marker-free dependencies are all present: `argcomplete`, `click`, `filelock`, `packaging`, `platformdirs`, `userpath`.

On Windows the failure surfaces as soon as stdout is a terminal, because `src/pipx/colors.py` imports `colorama` under `sys.platform == "win32" and PRINT_COLOR`:

```
File "...\.shiv\pipx.pyz_<hash>\site-packages\pipx\colors.py", line 13, in <module>
    import colorama
ModuleNotFoundError: No module named 'colorama'
```

On Python 3.10 `venv_inspect` fails the same way on `tomli`.

## Why CI and the build's own smoke test do not catch it

`[env.zipapp]` runs `pipx.pyz --version` right after building. That runs on Linux, where the `colorama` import is unreachable, with output redirected, which makes `sys.stdout.isatty()` false and skips the import even on Windows. It is exactly the combination in which both defects stay quiet. `python pipx.pyz --version` piped on Windows prints `1.16.3` and exits 0 for the same reason — the artifact only fails in a real console.

## Environment

pipx 1.16.3 release asset, Windows 11, Python 3.13.14, and inspected byte-wise so the platform of the reader does not matter.

## Suggested fix

Request the two marker-gated distributions explicitly in the shiv invocation so they land regardless of the builder's platform and version. Both wheels are pure-Python and small, so the artifact stays a single cross-platform file and grows by a few tens of kilobytes.

I have a patch and can open a PR against this issue. Verified locally by building the zipapp on Windows with Python 3.13, where `tomli`'s marker is false: before the change the artifact has no `tomli`, after it ships `tomli` 2.4.1 — same mechanism that covers `colorama` on the Linux release builder.

marko1olo:
Closes #1984.

`tox run -e zipapp` builds one `pipx.pyz` that is published for every platform and every supported Python, but shiv delegates to pip and pip evaluates the environment markers in `dependencies` against the interpreter that builds the artifact. The release workflow builds on `ubuntu-latest`, so `colorama; sys_platform=='win32'` and `tomli; python_version<'3.11'` are both resolved away and never reach the zipapp.

Requesting the two explicitly makes shiv install them regardless of the builder's platform and version. Both wheels are pure-Python and small, so the artifact stays a single cross-platform file.

I verified the mechanism rather than assuming it, by building the zipapp on Windows with Python 3.13 — where `tomli`'s marker is false and `colorama`'s is true, so `tomli` is the one the build drops:

```
before   packages in artifact: argcomplete, click, colorama, filelock, packaging, pipx, platformdirs, userpath
after    ... same, plus tomli 2.4.1
```

The same mechanism covers `colorama` on the Linux release builder, where it is the marker that evaluates false.

I did not touch the smoke test in `[env.zipapp]`, but it is worth saying why it never caught this: `pipx.pyz --version` runs on Linux with output redirected, which makes the `colorama` import unreachable and `sys.stdout.isatty()` false at the same time. If you would like a guard that asserts the marker-gated distributions are present in the built artifact, I am happy to add it here or in a follow-up — I left it out to keep this diff to the defect.

- [x] I have added a news fragment under `changelog.d/`
- [x] I have verified the change (built the artifact both ways and inspected its contents)

### DIFF
--- a/changelog.d/1984.bugfix.md
+++ b/changelog.d/1984.bugfix.md
@@ -1,0 +1,1 @@
+Fix the zipapp crashing at startup on Windows and on Python 3.10.
--- a/tox.toml
+++ b/tox.toml
@@ -95,7 +95,11 @@ description = "build zipapp via shiv"
 skip_install = true
 dependency_groups = ["zipapp"]
 allowlist_externals = ["{tox_root}{/}pipx.pyz"]
+# The zipapp is one artifact for every platform and interpreter, but pip evaluates the
+# environment markers in `dependencies` against the machine that builds it. Requesting the
+# marker-gated dependencies explicitly keeps them in the artifact, so it also runs on the
+# platforms the release builder is not.
 commands = [
-  ["shiv", "-c", "pipx", "-o", "{tox_root}{/}pipx.pyz", "."],
+  ["shiv", "-c", "pipx", "-o", "{tox_root}{/}pipx.pyz", ".", "colorama>=0.4.4", "tomli"],
   ["{tox_root}{/}pipx.pyz", "--version"],
 ]


gaborbernat:
superseeded by https://github.com/pypa/pipx/pull/1990

PR #1993, opened 2026-07-30
🐛 fix(tests): seed the pinned pip so upgrade-shared stays green

gaborbernat:
The `🧪 test` matrix fails on 3.14 and 3.15 with `No matching distribution found for pip==26.1.2`, which is what blocks the actions bump in #1992 (the failure has nothing to do with that PR). `test_upgrade_shared_pin_pip` pins pip to `26.1.2` to prove `--pip-args` places an exact version, but the local test index only ever seeded that value through the bare `pip` entry in `testdata/tests_packages/primary_packages.txt`, which tracks whatever pip is latest. 📌 Once pip `26.2` shipped, `26.1.2` dropped out of any index built from scratch.

The committed per-Python lists still carry `26.1.2`, so 3.10 through 3.13 stay green off their cached lists, while 3.14 and 3.15 have no committed list, rebuild the index from `primary_packages.txt`, and break. Seeding `pip==26.1.2` there by name pins it in place for every rebuild. It clears the `pip >= 26.1` floor that `upgrade-shared` appends yet stays behind the latest, so the assertion keeps proving the pin rather than matching whatever the default upgrade would pull.

Local checks confirm it two ways: the test passes on the committed-list path, and regenerating the index from the updated seed now lists `pip==23.3.2, 24.0, 26.1.2, 26.2`, so the 3.14 and 3.15 rebuild path has the version again. Once this lands, #1992 needs a `@dependabot rebase` to pick it up.

### DIFF
--- a/testdata/tests_packages/primary_packages.txt
+++ b/testdata/tests_packages/primary_packages.txt
@@ -22,6 +22,7 @@ nox[tox_to_nox]==2023.4.22
 pbr==5.6.0
 pip==23.3.2
 pip==24.0
+pip==26.1.2
 pip
 pycowsay==0.0.0.2
 pygdbmi==0.10.0.0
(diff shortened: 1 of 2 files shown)


gaborbernat:
Pushed this fix straight onto #1992 instead, so it rides in with the actions bump. Closing as redundant.

PR #1999, opened 2026-08-11
Handle multiple site-packages paths in virtual environments

Cycloctane:
Changes in Python 3.15 (<https://docs.python.org/3.15/whatsnew/3.15.html#venv>) make `venv` seperates platlib and purelib into different directories on platforms like fedora (which is also the behavior of `virtualenv`).

However, pipx only finds site-packages in purelib directory. This pull request adds platlib to search paths when necessary.


Cycloctane:
The reproducer test case in commit <https://github.com/pypa/pipx/pull/1999/commits/efcd89d53df47f7a830244be77e7e12754696a29> fails when running with python3.15b in fedora container. It is not reproducible in ubuntu runners.

gaborbernat (approved):

PR #2009, opened 2026-08-20
Fall back to 'pipx' when argv[0] is not pipx

### LINKED ISSUE #2005: Doc for pipx run uses sphinx-build as script name
On https://pipx.pypa.io/latest/reference/cli.html#pipx-run the first line of the help output is:

```
sphinx-build run [-h] [--quiet] [--verbose] [--skip-maintenance]
```

It should start with pipx.

SirHegel:
Fixes #2005.

## The problem

`prog_name()` returns `Path(sys.argv[0]).name` unconditionally:

```python
def prog_name() -> str:
    with contextlib.suppress(IndexError):
        prog = Path(sys.argv[0]).name
        return f"{sys.executable} -m pipx" if prog == "__main__.py" else prog
    return "pipx"
```

The CLI reference is generated by `sphinx_argparse_cli` from `pipx.main.build_parser`, which runs inside the docs build — so `sys.argv[0]` is `sphinx-build`, and every usage line on the published page reads:

```
sphinx-build run [-h] [--quiet] [--verbose] [--skip-maintenance]
```

## Why `:prog: pipx` does not already fix it

`docs/reference/cli.rst` does set `:prog: pipx`, but that only relabels the root. `argparse` computes each subparser's `prog` from its parent at `add_parser()` time, so `sphinx-build run` is baked in before the directive ever runs.

Reproduced on `main`:

```python
>>> sys.argv[0] = "/usr/local/bin/sphinx-build"
>>> p = build_parser()
>>> p.prog
'sphinx-build'
>>> # every subcommand inherits it
'sphinx-build run', 'sphinx-build install', 'sphinx-build list'
```

## The change

Only trust `argv[0]` when it actually names pipx:

| `argv[0]` | before | after |
|---|---|---|
| `/usr/bin/pipx` | `pipx` | `pipx` |
| `/usr/bin/pipx3` | `pipx3` | `pipx3` |
| `__main__.py` | `python -m pipx` | `python -m pipx` |
| `sphinx-build` | **`sphinx-build`** | `pipx` |
| `pytest` | **`pytest`** | `pipx` |

An alternate install name like `pipx3` is still the user's own invocation, so it is preserved. Anything else is some other process importing the parser, and falls back to the literal `pipx`.

## Tests

Added cases to the existing `test_prog_name` parametrisation, plus a test asserting no subparser usage line leaks the host program. Both fail before the change — `AssertionError: assert 'sphinx-build' == 'pipx'` — and pass after.

`tests/test_main.py`: **46 passed**.

### DIFF
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -75,7 +75,15 @@ def print_version() -> None:
 def prog_name() -> str:
     with contextlib.suppress(IndexError):
         prog = Path(sys.argv[0]).name
-        return f"{sys.executable} -m pipx" if prog == "__main__.py" else prog
+        if prog == "__main__.py":
+            return f"{sys.executable} -m pipx"
+        # Only trust argv[0] when it actually names pipx. The parser is also built
+        # by processes that are not pipx -- the docs build imports build_parser from
+        # sphinx-build -- and argparse bakes the parent prog into every subparser at
+        # construction time, so an unrelated argv[0] leaks into the whole CLI
+        # reference as "sphinx-build run", "sphinx-build install" and so on.
+        if prog.startswith("pipx"):
+            return prog
     return "pipx"
 
 
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -60,12 +60,33 @@ def test_version(capsys: pytest.CaptureFixture[str]) -> None:
     [
         ("/usr/bin/pipx", "", "pipx"),
         ("__main__.py", "/usr/bin/python", "/usr/bin/python -m pipx"),
+        # An alternate install name is still the user's own invocation, so keep it.
+        ("/usr/bin/pipx3", "", "pipx3"),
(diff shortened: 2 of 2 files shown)


SirHegel:
Closing — #2007 predates this by five days and fixes #2005 more surgically: it threads an explicit `prog` through `get_command_parser()` so only `build_parser()` overrides it, leaving `prog_name()` alone.

One difference worth having on record, since the two are not equivalent in scope. #2007 fixes the docs build. This one changed `prog_name()` itself to fall back to `"pipx"` whenever `argv[0]` is neither a `pipx*` binary nor `__main__.py`, which also covers any *other* host process that imports the parser — a test runner, an embedding tool. Whether that wider behaviour is wanted is a design call, not a bug fix, so #2007 is the right shape for this issue.

Sorry for the duplicate; I filtered #2005 on having no assignee and no comments and did not check for a PR already referencing it.

PR #2016, opened 2026-08-29
✨🐛 feat(cooldown): add PIPX_COOLDOWN, pin direct-reference spelling

### LINKED ISSUE #2014: Allow setting --cooldown in global config or env variable
Hello!

**What's the problem this feature will solve?** 

I would like to be able to set global dependency cooldown for pipx.
I know there is `--cooldown` option, but as far as I understand,
this requires me to specify every time I use `pipx install`,
`pipx upgrade` or `pipx upgrade-all`.

I would love for global dependency cooldown option that I can set in config,
or maybe a PIPX_COOLDOWN env variable.

**Alternative Solutions**

**Additional context**

https://github.com/pypa/pipx/issues/1811
https://pipx.pypa.io/stable/reference/cli.html

Thanks!

PS: If you accept PRs I can create a PR for this if you think this feature is needed.

### LINKED ISSUE #2015: pyproject declares packaging>=20, but the test suite only passes on packaging>=26.0
**Issue**

`pyproject.toml` declares `packaging>=20`, but pipx's own test suite fails on every packaging version below 26.0. 12 tests in `tests/test_package_specifier.py` fail from packaging 20 through 25.0 and pass from 26.0 onward.

Verified against `1138ac5` (Release 1.16.8):

| packaging | `pytest tests/test_package_specifier.py` |
| --- | --- |
| 20 | 12 failed, 71 passed |
| 22.0 | 12 failed, 71 passed |
| 24.2 | 12 failed, 71 passed |
| 25.0 | 12 failed, 71 passed |
| 26.0 | 83 passed |
| 26.3 | 83 passed |

**Cause**

packaging 26.0 changed how `Requirement.__str__` serializes direct-reference (URL) requirements — it now emits a space before the `@`:

```
packaging 25.0 -> 'black@ https://example.com/black.zip'
packaging 26.0 -> 'black @ https://example.com/black.zip'
```

`fix_package_name()` (`package_specifier.py:295`), `parse_specifier_for_metadata()` (`:73`) and `parse_specifier_for_upgrade()` all return `str(Requirement(...))`, and the tests assert the spaced form.

**Impact**

I don't want to overstate this: both forms are valid PEP 508 and re-parse to the same requirement, so I have not found a user-visible failure. The concrete problem is that the declared lower bound doesn't match what the code and tests actually require. That matters for downstream redistributors who build pipx against their own `packaging` — `>=20` tells them anything from 20 onward is supported, when the suite doesn't pass below 26.0. The specifier strings written into `pipx_metadata.json` and passed to pip on upgrade also differ depending on which packaging happens to be installed.

**Reproduce**

```bash
git clone https://github.com/pypa/pipx && cd pipx
python -m venv .venv
.venv/bin/pip install -e . --group test    # needs pip >= 25.1 for --group
.venv/bin/pip install 'packaging==25.0'
.venv/bin/python -m pytest -q tests/test_package_specifier.py   # 12 failed, 71 passed
.venv/bin/pip install 'packaging==26.0'
.venv/bin/python -m pytest -q tests/test_package_specifier.py   # 83 passed
```

**Environment**

- OS: Ubuntu 24.04.4 LTS
- Shell: bash
- Python version and path: 3.11.15, `.venv/bin/python`
- `pipx --version` output: 1.16.8

**Suggested fix**

Raise the floor to `packaging>=26`. Alternatively, if supporting older packaging matters, normalize the `@` spacing in `package_specifier.py` instead of returning `str(Requirement(...))` directly.

Happy to open a PR for whichever you'd prefer.

---

**How this was found**

I ran pipx's test suite across the whole range `packaging>=20` permits, then narrowed the diff to the exact call sites with [breakcheck](https://github.com/lovettsendit/breakcheck), a deterministic behavioural-comparison tool I maintain. It replays the same call in isolated environments built from both dependency versions and compares normalised observations:

```
CHANGED_UNDER_PROJECTION  packaging.requirements.Requirement
  src/pipx/package_specifier.py:295
    packaging 25.0 -> 'black[extra]@ https://github.com/ambv/black/archive/18.9b0.zip'
    packaging 26.0 -> 'black[extra] @ https://github.com/ambv/black/archive/18.9b0.zip'

CHANGED_UNDER_PROJECTION  packaging.requirements.Requirement
  src/pipx/package_specifier.py:73
    packaging 25.0 -> 'nox@ git+https://github.com/cs01/nox.git@5ea70723e9e6'
    packaging 26.0 -> 'nox @ git+https://github.com/cs01/nox.git@5ea70723e9e6'
```

The run writes a report plus a hash-bound evidence bundle that re-verifies independently (`breakcheck --verify report.json --evidence evidence.json` → `VERIFIED`), so those two observations are checkable without taking my word for them. Happy to attach the artifacts or the fixture file that reproduces the run if that's useful.

gaborbernat:
Two independent reports, one branch, one commit each.

[#2015](https://github.com/pypa/pipx/issues/2015) points out that `pyproject.toml` declares `packaging>=20` while the test suite only passed from `packaging` 26.0 on. `packaging` 26.0 changed how `Requirement.__str__` serializes direct references, adding PEP 508's space before the `@`, and pipx handed that string straight back from `parse_specifier_for_metadata`, `parse_specifier_for_upgrade` and `fix_package_name`. So the specifier pipx wrote into `pipx_metadata.json` and handed to the backend depended on which `packaging` sat in the environment. 🔧 Rather than raise the floor to 26, pipx now pins the separator itself, so redistributors building against their own older `packaging` keep working and the recorded spelling stays the same either way. The 12 failing cases in `tests/test_package_specifier.py` pass on 20, 22.0, 24.2, 25.0 and 26.2.

[#2014](https://github.com/pypa/pipx/issues/2014) asks for a way to set `--cooldown` once instead of on every command. A release-age policy belongs to the machine, not to a single invocation, and the backend-level workarounds (`PIP_UPLOADED_PRIOR_TO`, `UV_EXCLUDE_NEWER`) are backend-specific and leak into unrelated pip and uv runs, which is what `--cooldown` exists to avoid. ✨ `PIPX_COOLDOWN` covers `install`, `install-all`, `inject`, `upgrade`, `upgrade-all` and `run`. The issue also floated a config file; [the environment variable reference](https://pipx.pypa.io/stable/reference/environment-variables/) states that pipx has no user configuration file and every setting is an environment variable, so this stays with the existing pattern of `PIPX_DEFAULT_PYTHON` and `PIPX_DEFAULT_BACKEND`.

Precedence runs `--cooldown`, then `PIPX_COOLDOWN`, then the cooldown a venv recorded at install time. The variable outranks recorded metadata because it states the policy the machine wants now, and `--cooldown 0` opts a single command out. `install` keeps the flag and the variable as separate arguments, mirroring how it already treats `--backend` against `PIPX_DEFAULT_BACKEND`: a lock file pins every version, so `--lock` with an explicit `--cooldown` still errors, while `PIPX_COOLDOWN` yields rather than breaking locked installs for anyone who exports it. An invalid value fails with `PIPX_COOLDOWN must be unset or a non-negative integer`.

Closes #2014
Closes #2015


gaborbernat:
Splitting into one PR per issue: #2014 and #2015 are unrelated and should land separately.

PR #2017, opened 2026-08-29
🐛 fix(deps): require packaging>=26

### LINKED ISSUE #2015: pyproject declares packaging>=20, but the test suite only passes on packaging>=26.0

gaborbernat:
Install the same URL-based package on two machines and pipx can record it two different ways. `pipx install "black @ https://example.com/black.zip"` writes `black@ https://…` under one version of `packaging` and `black @ https://…` under another, then hands the matching string to pip or uv the next time it upgrades that package. The user typed the same command both times; the shape of what we store comes from a transitive dependency they never chose. 🔧

We also advertise a support window we never checked. We tell redistributors `packaging>=20` is fine, so a distro building pipx against its own older `packaging` would take that as a supported combination. As [#2015](https://github.com/pypa/pipx/issues/2015) reports, our own suite disagreed: 12 cases failed on releases 20 through 25.0 and passed from 26.0 on, where `packaging` started emitting PEP 508's space before the `@`.

Raising the floor to `packaging>=26` is the honest fix. The alternative, normalizing the separator inside pipx, buys support for a range that no CI job resolves against, so the claim would stay unverified and we would keep the workaround around to protect it. `packaging` is pure Python with no dependencies of its own and 26.0 landed in January 2026, so resolvers pick it up without friction, and pipx already sets comparable floors on `filelock`, `platformdirs` and `userpath`.

The cost sits with redistributors pinned to an older `packaging` in a frozen release; they hold their current pipx until they can move, which is the usual shape of a dependency bump. Existing environments keep working either way: an older recorded spelling still parses, and pipx rewrites it the next time it touches one.

Closes #2015

### DIFF
--- a/changelog.d/2015.bugfix.md
+++ b/changelog.d/2015.bugfix.md
@@ -1,0 +1,3 @@
+Require `packaging>=26`. Earlier releases serialize a direct reference as `name@ url` rather than PEP 508's
+`name @ url`, so the specifier pipx recorded in `pipx_metadata.json` and passed to the backend depended on which
+`packaging` happened to be installed.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,7 @@ dependencies = [
   "argcomplete>=1.9.4",
   "colorama>=0.4.4; sys_platform=='win32'",
   "filelock>=3.16",
-  "packaging>=20",
+  "packaging>=26",
   "platformdirs>=4.6",
   "tomli; python_version<'3.11'",
   "userpath>=1.6,!=1.9",


ap-wtioit:
This breaks compatiblity with debian trixie:
`pip3 install --break-system-packages pipx`
```
 > [ 4/12] RUN pip3 install --break-system-packages pipx:                                                                                                                                                                         
0.512 Collecting pipx
0.590   Downloading pipx-1.17.1-py3-none-any.whl.metadata (4.4 kB)
0.622 Collecting argcomplete>=1.9.4 (from pipx)
0.632   Downloading argcomplete-3.7.2-py3-none-any.whl.metadata (16 kB)
0.668 Collecting filelock>=3.16 (from pipx)
0.681   Downloading filelock-3.32.4-py3-none-any.whl.metadata (2.0 kB)
0.717 Collecting packaging>=26 (from pipx)
0.741   Downloading packaging-26.3-py3-none-any.whl.metadata (3.5 kB)
0.780 Collecting platformdirs>=4.6 (from pipx)
0.800   Downloading platformdirs-4.11.5-py3-none-any.whl.metadata (5.5 kB)
0.832 Collecting userpath!=1.9,>=1.6 (from pipx)
0.853   Downloading userpath-1.9.2-py3-none-any.whl.metadata (3.0 kB)
0.889 Collecting click (from userpath!=1.9,>=1.6->pipx)
0.908   Downloading click-8.5.0-py3-none-any.whl.metadata (2.6 kB)
0.935 Downloading pipx-1.17.1-py3-none-any.whl (153 kB)
0.987 Downloading argcomplete-3.7.2-py3-none-any.whl (43 kB)
1.012 Downloading filelock-3.32.4-py3-none-any.whl (99 kB)
1.040 Downloading packaging-26.3-py3-none-any.whl (129 kB)
1.069 Downloading platformdirs-4.11.5-py3-none-any.whl (23 kB)
1.096 Downloading userpath-1.9.2-py3-none-any.whl (9.1 kB)
1.112 Downloading click-8.5.0-py3-none-any.whl (125 kB)
1.133 Installing collected packages: platformdirs, packaging, filelock, click, argcomplete, userpath, pipx
1.150   Attempting uninstall: packaging
1.152     Found existing installation: packaging 25.0
1.155 error: uninstall-no-record-file
1.155 
1.155 × Cannot uninstall packaging 25.0
1.155 ╰─> The package's contents are unknown: no RECORD file was found for packaging.
1.155 
1.155 hint: The package was installed by debian. You should check if it can uninstall the package.
1.155 
```

gaborbernat:
`--break-system-packages` shows that you're doing the wrong thing, create a virtual environment and install i there.

PR #2022, opened 2026-08-29
🧪 test(conftest): run without symlink privilege on Windows

### LINKED ISSUE #2021: Test suite is unrunnable on Windows without symlink privilege: utils_temp_dir errors
with suppress(FileExistsError):
    (tmp_path / util_path.name).symlink_to(util_path)
Windows refuses symlink creation unless the process is elevated or Developer Mode is enabled. That raises OSError: [WinError 1314], which is not a FileExistsError, so suppress does not catch it and the fixture errors.

utils_temp_dir feeds pipx_temp_env, which most tests depend on, so effectively the whole suite errors on a normal Windows machine:

ERROR tests/test_uninstall_all.py::test_uninstall_all - OSError: [WinError 1314]
ERROR tests/test_uninstall_all.py::test_uninstall_all_legacy_venv[None] - OSError...
5 errors in 1.94s
Why CI doesn't catch it
tests.yml includes windows-2025, and it is green — those runners can create symlinks. The failure only appears for an unprivileged contributor, so the suite is permanently green in CI and permanently broken locally.

Suggested fix (implemented and verified)
Fall back to putting the utility's real directory on PATH when the symlink cannot be created. Copying or hard-linking is not an alternative: Windows resolves an executable's DLLs relative to its image path, and git.exe needs the siblings in mingw64/bin.

utils_temp_dir is referenced only inside conftest.py (three places), so returning a list of directories instead of one path is contained.

Result on Windows: tests/test_uninstall_all.py goes from 5 errors to 5 passed.

Patch: patches/pipx/0001-tests-make-utils_temp_dir-work-without-symlink-privi.patch

Environment
Windows-11-10.0.26200-SP0
Python 3.13.15
git core.symlinks=false (default for a non-elevated Windows user)

gaborbernat:
Fixes #2021, where a Windows user without symlink privilege cannot run the suite. `utils_temp_dir` symlinked `git` into a session scratch directory under `suppress(FileExistsError)`, but an unelevated Windows user without Developer Mode gets `OSError: [WinError 1314]`, which that suppress does not catch. The fixture feeds `pipx_temp_env`, so most tests error during setup. 🪟 The `windows-2025` runners hold the privilege, so CI stays green while local runs error out.

When the symlink call fails, fall back to putting the utility's own directory on PATH as a second entry. Copying or hard-linking `git.exe` is not an option, since Windows resolves an executable's DLLs relative to its image path and `git.exe` needs the siblings in `mingw64/bin`. Returning a list of directories instead of one path lets `pipx_temp_env_helper` join whatever entries it gets, and the fixture takes the name `utils_path_dirs` to match. `conftest.py` was its only consumer, so the rename stays contained.

Where symlinks work, PATH keeps the shape it had, and the scratch directory still hides the utility's neighbours from pipx. 🔗 The fallback widens PATH by one directory, the one holding `git` or `uv`. Credit to @iam-kira for the diagnosis and the original patch.

### DIFF
--- a/changelog.d/2022.misc.md
+++ b/changelog.d/2022.misc.md
@@ -1,0 +1,1 @@
+Make the test suite runnable on Windows without symlink privilege.
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -11,7 +11,7 @@ import subprocess
 import sys
 import tarfile
 import time
-from contextlib import closing, suppress
+from contextlib import closing
 from http import HTTPStatus
 from pathlib import Path
 from typing import TYPE_CHECKING, Final
@@ -266,7 +266,7 @@ def pipx_temp_env_helper(
     tmp_path: Path,
     monkeypatch: pytest.MonkeyPatch,
     request: pytest.FixtureRequest,
-    utils_temp_dir: Path,
+    utils_path_dirs: list[Path],
     *,
     pypi: str,
 ) -> None:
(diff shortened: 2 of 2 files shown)


gaborbernat:
Closing in favour of #2023, which @iam-kira opened with the same fix from the original report.

PR #2029, opened 2026-09-02
fix(cli): validate PIPX_MAX_LOGS before logging setup

hktitof:
### Summary

`PIPX_MAX_LOGS` set to a non-integer like `PIPX_MAX_LOGS=abc` crashed every pipx command (except `--version`) with a raw `ValueError: invalid literal for int() with base 10: 'abc'` traceback during logging setup, bcs `_setup_log_file()` parses it with `int()` before any command runs, the sibling env vars `PIPX_COOLDOWN` and `PIPX_FETCH_PYTHON` already validate their values at dispatch and stop with a clean error

This mirrors the cooldown pattern, the raw value is parsed once in `constants.py` (`_compute_max_logs` → `_MAX_LOGS` + `_MAX_LOGS_INVALID`), a new `_validate_max_logs()` runs in `_dispatch` before anything else, and an invalid value now stops with

```
PIPX_MAX_LOGS must be unset or a non-negative integer, got 'abc'.
```

Reproduced pre-change

```
$ pipx list
nothing has been installed with pipx 😴          # exit 0
$ PIPX_MAX_LOGS=abc pipx list
  ...
ValueError: invalid literal for int() with base 10: 'abc'
```

Post-fix exit code is 1 with the message above, valid values (0, 5, 10, 100) keep their existing pruning behavior unchanged, and blank values (empty or whitespace-only) fall back to the default 10 everywhere after a review caught `_setup_log_file()` still re-parsing the raw value with `int()` on them

### Related

Follows the same variable hot zone as #1988 (prune logs when `PIPX_MAX_LOGS` is 0) and #2024 (report `PIPX_MAX_LOGS` from `pipx environment`), no open issue tracks this exact crash, the env-var validation pattern is the one added for `PIPX_COOLDOWN` in #2018

### Tests

- `test_compute_max_logs`, 8 parse cases (unset, empty, whitespace-only, 10, whitespace-padded, 0, negative, garbage), fails without the change
- `test_cli_rejects_invalid_env_max_logs`, CLI exits 1 with the clean message, without the validation guard it raises AttributeError (no guard exists)
- `test_cli_blank_env_max_logs_falls_back_to_default`, empty and whitespace-only inputs run the whole CLI through log setup without crashing, fails on the old raw `int()` re-parse

`pytest tests/test_max_logs.py tests/test_main.py tests/test_environment.py` → 104 passed, `ruff check` and `ruff format` clean, all pre-commit hooks pass

### Checklist

- [x] added news fragment in `changelog.d/` folder (`2029.bugfix.md`, bugfix type)
- [ ] updated/extended the documentation, not needed bcs the env-var table already documents `PIPX_MAX_LOGS` (number of log files to keep, default 10), the code was what disagreed with it, after #2024

### DIFF
--- a/changelog.d/2029.bugfix.md
+++ b/changelog.d/2029.bugfix.md
@@ -1,0 +1,2 @@
+Report a clear error when PIPX_MAX_LOGS is not a non-negative integer, instead of crashing on every pipx command. Empty
+or whitespace-only values fall back to the default everywhere, including log setup.
--- a/src/pipx/constants.py
+++ b/src/pipx/constants.py
@@ -64,6 +64,22 @@ def _compute_cooldown(raw: str | None) -> tuple[int | None, bool]:
 _COOLDOWN, _COOLDOWN_INVALID = _compute_cooldown(_COOLDOWN_RAW)
 
 
+_MAX_LOGS_RAW = os.environ.get("PIPX_MAX_LOGS")
+
+
+def _compute_max_logs(raw: str | None) -> tuple[int, bool]:
+    if raw is None or not (stripped := raw.strip()):
+        return 10, False
+    try:
+        logs = int(stripped)
+    except ValueError:
+        return 10, True
+    return (logs, False) if logs >= 0 else (10, True)
+
+
+_MAX_LOGS, _MAX_LOGS_INVALID = _compute_max_logs(_MAX_LOGS_RAW)
+
+
 ExitCode = NewType("ExitCode", int)
 # pipx shell exit codes
 EXIT_CODE_OK = ExitCode(0)
--- a/src/pipx/main.py
+++ b/src/pipx/main.py
@@ -38,12 +38,15 @@ from pipx.constants import (
     _FETCH_PYTHON,
     _FETCH_PYTHON_INVALID,
     _FETCH_PYTHON_RAW,
+    _MAX_LOGS_INVALID,
+    _MAX_LOGS_RAW,
     EXIT_CODE_OK,
     EXIT_CODE_SPECIFIED_PYTHON_EXECUTABLE_NOT_FOUND,
     MINIMUM_PYTHON_VERSION,
     WINDOWS,
     ExitCode,
     FetchPythonOptions,
+    _compute_max_logs,
 )
 from pipx.emojis import hazard
 from pipx.interpreter import (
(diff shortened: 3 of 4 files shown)


yangfan-yf-yf (changes requested):
I reproduced a remaining traceback path on the current head. `_compute_max_logs()` treats both an empty string and whitespace-only input as the default `(10, False)`, so `_validate_max_logs()` returns normally. But logging setup still parses the raw environment value with `int(os.getenv("PIPX_MAX_LOGS", "10"))`.

On Windows, both of these still fail before `pipx list` can run:

```text
PIPX_MAX_LOGS="" pipx list
PIPX_MAX_LOGS="  " pipx list
```

They raise `ValueError` from `_setup_log_file`, whereas `PIPX_MAX_LOGS=garbage` now correctly produces the intended user-facing error. Could the default/validation and logging setup use the same parsed value, or could blank values be classified as invalid, with a CLI-level regression covering both empty and whitespace-only inputs? The new parser test passes for those values but does not exercise logging setup.

I also verified that `PIPX_MAX_LOGS=0`, `5`, and `10` still run `pipx list` successfully, and `pytest --net-pypiserver tests/test_max_logs.py -q` passes (14 tests).

hktitof:
good catch and thanks for testing this, reproduced here: blank PIPX_MAX_LOGS still crashed in _setup_log_file bcs it re-parsed the raw value with int() after the dispatch validation had already passed it

fixed by reading it through the same _compute_max_logs parse the validation uses, so empty and whitespace-only now fall back to the default 10 like the parser intended while garbage and negative keep the clean error, added your two inputs as a CLI regression through logging setup plus the parse matrix cases, 104 tests green and ruff clean, pushed in 66ee182

yangfan-yf-yf (approved):
Verified 66ee18211be6d6ff103db7a474b116d88dfe17ba. The shared parser now reaches logging setup, and the empty/whitespace-only crash I reported is fixed.

The 18 tests in `tests/test_max_logs.py` pass. I also checked fresh `python -m pipx list` processes for unset, empty, whitespace-only, `garbage`, `-1`, `0`, `5`, and `10`, with isolated home directories seeded with 15 old logs. Valid inputs exit successfully and retain the expected number of logs; invalid inputs report the intended error without a traceback or pruning the existing logs.

One correction to my previous review: the old parser test parameter list did not include empty or whitespace-only strings. Those two findings came from the CLI probes; my sentence saying the existing parser tests covered them was inaccurate. The new parameter cases and CLI regressions now cover both.

hktitof:
thanks for verifying on the exact head, that probe matrix with isolated home dirs is thorough, and the correction on the old test coverage note is fair, the new cli regression covers both the empty and whitespace-only inputs, all green here

PR #2034, opened 2026-09-11
Add redirect for old pipx Pages site

### LINKED ISSUE #2030: https://pypa.github.io/pipx/ is a 404 now
https://pypa.github.io/pipx/ used to be the site for pipx - it is now a 404.

It would be nicer if it redirected to the new site.

(This came up because a lot of my projects have READMEs that link to the old URL.)

GitHub Pages doesn't support regular HTTP redirects, but you can serve an HTML page like this instead:

    <meta http-equiv="refresh" content="0;url=https://pipx.pypa.io/">

zihjiang:
## Summary

Add a redirect page for the former `https://pipx.github.io/pipx/` site.

## Changes

- Add a standalone `index.html` with an immediate redirect to `https://pipx.pypa.io/`.
- Include a canonical link and a visible fallback link for browsers that do not follow the redirect.

## Validation

- Parsed the page with Python's standard-library HTML parser.
- `git diff --check`

Fixes #2030


gaborbernat:
This is not the righyt fix.

gaborbernat:
Thanks for picking this up, and sorry for the detour.

An `index.html` on `main` cannot serve the redirect. This repository had Pages switched off, so no file on `main` reaches a web server, and the docs themselves build on Read the Docs now. Such a page also covers the front page alone, while the links people follow are deep ones such as https://pypa.github.io/pipx/installation/.

I pushed a `gh-pages` branch with an `index.html` plus a `404.html` that Pages serves for unmatched paths, which lets the old page URLs redirect to their current homes. That branch supersedes this PR, and #2030 has the details.

